PR #133, opened 2020-03-18
Add docs for schema

rynowak:


rynowak:
`yaml >>> yml`.

Also tye always generates `.yaml` - so we're being consistent with how we talk about it.

rynowak:
This isn't the cause of the build break (docs only), so ima YOLO merge this.

PR #134, opened 2020-03-18
Remove access token from redis docs

rynowak:
This is no longer needed now that we're public.


rynowak:
This isn't the cause of the build break (docs on), so ima YOLO merge this.

PR #136, opened 2020-03-19
Update getting started to point at a real build.

rynowak:


glennc:
Worked for me.

rynowak:
Docs only, so bypassing the (queued) CI.

PR #141, opened 2020-03-19
Add support for running local containers

davidfowl:
- This support builds the application locally then volume mounts the application into the container when the --docker flag is specified. Added support for volume mounts and working directory to the DockerRunner.

Here's what mult-project looks like with `tye run --docker`

image

### DIFF
--- a/src/Microsoft.Tye.Hosting/DockerRunner.cs
+++ b/src/Microsoft.Tye.Hosting/DockerRunner.cs
@@ -69,7 +69,7 @@ namespace Microsoft.Tye.Hosting
             var serviceDescription = service.Description;
             var environmentArguments = "";
             var volumes = "";
-            var workingDirectory = docker.WorkingDirectory == null ? "" : $"-w {docker.WorkingDirectory}";
+            var workingDirectory = docker.WorkingDirectory != null ? $"-w {docker.WorkingDirectory}" : "";
 
             var dockerInfo = new DockerInformation(new Task[service.Description.Replicas]);
 
--- a/src/Microsoft.Tye.Hosting/DockerRunner.cs
+++ b/src/Microsoft.Tye.Hosting/DockerRunner.cs
@@ -68,6 +68,8 @@ namespace Microsoft.Tye.Hosting
 
             var serviceDescription = service.Description;
             var environmentArguments = "";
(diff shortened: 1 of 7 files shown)


src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        public async Task StartAsync(Model.Application application)
+        {
+            // This transforms a ProjectRunInfo into
+
+            foreach (var s in application.Services.Values)
+            {
+                if (s.Description.RunInfo is ProjectRunInfo project)
+                {
+                    await TransformProjectToContainer(application, s, project);
```
davidfowl: This builds the projects and can be done in parallel (like the process runner currently does)

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -68,6 +68,8 @@ private async Task StartContainerAsync(Tye.Hosting.Model.Application application
 
             var serviceDescription = service.Description;
             var environmentArguments = "";
+            var volumes = "";
+            var workingDirectory = docker.WorkingDirectory == null ? "" : $"-w {docker.WorkingDirectory}";
```
jkotalik: nit: invert condition

src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs
```diff
@@ -0,0 +1,93 @@
+﻿using System;
```
jkotalik: nit: copyright

src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs
jkotalik: But it isn't in parallel today, correct? Why can't you do a Task.Run on each and await all of them at the end?

davidfowl: I copied what the other processors do. Not quite Task.Run but it goes parallel if the build finishes asynchronously.

PR #142, opened 2020-03-19
States we're already in containers, when we're not

ChadJessup:
Changed the wording a little at the beginning that states that our tye app is already in containers. This isn't true, as the next step generates the images for the first time.

I'm not married to the new wording, feel free to suggest/edit if the voice is wrong.

### DIFF
--- a/docs/frontend_backend_deploy.md
+++ b/docs/frontend_backend_deploy.md
@@ -20,7 +20,7 @@ Before we deploy, make sure you have the following ready...
 
 ## Deploying the application
 
-Now that we have our application running locally, let's get it into multiple containers and deploy the application. In this example, we will deploy to Kubernetes by using `tye deploy`.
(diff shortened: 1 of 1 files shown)


docs/frontend_backend_deploy.md
```diff
@@ -20,7 +20,7 @@ Before we deploy, make sure you have the following ready...
 
 ## Deploying the application
 
-Now that we have our application running locally with multiple containers, let's deploy the application. In this example, we will deploy to Kubernetes by using `tye deploy`.
+Now that we have our application running locally, let's get it into multiple containers and deploy the application. In this example, we will deploy to Kubernetes by using `tye deploy`.
```
jkotalik:
```suggestion
Now that we have our application running locally, let's deploy the application. In this example, we will deploy to Kubernetes by using `tye deploy`.
```

I think we don't need to mention "multiple" here.

jkotalik:
Thanks

PR #147, opened 2020-03-19
Add link to K3s

ericsampson:
K3s seems like a good fit for people wanting to experiment with Tye.


RichiCoder1:
Only issue with adding k3s to this list is current the list is pretty Windows-centric while k3s is (by default) a linux-based solution and somewhat weird to get up and running in WSL.

rynowak:
We definitely don't intend to be windows-centric. Lot of people are dev'ing .net on windows/wsl/mac/linux

RichiCoder1:
@rynowak Absolutely. Just saying it has some caveats. Maybe at some point closer to maturity doing something like: https://docs.tilt.dev/choosing_clusters.html

ericsampson:
> We definitely don't intend to be windows-centric. Lot of people are dev'ing .net on windows/wsl/mac/linux

@rynowak yeah just as one data point, I work at a wholly .NET shop and probably 50% of our devs use a Mac laptop running OS X :)

PR #148, opened 2020-03-19
Use dotnet publish to make static assets work

davidfowl:
- StaticAssets uses a physical path which doesn't work in the container without doing something crazy like rewriting the path. Instead, just publish the application before running it.
- This will prevent it from using relative paths for static assets

Found this while running @rynowak's application.


rynowak:
This is just for the container code path right? Doesn't affect a regular `tye run`?

davidfowl:
Correct.

Issue #98, opened 2020-03-12
Use the Tye dashboard APIs in our tests instead of the OM directly


rynowak:
We're using the OM directly for our tests, but it's a bit #problematic. 

- This allows the APIs to regress
- Doesn't validate that the APIs are useful
- Couples us to running the dashboard/api-server in memory with the tests

If we ever want to test stuff running on a remote cluster (swag) we'd have to rewrite all of this code.

Issue #97, opened 2020-03-12
Tye service APIs are broken


rynowak:
### Describe the bug

Tye service APIs are broken

### To Reproduce

Try to call `/api/v1/services` - it will throw on the server side because the types are no longer JSON serializable.

jkotalik:
I'm not seeing that issue locally with latest master bits:
![image](https://user-images.githubusercontent.com/8302101/77108265-1dfd8980-69df-11ea-8d8e-4d118a6b919b.png)

PR #154, opened 2020-03-20
Create new json objects for service models

jkotalik:
Fixes https://github.com/dotnet/tye/issues/97. Fixes https://github.com/dotnet/tye/issues/98.

I'm still using the OM to get the dashboard URL. Other notes inline.


src/Microsoft.Tye.Hosting/Model/JSON/ReplicaStatusJson.cs
```diff
@@ -0,0 +1,34 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Tye.Hosting.Model
+{
```
jkotalik: I'm not a fan of just appending JSON to the end of these names. Anything better?

src/Microsoft.Tye.Hosting/Model/ServiceDescription.cs
```diff
@@ -25,13 +25,7 @@ public ServiceDescription(string name, RunInfo? runInfo)
...
     {
-        public ConfigurationSource(string name, string value)
-        {
-            Name = name;
-            Value = value;
-        }
-
-        public string Name { get; }
-        public string Value { get; }
+        public string? Name { get; set; }
```
jkotalik: Organizationally, these are objects that are used by the JSON and non-JSON. I left them in place for now.

src/Microsoft.Tye.Hosting/Model/JSON/ServiceDescriptionJson.cs
```diff
@@ -0,0 +1,16 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Tye.Hosting.Model
+{
+    public class ServiceDescriptionJson
+    {
+        public string? Name { get; set; }
```
jkotalik: I realize I'm currently not serializing/deserializing RunInfo. I'll take a stab at it now, but it's a bit tricky with all the subclasses.

src/Microsoft.Tye.Hosting/Model/Application.cs
```diff
@@ -30,7 +30,7 @@ public void PopulateEnvironment(Service service, Action<string, string> set, str
                 // Inject normal configuration
                 foreach (var pair in service.Description.Configuration)
                 {
-                    set(pair.Name, pair.Value);
+                    set(pair.Name ?? "", pair.Value ?? "");
```
rynowak: Is this right? is a empty configuration value valid? or should we skip entries with null?

src/Microsoft.Tye.Hosting/Model/JSON/ReplicaStatusJson.cs
rynowak: I'd use a separate namespace and possibly prepend V1 to all of the things that are serialization DTOs.

src/Microsoft.Tye.Hosting/Model/Application.cs
jkotalik: Yeah I'll skip it, I don't think these will be null but might as well prevent it.

src/Microsoft.Tye.Hosting/Model/PortMapping.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.Tye.Hosting.Model
+{
+    public class PortMapping
+    {
+        public int ExternalPort { get; set; }
+
+        public List<int> InteralPorts { get; set; } = new List<int>();
```
rynowak: InternalPorts

src/Microsoft.Tye.Hosting/Model/JSON/ReplicaStatusJson.cs
jkotalik: Alright, I'll fully commit to it.

PR #155, opened 2020-03-20
Purging Hanging Replicas (On Startup and via Command)

areller:
This is copy + improvement on https://github.com/davidfowl/Micronetes/pull/48

If the host closes unexpectedly, it leaves processes and containers hanging and it might cause errors like this 

image

Or complain that a port is in use in subsequent runs.

To address this I register all `ReplicaState.Started` events in a file (TBD: Use Sqlite or some other established db) and delete (`docker rm` or `pkill`) the registered replicas whenever the host starts.

I also expose a command (`tye purge`) that does the same thing.


areller:
There seems to be an edge case when I use

```
using var tempDirectory = TempDirectory.Create();
```

In the tests.
Sometimes the dispose fails with `UnauthorizedAccessException`

I've wrapped the Dispose with try/catch for now

rynowak:
I'll take a look at this later tonite. @davidfowl - you should look too 😆 since you asked for it!

areller:
@rynowak @davidfowl Are you guys gonna look at it. Just want to know whether to fix the merge conflicts or not :)

src/Microsoft.Tye.Hosting/IReplicaInstantiator.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Collections.Generic;
```
rynowak: Needs copyright header (see other files in the repo)

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -294,6 +294,37 @@ private async Task StopContainerAsync(Tye.Hosting.Model.Service service)
...
+
+            if (container != "")
+            {
+                await ProcessUtil.RunAsync("docker", $"rm -f {container}",
+                    throwOnError: false,
+                    outputDataReceived: data => _logger.LogInformation("removed container {container} from previous run", container));
+            }
+        }
+
+        public ValueTask<IDictionary<string, string>> SerializeReplica(ReplicaEvent replicaEvent)
```
rynowak: Should this just be its own data type instead of piggy-backing on `ReplicaEvent`?

This doesn't roundtrip since we're just using the service name instead of the service definition.

src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -330,6 +330,39 @@ private static string GetExePath(string projectFilePath)
             return null;
         }
 
+        public Task HandleStaleReplica(ReplicaEvent replicaEvent)
+        {
+            var processStatus = (ProcessStatus)replicaEvent.Replica;
+            var pid = processStatus.Pid ?? -1;
+            if (pid != -1)
+                ProcessUtil.KillProcess(pid);
```
rynowak: braces (here and other places)

src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -330,6 +330,39 @@ private static string GetExePath(string projectFilePath)
             return null;
         }
 
+        public Task HandleStaleReplica(ReplicaEvent replicaEvent)
+        {
+            var processStatus = (ProcessStatus)replicaEvent.Replica;
+            var pid = processStatus.Pid ?? -1;
```
rynowak: nitpick  - there's no real advantage to using `-1` as a sentinel value instead of null. `-1` is what languages use when they don't have a nullable type 😆

test/E2ETest/DockerAssert.cs
```diff
@@ -76,6 +87,34 @@ void OnOutput(string text)
...
+                $"ps --format \"{{{{.ID}}}}\"",
+                stdOut: OnOutput,
+                stdErr: OnOutput);
+            if (exitCode != 0)
+            {
+                throw new XunitException($"Running `docker ps` failed." + Environment.NewLine + builder.ToString());
+            }
+
+            var lines = new HashSet<string>(builder.ToString().Split(new[] { '\r', '\n', }, StringSplitOptions.RemoveEmptyEntries));
+            if (predicate(lines, containers))
```
rynowak: braces

test/E2ETest/DockerAssert.cs
```diff
@@ -76,6 +87,34 @@ void OnOutput(string text)
             }
         }
 
+        private static async Task AssertAboutContainersAsync(ITestOutputHelper output, string[] containers, Func<HashSet<string>, string[], bool> predicate, Func<string[], string> errorText)
```
rynowak: Do we need all of this complex layering with the error message and predicate?

It seems like this could just be a method returns `IEnumerable<string>`of the containers, and then the validation and error message building wouldn't have to be packed into lambdas.

test/E2ETest/TyePurgeTests.cs
```diff
@@ -0,0 +1,159 @@
+﻿using System;
```
rynowak: Needs copyright header

test/E2ETest/TyePurgeTests.cs
```diff
@@ -0,0 +1,159 @@
...
+        [Fact]
+        public async Task FrontendBackendPurgeTest()
+        {
+            var projectDirectory = new DirectoryInfo(Path.Combine(TestHelpers.GetSolutionRootDirectory("tye"), "samples", "frontend-backend"));
+            var tempDirectory = TempDirectory.Create();
+            DirectoryCopy.Copy(projectDirectory.FullName, tempDirectory.DirectoryPath);
+
+            var projectFile = new FileInfo(Path.Combine(tempDirectory.DirectoryPath, "tye.yaml"));
+            var tyeDir = new DirectoryInfo(Path.Combine(tempDirectory.DirectoryPath, ".tye"));
+            using var host = new TyeHost(ConfigFactory.FromFile(projectFile).ToHostingApplication(), Array.Empty<string>())
```
rynowak: There will be some slight conflicts here due to my changes, but it should be copyable from other tests.

test/E2ETest/TyePurgeTests.cs
```diff
@@ -0,0 +1,159 @@
...
+
+                var pids = GetAllPids(host.Application);
+                var containers = GetAllContainerIds(host.Application);
+
+                Assert.True(Directory.Exists(tyeDir.FullName));
+                Assert.True(AllRunning(pids));
+                await DockerAssert.AssertAllContainersExistAsync(output, containers);
+
+                await host.PurgeAsync();
+                await Task.Delay(500);
```
rynowak: What's the reason for the wait? Generally if there are cases where you need to put in a wait, we'll eventually find a CI machine that is too slow 😆

test/E2ETest/TyePurgeTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            return ids;
+        }
+
+        private async Task WaitUntilSpawned(Microsoft.Tye.Hosting.Model.Application application)
+        {
+            int reties = 0;
+            do
+            {
+                if (reties > MaxRetries)
+                    throw new TimeoutException($"Could not reach response after {MaxRetries} reties");
```
rynowak: braces

test/E2ETest/TyePurgeTests.cs
```diff
@@ -0,0 +1,159 @@
...
+
+                await Task.Delay(500);
+                reties++;
```
rynowak: typo, should be retries

test/E2ETest/TyePurgeTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            finally
+            {
+                await host.StopAsync();
+            }
+
+            try
+            {
+                tempDirectory.Dispose();
+            }
+            catch (UnauthorizedAccessException) { }
```
rynowak: Usually the cause of this would be things like AV software.

test/E2ETest/DockerAssert.cs
```diff
@@ -46,6 +47,16 @@ void OnOutput(string text)
             }
         }
 
+        public static Task AssertAllContainersExistAsync(ITestOutputHelper output, string[] containers)
+            => AssertAboutContainersAsync(output, containers,
+                (existing, given) => given.All(c => existing.Contains(c)),
+                given => $"Not all containers ({string.Join(", ", given)}) exist");
```
rynowak: This would be better if it could print the list of containers that were still running, instead of the full set.

src/tye/Program.cs
```diff
@@ -14,7 +14,7 @@
 
 namespace Microsoft.Tye
 {
-    static partial class Program
+    public static partial class Program
```
rynowak: Is there a reason for this? I don't see `Program` referenced directly anywhere

src/tye/Program.PurgeCommand.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.CommandLine;
```
rynowak: Needs copyright header

src/tye/Program.PurgeCommand.cs
```diff
@@ -0,0 +1,34 @@
...
+            {
+                if (path is null)
+                {
+                    throw new CommandException("No project or solution file was found.");
+                }
+
+                var application = ConfigFactory.FromFile(path);
+
+                using var host = new TyeHost(application.ToHostingApplication(), args);
+                await host.PurgeAsync();
```
rynowak: One of the flaws with this approach is that it's based on reading the previous configuration, and assuming that it's the same, rather than just looking at the `./tye` directory as its source of truth.

That said, it's still an improvement

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            public async ValueTask<IList<StoreEvent>> GetEvents()
+            {
+                if (!File.Exists(_eventsFile))
+                {
+                    return Array.Empty<StoreEvent>();
+                }
+
+                var contents = await File.ReadAllTextAsync(_eventsFile);
+                var events = contents.Split(Environment.NewLine);
```
rynowak: Are the events forbidden from containing a newline?

rynowak: I suppose that's the case if you are using `WriteIndented = false` since newlines in the strings in JSON would be escaped already.

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
```diff
@@ -0,0 +1,126 @@
...
+                    .ToList();
+            }
+
+            public void Reset(bool delete, bool create)
+            {
+                if (delete && Directory.Exists(_tyeFolderPath))
+                    Directory.Delete(_tyeFolderPath, true);
+
+                if (create)
+                    Directory.CreateDirectory(_tyeFolderPath);
```
rynowak: braces

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
```diff
@@ -0,0 +1,126 @@
...
+            foreach (var @event in events)
+            {
+                if (_replicaInstantiators.TryGetValue(@event.ServiceType, out var instantiator))
+                {
+                    var replica = await instantiator.DeserializeReplicaEvent(@event.SerializedEvent);
+                    tasks.Add(instantiator.HandleStaleReplica(replica));
+                }
+            }
+
+            await Task.WhenAll(tasks);
```
rynowak: Should we log anything here?

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
```diff
@@ -0,0 +1,126 @@
...
+        public ReplicaRegistry(Model.Application application, IDictionary<ServiceType, IReplicaInstantiator> replicaInstantiators)
+        {
+            _replicaInstantiators = replicaInstantiators;
+            _store = new StateStore(application);
+        }
+
+        public async Task Reset()
+        {
+            await PurgeAll();
+            _store.Reset(true, true);
```
rynowak: nitpick: these cases would be much more comprehensible/readable with named parmaters. `Reset(delete: true, create: false)` is much more appealing 👍

src/Microsoft.Tye.Hosting/ReplicaStateRecorder.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System;
```
rynowak: Copyright

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
rynowak: Also do we need any error handling around this? Or is the idea that error handling is inside each of the processors?

src/Microsoft.Tye.Hosting/ReplicaStateRecorder.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        private void OnReplicaChanged(ReplicaEvent replicaEvent)
+        {
+            OnReplicaChangedAsync(replicaEvent).Wait(); // :(
+        }
+
+        private async Task OnReplicaChangedAsync(ReplicaEvent replicaEvent)
+        {
+            if (replicaEvent.State != ReplicaState.Started)
+                return;
```
rynowak: braces

src/Microsoft.Tye.Hosting/ReplicaStateRecorder.cs
```diff
@@ -0,0 +1,64 @@
...
+                }
+            }
+
+            _registry.Dispose();
+            return Task.CompletedTask;
+        }
+
+        private void OnReplicaChanged(ReplicaEvent replicaEvent)
+        {
+            OnReplicaChangedAsync(replicaEvent).Wait(); // :(
```
rynowak: `:( ` is correct 😆 

It's probably a better fix to make ` _registry.WriteReplicaEvent` synchronous. The disavantage of `Task.Wait` is that it uses two threads (the thread that waits, and the thread to does the underlying async operation). Just using `File.WriteAllText` for this case is going to be better.

The other option would be to fire-and-forget, but just being synchronous sounds better.

src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -59,6 +60,15 @@ public async Task RunAsync()
             await StopAsync();
         }
 
+        public Task PurgeAsync()
+        {
+            var app = BuildWebApplication(_application, _args, Sink);
+            var runners = CreateRunners(_application, _args, app.Logger, app.Configuration);
+
+            using var replicaRegistry = new ReplicaRegistry(_application, runners);
+            return replicaRegistry.Reset();
```
rynowak: meta-question:

All this work is done to preserve the layering that each kind of runner knows how to clean up it's own services, and that is expressed in the abstraction now.

As a result, we've got to read all of the configuration, and create the runners, and create a registry just so we can loop through the events and kill processes. 

I don't think it's terribly wrong, it's just different from what I expected when I think about "cleanup" of lingering services.

src/Microsoft.Tye.Hosting/DockerRunner.cs
areller: The service name of the PID if it's a process.
But yeah, most of the fields are not necessary.
I could 1) Just use a dictionary or 2) Use a string (that contains the container name or PID).
In both cases, the serialize and deserialize will become obsolete and you would only have a `HandleStaleReplica` method

test/E2ETest/TyePurgeTests.cs
areller: yeah. I wanted to avoid it but it seems that the kill of a process/container is asynchronous so without the delay the test would sometimes pass but sometimes won't
I can probably do the same thing that I've done with `WaitUntilSpawned` (e.g. `WaitUntilDespawned`)

test/E2ETest/TyePurgeTests.cs
areller: lol probably windows defender, since it works on my MacBook

src/tye/Program.PurgeCommand.cs
areller: It's basically just ceremonial since it ignores the configuration anyway and just looks at the file in the `./tye` dir.
But I agree that it's not ideal.

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
areller: Yeah. The `WriteIndented = false` should prevent line breaks and other indentations.

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
areller: Yeah, the processors already catch the relevant exception.
I don't know which exceptions to expect at this level since every processor has a different approach for destroying his stale replicas.

test/E2ETest/TyePurgeTests.cs
areller: The tests also fail for other reasons in rare cases. For example, `WaitUntilSpawned` waits until the replica count equals to the replica count in the definition (for all services), but it doesn't wait until all replicas are of `Running` state, because the state is not saved in the ReplicaStatus class.
Maybe I should rethink my testing strategy. Maybe go with what was done in the other test classes (creating an HttpClient that interacts with the services directly)

davidfowl:
@rynowak yo think it’s cleaner to pass a storage interface to all of the runners to save and restore?

src/tye/Program.PurgeCommand.cs
rynowak: OK. Let's consider this resolved then for now, this is still an improvment.

test/E2ETest/TestHelpers.cs
```diff
@@ -33,5 +43,91 @@ public static string GetSolutionRootDirectory(string solution)
 
             throw new Exception($"Solution file {solution}.sln could not be found in {applicationBasePath} or its parent directories.");
         }
+
+        public static async Task StartHostAndWaitForReplicasToStart(TyeHost host)
+        {
+            var startedTask = new TaskCompletionSource<bool>();
+            var alreadyStarted = 0;
+            var totalReplicas = host.Application.Services.Sum(s => s.Value.Description.Replicas);
```
rynowak: It seems like there's a logic error here - shouldn't the success condition be `totalReplicas - alreadStarted == 0`?

test/E2ETest/TestHelpers.cs
```diff
@@ -33,5 +43,91 @@ public static string GetSolutionRootDirectory(string solution)
...
+            var alreadyStopped = replicas.Length;
+
+            void OnReplicaChange(ReplicaEvent ev)
+            {
+                if (replicas.Contains(ev.Replica.Name) && ev.State == ReplicaState.Stopped)
+                {
+                    Interlocked.Decrement(ref alreadyStopped);
+                }
+
+                if (alreadyStopped == 0)
```
rynowak: The name of this seems misleading, like it should be `remaining`

test/E2ETest/TestHelpers.cs
```diff
@@ -33,5 +43,91 @@ public static string GetSolutionRootDirectory(string solution)
 
             throw new Exception($"Solution file {solution}.sln could not be found in {applicationBasePath} or its parent directories.");
         }
+
+        public static async Task StartHostAndWaitForReplicasToStart(TyeHost host)
```
rynowak: I like the approach used here better than polling with random HTTP clients and retries 👍

test/E2ETest/TyePurgeTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                Sink = sink,
+            };
+
+            await TestHelpers.StartHostAndWaitForReplicasToStart(host);
+            try
+            {
+                var pids = GetAllPids(host.Application);
+
+                Assert.True(Directory.Exists(tyeDir.FullName));
```
rynowak: Can we also verify that the directory gets deleted

src/Microsoft.Tye.Hosting/DockerRunner.cs
rynowak: It could also just be its own type. It's a little strange to use `ReplicaEvent` because you're not going to rehydrate all of its fields, just the basic stuff like the PID. I think either using a dictionary or using a new type would both be improvments.

rynowak:
> @rynowak yo think it’s cleaner to pass a storage interface to all of the runners to save and restore?

Maybe you want to make that suggestion to @areller since this is his PR :)

rynowak:
But yes, that might be a better idea.

areller:
@rynowak @davidfowl I don't mind calling the `WriteEvent` function directly from the runners if you want, but I'm not sure there is a benefit to it.
The benefit of doing it via listening to the `ReplicaEvents` is that it saves you maintenance in the case you want to edit/refactor the runners or in the case you want to add more tasks that should run when a replica is started/stopped. 

Here's a funny story.
I've spent over an hour trying to find a race condition that caused the purge tests to fail sometimes.

Turns out that it was partly due to this

```
using var replicaRegistry = new ReplicaRegistry(_application, app.Logger, runners);
return replicaRegistry.Delete();
```

Instead of awaiting `Delete()` I used to return the task directly, but it meant that `replicaRegistry` got cleaned before `Delete()` was completed which triggered some other race condition.  

Be careful with those new C# 8 features :laughing:

src/tye/Program.PurgeCommand.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.CommandLine.Invocation;
+using System.IO;
+using Microsoft.Tye.ConfigModel;
+using Microsoft.Tye.Hosting;
+
+namespace Microsoft.Tye
+{
+    static partial class Program
+    {
+        public static Command CreatePurgeCommand(string[] args)
```
davidfowl: I'm not a fan of this command? Or maybe I'm just not a fan of the fact that we need to call into the runner to to a purge because the logic is there.

rynowak: I think it's the latter. The other question is whether we'd auto-purge if you do run again.

areller: @davidfowl It sounds reasonable that the entity which knows how to launch a certain type of replica would know how to kill replicas of that type.  
I don't think I like the idea of having a class like `ReplicaKiller` that switch/cases through the types of replicas.

The best alternative that I can think of is, you can have a virtual `Kill` method in the `ReplicaStatus` class (`ProcessStatus` and `DockerStatus` provide their own implementation)

Then, the `WriteReplicaEvent` in `ReplicaRegistry` would serialize the entire replica (as JSON?) together with its type.
`GetEvents` will return a list of `ReplicaStatus` (instead of list of `StoreEvent`) and the purge method will loop over all `ReplicaStatus`es and call `.Kill()`

How does it sound?

davidfowl: Maybe a cleaner design is just to pass the registry to the docker runner and process runner. I recall me guiding you this direction but it seems like it might be much simpler now to delete the state when replicas are spun down in both the docker and process runner.

areller: @davidfowl  I could do it but how does it address your initial concern about the purge command or the fact that the runners contain the purge logic? Unless I'm misunderstanding something.

davidfowl: Lets stick to auto purge on run and remove the purge command for now.

areller: @davidfowl Sure thing. I've removed the purge command but for now I've left the `PurgeAsyc` method in the host.
This is because the purging logic is in the runners and the host has access to create the runners, unless we change something structurally.
This can also be changed if I implement this
```
The best alternative that I can think of is, you can have a virtual Kill method in the ReplicaStatus class (ProcessStatus and DockerStatus provide their own implementation)
```

By the way, I've made this modification
```
while (!dockerInfo.StoppingTokenSource.Token.IsCancellationRequested)
{
    var logsRes = await ProcessUtil.RunAsync("docker", $"logs -f {containerId}",
        outputDataReceived: data => service.Logs.OnNext($"[{replica}]: {data}"),
        errorDataReceived: data => service.Logs.OnNext($"[{replica}]: {data}"),
        throwOnError: false,
        cancellationToken: dockerInfo.StoppingTokenSource.Token);

    if (logsRes.ExitCode != 0)
    {
        break;
    }
}
```
to https://github.com/dotnet/tye/pull/209 (added a break from the loop if the exit code is non zero)

Turns out that if you kill the container externally, it enters an infinite loop.
It caused my tests to fail but I also think that it's a good check anyway.
What do you think?

src/Microsoft.Tye.Hosting/ReplicaStateRecorder.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Tye.Hosting.Model;
+
+namespace Microsoft.Tye.Hosting
+{
+    public class ReplicaStateRecorder : IApplicationProcessor
```
davidfowl: Lets remove this for now and move the logic back into the process and docker runners.

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
```diff
@@ -0,0 +1,151 @@
...
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.Tye.Hosting.Model;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Tye.Hosting
+{
+    public class ReplicaRegistry : IDisposable
```
davidfowl: Lets pass this directly to the runners.

src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -67,6 +67,25 @@ public async Task RunAsync()
             await StopAsync();
         }
 
+        public async Task PurgeAsync()
```
davidfowl: Remove this.

src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -246,32 +265,48 @@ private static bool IsPortInUseByBinding(Application application, int port)
             return false;
         }
 
-        private static AggregateApplicationProcessor CreateApplicationProcessor(string[] args, string[] servicesToDebug, Microsoft.Extensions.Logging.ILogger logger, IConfiguration configuration)
+        private static AggregateApplicationProcessor CreateApplicationProcessor(Application application, string[] args, string[] servicesToDebug, Microsoft.Extensions.Logging.ILogger logger, IConfiguration configuration)
```
davidfowl: This can be done.

src/tye/Program.PurgeCommand.cs
davidfowl: - There should be more purge command

> Turns out that if you kill the container externally, it enters an infinite loop.
It caused my tests to fail but I also think that it's a good check anyway.
What do you think?

This seems fine if it normally doesn't infinitely loop when logs runs and ends.

src/Microsoft.Tye.Hosting/TyeHost.cs
davidfowl: I meant undone 😬

davidfowl:
Sweet!

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
```diff
@@ -0,0 +1,104 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.Tye.Hosting.Model;
+using Microsoft.Extensions.Logging;
+using System.Threading;
+using System.Collections.Concurrent;
```
davidfowl: nit: Sort usings

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
```diff
@@ -0,0 +1,104 @@
...
+
+namespace Microsoft.Tye.Hosting
+{
+    public class ReplicaRegistry : IDisposable
+    {
+        private readonly ILogger _logger;
+        private readonly ConcurrentDictionary<string, SemaphoreSlim> _fileWriteSemaphores;
+        private readonly string _tyeFolderPath;
+
+        public ReplicaRegistry(Model.Application application, ILogger logger)
```
davidfowl: Directly pass in the path, not the application.

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
```diff
@@ -0,0 +1,104 @@
...
+using Microsoft.Extensions.Logging;
+using System.Threading;
+using System.Collections.Concurrent;
+
+namespace Microsoft.Tye.Hosting
+{
+    public class ReplicaRegistry : IDisposable
+    {
+        private readonly ILogger _logger;
+        private readonly ConcurrentDictionary<string, SemaphoreSlim> _fileWriteSemaphores;
```
davidfowl: Make these normal locks, not semaphores.

src/Microsoft.Tye.Hosting/ReplicaRegistry.cs
```diff
@@ -0,0 +1,104 @@
...
+            }
+        }
+
+        public bool WriteReplicaEvent(string storeName, IDictionary<string, string> replicaRecord)
+        {
+            var filePath = Path.Join(_tyeFolderPath, GetStoreFile(storeName));
+            var contents = JsonSerializer.Serialize(replicaRecord, new JsonSerializerOptions { WriteIndented = false });
+            var semaphore = GetSempahoreForStore(storeName);
+
+            semaphore.Wait();
```
davidfowl: Why a semaphore vs a lock? Lets make this a regular lock object.

test/E2ETest/TestHelpers.cs
davidfowl: It's counting up to totalReplicas and these description.replicas is the desired state, it's correct.

test/E2ETest/TestHelpers.cs
```diff
@@ -64,5 +74,104 @@ internal static TempDirectory CopyTestProjectDirectory(string projectName)
             DirectoryCopy.Copy(GetTestProjectDirectory(projectName).FullName, temp.DirectoryPath);
             return temp;
         }
+
+        public static async Task StartHostAndWaitForReplicasToStart(TyeHost host)
+        {
+            var startedTask = new TaskCompletionSource<bool>();
```
davidfowl: This should pass TaskCreationOptions.RunContinuationsAsynchronously

PR #160, opened 2020-03-20
Renamed dockerImage to image

### LINKED ISSUE #122: `dockerImage` should be just `image`
I don't want to leave it as `dockerImage` I am proposing we just call it image.

davidfowl:
- Added a new image field and marked dockerImage as deprecated. Will remove in a future version.
- Fixes #122

### DIFF
--- a/docs/redis.md
+++ b/docs/redis.md
@@ -90,11 +90,11 @@ We just showed how `tye` makes it easier to communicate between 2 applications r
     - name: frontend
       project: frontend\frontend.csproj
     - name: redis
-      dockerImage: redis
+      image: redis
       bindings:
       - port: 6379
     - name: redis-cli
-      dockerImage: redis
+      image: redis
       args: "redis-cli -h host.docker.internal MONITOR"
    ```
 
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -21,7 +21,7 @@ services:
 - name: worker
   project: worker/worker.csproj
 - name: rabbit
-  dockerImage: rabbitmq:3-management
+  image: rabbitmq:3-management
   bindings:
     - port: 5672
       protocol: rabbitmq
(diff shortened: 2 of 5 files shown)


src/tye/ConfigModel/ConfigService.cs
```diff
@@ -14,6 +14,7 @@ public class ConfigService
         public string Name { get; set; } = default!;
         public bool External { get; set; }
         public string? DockerImage { get; set; }
```
davidfowl: I was looking for a clean place to log this as a warning deprecated but I just left it (mostly because I was being lazy).

jkotalik:
> Added a new image field and marked dockerImage as deprecated. Will remove in a future version.
 
I really think we should just make the breaking change now and completely drop dockerImage for image if we can't write a deprecation message.

davidfowl:
> I really think we should just make the breaking change now and completely drop dockerImage for image if we can't write a deprecation message.

Done. Have broken all of our loyal customers. All 3 of us

docs/schema.md
```diff
@@ -118,11 +118,11 @@ Including a `project` entry marks the service as a *project*:
 - It will build and run locally using the .NET project during development. 
 - It will be packaged and deployed during deployments.
 
-#### `dockerImage` (string)
```
rynowak: The docs should call out that this might still be `dockerImage` depending on your build. For instance our instructions point at a version number that will use `dockerImage`.

docs/redis.md
```diff
@@ -90,11 +90,11 @@ We just showed how `tye` makes it easier to communicate between 2 applications r
     - name: frontend
       project: frontend\frontend.csproj
     - name: redis
-      dockerImage: redis
```
rynowak: samples folder should be updated

docs/schema.md
davidfowl: Yes this is my fear with this change. What do we want to do here?

docs/redis.md
davidfowl: Fixed

PR #161, opened 2020-03-20
Don't assume executables are relative to the context directory.

davidfowl:
- This allows running executables that are on the path.

```yaml
services:
- name: myservice
  executable: node
  args: app.js
```


jkotalik:
I thought I got all of these (there were a lot of places we assumed this) but guess I missed this one 😄

Issue #159, opened 2020-03-20
Possibility to debug only a specific service / multiple services


btastic:
### Some details

**How many services are in your application? (including things like databases, redis):**

- [ ] 1-2
- [x] 3-5
- [ ] 6-10
- [ ] 10+

**What did you use tye for?**

- [x] Local development
- [ ] Deployment

**How is your code organized?**

- [x] One service per repo
- [ ] Many services in one big repo
- [ ] Other (please explain)

### What did you think of tye?
Right now you can only use `--debug` to make all services wait for a debugger to be attached. When working on multiple services, we usually only change things in one or two services at the same time, and don't need a debugger on every instance.

It would be nice to specify which services I want to debug.

`tye run --debug service1,service2,service3` is something I could think of.

jkotalik:
Thanks for the feedback.

We definitely should support specifying specific services to debug locally (and potentially even remotely 😱 ).

rynowak:
@btastic - did using Tye with one service per repo cause you any trouble?

<img width="302" alt="image" src="https://user-images.githubusercontent.com/1430011/77192319-9a54a300-6a99-11ea-8613-d642c97565f5.png">

How did you end up structuring things?

Like this?

```txt
/somefolder/tye.yaml
/somefolder/repo1/...
/somefolder/repo2/...
/somefolder/repo3/...
```

btastic:
I created a new sln file containing all projects I needed. Then i ran `tye run project.sln`

btastic:
Projects were all somewhat scattered in my repo folder. So Something Like C:\repos\repo1 repo2 repo3 and the repos folder has a sln which references projects from repo1 repo2 and so on.

rynowak:
ok cool. Since you keep them in separate repos, do you have any ideas about what you think would be ideal?

btastic:
If we're talking ideal - make it so it is activatable in the dashboard UI. If we're talking what I can live with right now: Just a simple command line argument like I mentioned in the beginning.. eg

`tye run --debug service1,service2,service3`

btastic:
There is now a NRE when not debugging any services. ProcessRunnerOptions.cs:24: `DebugAllServices = servicesToDebug.Contains("*", StringComparer.OrdinalIgnoreCase)`

if servicesToDebug is null, which will be if you don't specify any, this leads to a NullReferenceException.

btastic:
Other than that, it is working great :)

jkotalik:
[...]

PR #163, opened 2020-03-20
Allow debugging specific projects

jkotalik:
For part of https://github.com/dotnet/tye/issues/159

### DIFF
--- a/src/Microsoft.Tye.Hosting/ProcessRunnerOptions.cs
+++ b/src/Microsoft.Tye.Hosting/ProcessRunnerOptions.cs
@@ -11,19 +11,17 @@ namespace Microsoft.Tye.Hosting
     {
         public bool DebugMode { get; set; }
         public bool BuildProjects { get; set; }
-        public string[]? ProjectsToDebug { get; set; }
-        public bool DebugAllProjects { get; set; }
+        public string[]? ServicesToDebug { get; set; }
+        public bool DebugAllServices { get; set; }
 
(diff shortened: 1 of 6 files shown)


davidfowl:
Projects or services?

jkotalik:
What do you prefer? I used projects because we don't explicitly debug other services (like redis).

src/tye/Program.RunCommand.cs
```diff
@@ -51,7 +51,8 @@ private static Command CreateRunCommand(string[] args)
 
             command.AddOption(new Option("--debug")
             {
-                Description = "Wait for debugger attach in all services.",
+                Argument = new Argument<string>("services|\"all\""),
```
rynowak: I'd suggest making this multiple arguments instead of comma-delimited. 

```
--debug foo --debug bar
```

We have the technology. We also might want to use `*` as a sentinel instead of `all` since I GUESS `all` is a valid service name.

davidfowl:
I like services

src/Microsoft.Tye.Hosting/ProcessRunnerOptions.cs
```diff
@@ -11,13 +11,19 @@ public class ProcessRunnerOptions
     {
         public bool DebugMode { get; set; }
         public bool BuildProjects { get; set; }
+        public string[]? ProjectsToDebug { get; set; }
```
davidfowl: ServicesToDebug

PR #165, opened 2020-03-21
Clean up port injection

davidfowl:
- We inject ASPNETCORE_URLS, HTTPS_PORT and a PORTS env variable for binding.

### DIFF
--- a/src/Microsoft.Tye.Core/CombineStep.cs
+++ b/src/Microsoft.Tye.Core/CombineStep.cs
@@ -28,6 +28,7 @@ namespace Microsoft.Tye
                     var port = binding.Port ?? 80;
                     var urls = $"http://*{(port == 80 ? "" : (":" + port.ToString()))}";
                     service.Service.Environment.Add("ASPNETCORE_URLS", urls);
+                    service.Service.Environment.Add("PORT", port);
                     break;
                 }
             }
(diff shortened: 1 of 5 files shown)


src/Microsoft.Tye.Core/CombineStep.cs
```diff
@@ -28,6 +28,7 @@ public override Task ExecuteAsync(OutputContext output, Application application,
                     var port = binding.Port ?? 80;
                     var urls = $"http://*{(port == 80 ? "" : (":" + port.ToString()))}";
                     service.Service.Environment.Add("ASPNETCORE_URLS", urls);
+                    service.Service.Environment.Add("PORT", port);
```
rynowak: Nice. Is the idea that this is the way to get the listening port for any non-asp.net tech?

src/Microsoft.Tye.Hosting/ProxyService.cs
```diff
@@ -69,6 +69,15 @@ static int GetNextPort()
                                         binding.Port = GetNextPort();
                                     }
 
+                                    if (binding.Protocol == "http" || binding.Protocol == null)
+                                    {
+                                        binding.InternalPort = 80;
+                                    }
+                                    else if (binding.Protocol == "https")
+                                    {
+                                        binding.InternalPort = 443;
```
rynowak: Can you add internal port to the docs when you get a chance? I'm not 💯 on what it does, and so it's not there.

src/Microsoft.Tye.Core/CombineStep.cs
davidfowl: Yap

src/Microsoft.Tye.Hosting/ProxyService.cs
davidfowl: It’s possible the same is bad as well. It’s really the container port.

PR #168, opened 2020-03-22
Minor white-space, typo, structured logging, and consistent usage of 'Dockerfile' changes

ChadJessup:
Minor white-space, typo, structured logging, and consistent usage of 'Dockerfile' (and multi/single-phase) pass.

Also, addressed a couple typos in the CLI output in the help text.

### DIFF
--- a/src/Microsoft.Tye.Core/DockerContainerBuilder.cs
+++ b/src/Microsoft.Tye.Core/DockerContainerBuilder.cs
@@ -44,7 +44,7 @@ namespace Microsoft.Tye
             var dockerFilePath = Path.Combine(application.GetProjectDirectory(project), Path.GetDirectoryName(project.RelativeFilePath)!, "Dockerfile");
             if (File.Exists(dockerFilePath))
             {
-                output.WriteDebugLine($"Using existing dockerfile '{dockerFilePath}'.");
+                output.WriteDebugLine($"Using existing Dockerfile '{dockerFilePath}'.");
             }
             else
             {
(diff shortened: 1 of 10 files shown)


src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -192,13 +192,13 @@ private int ComputePort(WebApplication app)
             else if (IsPortInUseByBinding(_application, DefaultPort))
             {
                 // Port has been reserved for the app.
-                app.Logger.LogInformation($"Default dashboard port {DefaultPort} has been reserved by the application, choosing random port.");
+                app.Logger.LogInformation("Default dashboard port {DefaultPort} has been reserved by the application, choosing random port.", DefaultPort);
```
ChadJessup: Unsure if structure logging was intended here - but majority of code-base uses it, so I updated this to match the majority...

src/Microsoft.Tye.Core/DockerContainerBuilder.cs
```diff
@@ -44,15 +44,15 @@ public static async Task BuildContainerImageAsync(OutputContext output, Applicat
             var dockerFilePath = Path.Combine(application.GetProjectDirectory(project), Path.GetDirectoryName(project.RelativeFilePath)!, "Dockerfile");
             if (File.Exists(dockerFilePath))
             {
-                output.WriteDebugLine($"Using existing dockerfile '{dockerFilePath}'.");
+                output.WriteDebugLine($"Using existing Dockerfile '{dockerFilePath}'.");
```
ChadJessup: Casing in `Dockerfile` is often important, due to it being a default file with the docker CLI, and since docker is used in case-sensitive file systems regularly - I figured being consistent inside and out of the code is important.

src/Microsoft.Tye.Hosting/TyeHost.cs
rynowak: Yup. Right now there's a mix of two different logging systems at play here, but the stuff using `ILogger`/Serilog should deffo be structured logs.

rynowak:
Thanks!

PR #169, opened 2020-03-22
Use MSBuild in process to grovel project information

rynowak:
This replaces the (dirty) technique of target-based injection with
hosting MSBuild in process.

This is a speedup of about 800ms per project for every project past the first
one. Given that Tye is for microservices, hopefully you have more than one
project.


src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -30,94 +39,136 @@ public static async Task ReadProjectDetailsAsync(OutputContext output, FileInfo
...
-                        continue;
-                    }
-
-                    if (line.StartsWith("frameworks="))
-                    {
-                        var right = line.Substring("frameworks=".Length).Trim();
-                        project.Frameworks.AddRange(right.Split(",").Select(s => new Framework(s)));
-                        output.WriteDebugLine($"Found shared frameworks: {line}");
-                        continue;
+                        return Assembly.LoadFrom(assemblyFilePath);
```
davidfowl: Any reason to avoid using context.LoadFromAssemblyPath? It won't load deps but you'll likely be called for those anyways?

src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -30,94 +39,136 @@ public static async Task ReadProjectDetailsAsync(OutputContext output, FileInfo
...
+            {
+                // It says VisualStudio - but we'll just use .NET SDK
+                var instances = MSBuildLocator.QueryVisualStudioInstances(new VisualStudioInstanceQueryOptions()
+                {
+                    DiscoveryTypes = DiscoveryType.DotNetSdk,
+
+                    // Using the project as the working directory. We're making the assumption that
+                    // all of the projects want to use the same SDK version. This library is going
+                    // load a single version of the SDK's assemblies into our process, so we can't
+                    // use supprt SDKs at once without getting really tricky.
```
jkotalik: Comment confuses me at the end. Is the point that we can't use multiple sdks in the same process without getting really tricky?

src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -30,94 +39,136 @@ public static async Task ReadProjectDetailsAsync(OutputContext output, FileInfo
...
+                    //
+                    // The .NET SDK-based discovery uses `dotnet --info` and returns the SDK
+                    // in use for the directory.
+                    //
+                    // https://github.com/microsoft/MSBuildLocator/blob/master/src/MSBuildLocator/MSBuildLocator.cs#L320
+                    WorkingDirectory = projectFile.DirectoryName,
+                });
+
+                var instance = instances.SingleOrDefault();
```
jkotalik: Will this ever return more than one? Should we do `FirstOrDefault` instead?

src/Microsoft.Tye.Core/ProjectReader.cs
rynowak: Correct.

src/Microsoft.Tye.Core/ProjectReader.cs
rynowak: It will always return one. It does `dotnet --info` and then returns the SDK would be used if you ran `dotnet`. If it starts to return multiple, we want to know because that would cause bugs - we'd need to know how to choose between different versions.

src/Microsoft.Tye.Core/ProjectReader.cs
rynowak: I'll try and it and see. That's a bit more primitive so I don't mind

rynowak: ya it works 👍

rynowak:
AZDO seems like it's not able to create windows jobs right now. #YOLO

src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -30,94 +39,136 @@ public static async Task ReadProjectDetailsAsync(OutputContext output, FileInfo
...
-                    stdErr: capture.StdErr);
-
-                output.WriteDebugLine($"Done running 'dotnet msbuild' exit code: {exitCode}");
-                if (exitCode != 0)
-                {
-                    throw new CommandException("'dotnet msbuild' failed.");
-                }
+                AssemblyLoadContext.Default.Resolving += ResolveAssembly;
+                var result = projectInstance.Build(
+                    targets: new[] { "Restore", "ResolveReferences", "ResolvePackageDependenciesDesignTime", "PrepareResources", "GetAssemblyAttributes", },
```
rainersigwald: Running restore in the same operation as other things might cause some problems here. MSBuild 1) evaluates a project and then 2) runs targets on that evaluated project. When `Restore` happens, it may cause additional MSBuild logic to exist in places that would affect a subsequent evaluation--but it won't affect the current instance of the project (because that was already evaluated).

Looking at this list of targets, the concrete problems I'd expect are:
* Missing resources if they're pulled in through a package, and
* Incorrect assembly attributes if they're modified through a package (for example versioning through GitVersion or something).

MSBuild.exe implements `-restore` with 

https://github.com/microsoft/msbuild/blob/3453beee039fb6f5ccc54ac783ebeced31fec472/src/MSBuild/XMake.cs#L1427-L1438

It might be best to copy that approach here.

src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -30,94 +39,136 @@ public static async Task ReadProjectDetailsAsync(OutputContext output, FileInfo
...
+            // The Microsoft.Build.Locator doesn't handle the loading of other assemblies
+            // that are shipped with MSBuild (ex NuGet).
+            //
+            // This means that the set of assemblies that need special handling depends on the targets
+            // that we run :(
+            //
+            // This is workaround for this limitation based on the targets we need to run
+            // to resolve references and versions.
+            //
+            // See: https://github.com/microsoft/MSBuildLocator/issues/86
```
rainersigwald: :heart: (Yes, we should do this for you)

src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -30,94 +39,136 @@ public static async Task ReadProjectDetailsAsync(OutputContext output, FileInfo
...
             }
-            catch (Exception ex)
+            catch
             {
-                throw new CommandException("Failed to install targets.", ex);
+                throw new CommandException($"Failed to load project: '{projectFile.FullName}'.");
             }
 
-            var outputFilePath = Path.GetTempFileName();
+            // Currently we only log at debug level.
```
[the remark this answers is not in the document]
rynowak: Cool beans 👍 Good to know that we can do that here.

src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -30,94 +39,136 @@ public static async Task ReadProjectDetailsAsync(OutputContext output, FileInfo
...
+            var logger = new ConsoleLogger(
+                verbosity: LoggerVerbosity.Normal,
+                write: message => output.WriteDebug(message),
+                colorSet: null,
+                colorReset: null);
```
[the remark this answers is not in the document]
rynowak: It's in MSBuild, I'm a little surprised 😆

src/Microsoft.Tye.Core/ProjectReader.cs
rynowak: Cool, let us know if/when that happens.

src/Microsoft.Tye.Core/ProjectReader.cs
rynowak: We've already hit some of these issues, I'm going to try this 👍

PR #171, opened 2020-03-22
Added sample showing how to use nginx as ingress

davidfowl:
- This samples shows how to use tye with nginx locally to serve
multiple applications based on path.
- There are a few gross things that coud be make better
(like service discovery) within the nginx container itself.

### DIFF
--- a/samples/apps-with-ingress/nginx.conf
+++ b/samples/apps-with-ingress/nginx.conf
@@ -3,9 +3,23 @@ server {
 
     location /A {
         proxy_pass http://host.docker.internal:5005/;
+        proxy_http_version 1.1;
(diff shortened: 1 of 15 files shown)


samples/apps-with-ingress/nginx.conf
```diff
@@ -0,0 +1,11 @@
+server {
+    listen 80;
+
+    location /A {
+        proxy_pass http://host.docker.internal:5005/;
```
davidfowl: The nginx config doesn't support env variables so we hard code this for now (which is breaks if we change the execution mode). We'll need to make file transformations like this more first class for these scenarios.

davidfowl:
Closing in favor of https://github.com/dotnet/tye/pull/172

PR #172, opened 2020-03-22
Added ingress support in local orchestrator

davidfowl:
- Supports host and port mapping to other services
- Added sample
- The proxy only supports round robin load balancing.

Ingress is just another service with a set of rules. Rules determine what the host and port are and what service to route to as a result. See the example below.

```yaml
name: apps-with-ingress
ingress:
  - name: ingress
    bindings:
      - port: 8080
    rules:
      - path: /A
        service: appA
      - path: /B
        service: appB
      - host: a.example.com
        service: appA
      - host: b.example.com
        service: appB  

services:
- name: appA
  project: ApplicationA/ApplicationA.csproj
  replicas: 2
- name: appB
  project: ApplicationB/ApplicationB.csproj
  replicas: 2
```

This uses ASP.NET Core to implement the proxy.

~~TODO:  Add test for the sample~~

### DIFF
--- a/samples/apps-with-ingress/ApplicationB/Startup.cs
+++ b/samples/apps-with-ingress/ApplicationB/Startup.cs
@@ -1,0 +1,40 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace ApplicationB
+{
+    public class Startup
+    {
+        // This method gets called by the runtime. Use this method to add services to the container.
+        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
+        public void ConfigureServices(IServiceCollection services)
+        {
+        }
+
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+        {
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapGet("/", async context =>
+                {
+                    await context.Response.WriteAsync("Hello from Application B " + Environment.GetEnvironmentVariable("APP_INSTANCE") ?? Environment.GetEnvironmentVariable("HOSTNAME"));
+                });
+            });
+        }
+    }
+}
--- a/src/Microsoft.Tye.Hosting/Dashboard/Pages/Logs.razor
+++ b/src/Microsoft.Tye.Hosting/Dashboard/Pages/Logs.razor
@@ -13,7 +13,7 @@ else
     <div style="overflow-y: scroll;position: absolute;height: 84%; width:75%; color:white;background-color:black;padding:10px">
         @foreach (var log in ApplicationLogs)
         {
-            <div @key="@log.Id" style="width:100%">@log.Text</div>
+            <div style="width:100%">@log.Text</div>
         }
     </div>
 }
--- a/src/Microsoft.Tye.Hosting/Infrastructure/IngressHostMatcherPolicy.cs
+++ b/src/Microsoft.Tye.Hosting/Infrastructure/IngressHostMatcherPolicy.cs
@@ -1,0 +1,209 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Matching
+{
+    internal sealed class IngressHostMatcherPolicy : MatcherPolicy, IEndpointSelectorPolicy
+    {
+        private const string WildcardHost = "*";
+        private const string WildcardPrefix = "*.";
+
+        // Run after HTTP methods, but before 'default'.
+        public override int Order { get; } = -100;
+
+        public bool AppliesToEndpoints(IReadOnlyList<Endpoint> endpoints)
+        {
+            return endpoints.Any(e =>
+            {
+                var hosts = e.Metadata.GetMetadata<IngressHostMetadata>()?.Hosts;
+                if (hosts == null || hosts.Count == 0)
+                {
+                    return false;
+                }
+
+                foreach (var host in hosts)
+                {
+                    // Don't run policy on endpoints that match everything
+                    var key = CreateEdgeKey(host);
+                    if (!key.MatchesAll)
+                    {
+                        return true;
+                    }
+                }
+
+                return false;
+            });
+        }
+
+        public Task ApplyAsync(HttpContext httpContext, CandidateSet candidates)
+        {
+            if (httpContext == null)
+            {
+                throw new ArgumentNullException(nameof(httpContext));
+            }
+
+            if (candidates == null)
+            {
+                throw new ArgumentNullException(nameof(candidates));
+            }
+
+            for (var i = 0; i < candidates.Count; i++)
+            {
+                if (!candidates.IsValidCandidate(i))
+                {
+                    continue;
+                }
+
+                var hosts = candidates[i].Endpoint.Metadata.GetMetadata<IngressHostMetadata>()?.Hosts;
+                if (hosts == null || hosts.Count == 0)
+                {
+                    // Can match any host.
+                    continue;
+                }
+
+                var matched = false;
+                var (requestHost, requestPort) = GetHostAndPort(httpContext);
+                for (var j = 0; j < hosts.Count; j++)
+                {
+                    var host = hosts[j].AsSpan();
+                    var port = ReadOnlySpan<char>.Empty;
+
+                    // Split into host and port
+                    var pivot = host.IndexOf(':');
+                    if (pivot >= 0)
+                    {
+                        port = host.Slice(pivot + 1);
+                        host = host.Slice(0, pivot);
+                    }
+
+                    if (host == null || MemoryExtensions.Equals(host, WildcardHost, StringComparison.OrdinalIgnoreCase))
+                    {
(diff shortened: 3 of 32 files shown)


src/Microsoft.Tye.Hosting/ProxyExtensions.cs
```diff
@@ -0,0 +1,213 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
davidfowl: This comes from https://github.com/aspnet/AspLabs/blob/85ff16588fe970d738aa51ff3ba1c675bc678f28/src/Proxy/src/ProxyAdvancedExtensions.cs

samples/apps-with-ingress/ApplicationB/Startup.cs
```diff
@@ -0,0 +1,40 @@
...
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapGet("/", async context =>
+                {
+                    await context.Response.WriteAsync("Hello from Application B");
```
jkotalik: Would be nice to also have replica information here as well in the sample.

davidfowl: That'll make my test more complicated 😄

davidfowl: Done

rynowak:
> Ingress is just another service with a set of rules. Rules determine what the host and port are and what service to route to as a result. See the example below.

This really doesn't seem right. As this is written we can't really validate it, provide a schema for it, or avoid having to check for it for all services.

src/Microsoft.Tye.Hosting/Dashboard/Pages/Logs.razor
```diff
@@ -13,7 +13,7 @@ else
     <div style="overflow-y: scroll;position: absolute;height: 84%; width:75%; color:white;background-color:black;padding:10px">
         @foreach (var log in ApplicationLogs)
         {
-            <div @key="@log.Id" style="width:100%">@log.Text</div>
+            <div style="width:100%">@log.Text</div>
```
rynowak: Was this causing a crash?

davidfowl: Yah it’s generating duplicate ids

davidfowl:
> This really doesn't seem right. As this is written we can't really validate it, provide a schema for it, or avoid having to check for it for all services.

I don’t mind changing where it goes and I don’t even have a good idea of what it means to deploy this. I do know that there still needs to be a service that acts like the controller that reads these rules (even if these rules themselves are decoupled from the service)

src/Microsoft.Tye.Hosting/Infrastructure/IngressHostMatcherPolicy.cs
```diff
@@ -0,0 +1,209 @@
...
+                    {
+                        // If we get here then the host doesn't match.
+                        continue;
+                    }
+
+                    if (MemoryExtensions.Equals(port, WildcardHost, StringComparison.OrdinalIgnoreCase))
+                    {
+                        // Port is a wildcard, we allow any port.
+                    }
+                    else if (port.Length > 0 && (!int.TryParse(port, out var parsed) || parsed != requestPort))
```
jkotalik: Just some thoughts in my head: this all works when you have multiple replicas per app. The port selection is based on the exposed port, not the replica ports, right?

src/Microsoft.Tye.Hosting/Infrastructure/IngressHostMatcherPolicy.cs
```diff
@@ -0,0 +1,209 @@
...
+                var hosts = e.Metadata.GetMetadata<IngressHostMetadata>()?.Hosts;
+                if (hosts == null || hosts.Count == 0)
+                {
+                    return false;
+                }
+
+                foreach (var host in hosts)
+                {
+                    // Don't run policy on endpoints that match everything
+                    var key = CreateEdgeKey(host);
```
jkotalik: What is an EdgeKey supposed to mean? Name isn't very clear.

src/Microsoft.Tye.Hosting/Infrastructure/IngressHostMatcherPolicy.cs
```diff
@@ -0,0 +1,209 @@
...
+
+            var hostParts = host.Split(':');
+            if (hostParts.Length == 1)
+            {
+                if (!string.IsNullOrEmpty(hostParts[0]))
+                {
+                    return new EdgeKey(hostParts[0], null);
+                }
+            }
+            if (hostParts.Length == 2)
```
jkotalik:
```suggestion

            if (hostParts.Length == 2)
```

src/Microsoft.Tye.Hosting/Infrastructure/IngressHostMatcherPolicy.cs
davidfowl: Everything here is based on replica port.

src/Microsoft.Tye.Hosting/Infrastructure/IngressHostMatcherPolicy.cs
davidfowl: It's some copy pasta

src/Microsoft.Tye.Hosting/Infrastructure/IngressHostMatcherPolicy.cs
```diff
@@ -0,0 +1,209 @@
...
+
+                var hosts = candidates[i].Endpoint.Metadata.GetMetadata<IngressHostMetadata>()?.Hosts;
+                if (hosts == null || hosts.Count == 0)
+                {
+                    // Can match any host.
+                    continue;
+                }
+
+                var matched = false;
+                var (requestHost, requestPort) = GetHostAndPort(httpContext);
```
jkotalik: Be careful with requestPort being null here. The check for `parsed != requestPort` at least deserves a comment about checking for null.

davidfowl:
Moved ingress to a top level concept.

src/Microsoft.Tye.Hosting/IngressService.cs
```diff
@@ -0,0 +1,287 @@
...
+                        {
+                            foreach (var binding in targetServiceDescription.Bindings)
+                            {
+                                if (binding.Port == null)
+                                {
+                                    continue;
+                                }
+
+                                var port = target.PortMap[binding.Port.Value][i];
+                                var url = $"{binding.Protocol ?? "http"}://localhost:{port}";
```
jkotalik: I see this pattern of `binding.Protocol ?? "http"` everywhere. Can we just default to http if it isn't set?

src/Microsoft.Tye.Hosting/IngressService.cs
```diff
@@ -0,0 +1,287 @@
...
+
+                }
+                finally
+                {
+                    webApp.Dispose();
+                }
+            }
+        }
+
+        private class ConnectionRetryHandler : DelegatingHandler
```
jkotalik: Put this in a separate class and share with test code. https://github.com/dotnet/tye/blob/master/test/E2ETest/RetryHandler.cs already exists.

src/Microsoft.Tye.Hosting/Model/IngressRule.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
```
jkotalik: copyright.

src/Microsoft.Tye.Hosting/Model/IngressRule.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Tye.Hosting.Model
+{
+    public class IngressRule
+    {
+        public IngressRule(string? host, string? path, string service)
+        {
+            Host = host;
```
jkotalik: Do we want the dashboard API to return these rules in api/v1/services?

src/Microsoft.Tye.Hosting/Model/IngressRunInfo.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
```
jkotalik: copyright.

src/Microsoft.Tye.Hosting/Model/IngressStatus.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Tye.Hosting.Model
+{
+    public class IngressStatus : ReplicaStatus
+    {
+        public IngressStatus(Service service, string name) : base(service, name)
+        {
+        }
+
```
jkotalik: nit: newline

src/tye/ConfigModel/ConfigIngress.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
```
jkotalik: copyright.

src/Microsoft.Tye.Hosting/Model/IngressRule.cs
davidfowl: Yea, I haven't decided if it should differ from services. Right now it'll show IngressStatus, I could add the rules there but it's not really important at the moment.

src/Microsoft.Tye.Hosting/IngressService.cs
davidfowl: I can delete the test code but these are a bit different.

src/tye/InitHost.cs
```diff
@@ -71,6 +72,8 @@ public static (string, string) CreateTyeFileContent(FileInfo? path, bool force)
                     // If the input file is a project or solution then use that as the name
                     application.Name = Path.GetFileNameWithoutExtension(path.Name).ToLowerInvariant();
 
+                    application.Ingress = null!;
```
jkotalik: Why do you need to set this to null?

src/Microsoft.Tye.Hosting/IngressService.cs
davidfowl: If we ever need to infer anything from "not set" vs "explicitly set to http" then it's potentially an issue.

src/tye/InitHost.cs
davidfowl: So it doesn't show up as an empty array when you call `tye init`

PR #180, opened 2020-03-23
Fix NullRef in Tye run when not specifying and projects to debug.

jkotalik:
Should have tested doing `tye run` without debugging projects. Whoops.


davidfowl:
This breaks normal `tye run`?

jkotalik:
Yep.

davidfowl:
Don't we have a test for that?

jkotalik:
Added a new test. We have tests for `tye run` but it didn't call the right constructor.

PR #188, opened 2020-03-24
Use msbuild project info for running

### LINKED ISSUE #184: Use MSBuild to locate output folder and binaries for run scenarios
### Describe the bug
Today `tye run` hackishly finds the output of the binary by looking at the debug folder, picking the first TFM, and assmuming there's a file called `{ProjectName}.dll` in the output folder. Instead, we need to use msbuild to resolve the target assembly.

davidfowl:
- This fixes several subtle bugs around how we resolve output paths adhoc and uses msbuild to find various outputs.

Fixes #184

### DIFF
--- a/src/Microsoft.Tye.Core/ProjectServiceBuilder.cs
+++ b/src/Microsoft.Tye.Core/ProjectServiceBuilder.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 
 namespace Microsoft.Tye
@@ -39,6 +40,9 @@ namespace Microsoft.Tye
 
         // This is always set on the ApplicationFactory codepath.
         public string RunCommand { get; set; } = default!;
+
+        // This is always set on the ApplicationFactory codepath.
+        public string RunArguments { get; set; } = default!;
 
         // This is always set on the ApplicationFactory codepath.
         public string AssemblyName { get; set; } = default!;
--- a/src/Microsoft.Tye.Core/ProjectServiceBuilder.cs
+++ b/src/Microsoft.Tye.Core/ProjectServiceBuilder.cs
@@ -29,7 +29,22 @@ namespace Microsoft.Tye
         public string TargetFramework { get; set; } = default!;
 
         // This is always set on the ApplicationFactory codepath.
+        public string[] TargetFrameworks { get; set; } = default!;
+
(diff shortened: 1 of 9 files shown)


src/Microsoft.Tye.Core/ProjectServiceBuilder.cs
```diff
@@ -28,9 +28,24 @@ public ProjectServiceBuilder(string name, FileInfo projectFile)
         // This is always set on the ApplicationFactory codepath.
         public string TargetFramework { get; set; } = default!;
 
+        // This is always set on the ApplicationFactory codepath.
+        public string[] TargetFrameworks { get; set; } = default!;
```
jkotalik: Ew, the default! isn't pretty.

src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs
```diff
@@ -41,16 +41,17 @@ private async Task TransformProjectToContainer(Model.Application application, Mo
             var serviceDescription = service.Description;
             var serviceName = serviceDescription.Name;
 
-            var expandedProject = Environment.ExpandEnvironmentVariables(project.Project);
-            var fullProjectPath = Path.GetFullPath(Path.Combine(application.ContextDirectory, expandedProject));
-            service.Status.ProjectFilePath = fullProjectPath;
+            service.Status.ProjectFilePath = project.ProjectFile.FullName;
+            var targetFramework = project.TargetFramework;
```
jkotalik: It looks like TargetFramework can be null if TargetFrameworks isn't null. Should we assume one of the target frameworks as we are publishing on behalf of the person doing tye run?

src/tye/ApplicationBuilderExtensions.cs
```diff
@@ -52,7 +52,17 @@ public static Tye.Hosting.Model.Application ToHostingApplication(this Applicatio
                 }
                 else if (service is ProjectServiceBuilder project)
                 {
-                    var projectInfo = new ProjectRunInfo(project.ProjectFile.FullName, project.Args, project.Build);
+                    if (project.TargetFrameworks.Length > 1)
```
jkotalik: Ah nvrm, I see that you are checking here as well. I think it's fine then.

src/Microsoft.Tye.Core/ProjectServiceBuilder.cs
rynowak: get used to it. It's either massive constructor or this 😆

Issue #189, opened 2020-03-25
tye build cannot work when source code is on another drive


bravecobra:
### Describe the bug
I followed the sample but generated the code on my D: drive. The build command failed when tye tried to generate the Dockerfile for each project.

```
D:\tmp\microservice> tye build
Reading Project Details...
Reading Project Details...
Processing Service 'frontend'...
    Compiling Services...
    Publishing Project...
        Created Publish Output: 'D:\tmp\microservice\frontend\bin\Release\netcoreapp3.1\publish'
    Building Docker Image...
            unable to prepare context: unable to get relative Dockerfile path: Rel: can't make C:\Users\james\AppData\Lo
cal\Temp\tmp1413.tmp relative to D:\tmp\microservice\frontend\bin\Release\netcoreapp3.1\publish
Drats! 'build' failed:
        'docker build' failed.
```
### Further technical details
version 0.1.0-alpha.20168.8+f45b9444d4894009bde48e2a3411a52dd09497b1

bravecobra:
Seems similar to #153

rynowak:
This should be fixed as of `0.1.0-alpha.20174.3` - I don't have easy access to a windows machine to test it right now, so please let us know if you hit issues.

Issue #153, opened 2020-03-19
'docker build'-step of tye deploy fails when project is not on system drive


SebastianLng:
### Describe the bug
The 'tye deploy' command fails when the project is not on the system drive. The 'docker build' step results in an error.

### To Reproduce
Create and deploy a solution on a drive other than the system drive:
```
dotnet new razor -n frontend
dotnet new sln
dotnet sln add frontend
tye deploy --interactive
```

This results in the following error:

![error](https://user-images.githubusercontent.com/25402641/77123309-0a363f80-6a40-11ea-8944-6699a3ad0eef.jpg)

These steps work on the system drive and fail on any other.

### Further technical details
- Version: 0.1.0-alpha.20168.8+f45b9444d4894009bde48e2a3411a52dd09497b1
- Windows 10 1909
- Docker Desktop 2.2.0.4

rynowak:
Thanks for the report! We'll fix this.

rynowak:
This should be fixed as of `0.1.0-alpha.20174.3` - I don't have easy access to a windows machine to test it right now, so please let us know if you hit issues.

SebastianLng:
The build still fails with the same error:
![builderror](https://user-images.githubusercontent.com/25402641/77545035-1ee45e80-6eaa-11ea-955e-1f65ea07476c.jpg)

rynowak:
Ok, I'll take a look at this again.

rynowak:
@SebastianLng -  I cannot reproduce this failure. I suspect I may have told you the wrong version for the fix 😢 

The fix was to publish to a temp folder instead of inside the app's folder. Relevant bit from the logs (which I don't see in your example) 

```txt
Running 'dotnet publish'.
        > dotnet publish "D:\testing\testing.csproj" -c Release -o "C:\Users\rynowak.REDMOND\AppData\Local\Temp\pyaopkfm.cmx"
            Microsoft (R) Build Engine version 16.5.0+d4cbfca49 for .NET Core
            Copyright (C) Microsoft Corporation. All rights reserved.
            
              Restore completed in 33.83 ms for D:\testing\testing.csproj.
              You are using a preview version of .NET Core. See: https://aka.ms/dotnet-core-preview
              testing -> D:\testing\bin\Release\netcoreapp3.1\testing.dll
              testing -> C:\Users\rynowak.REDMOND\AppData\Local\Temp\pyaopkfm.cmx\
        Done running 'dotnet publish' exit code: 0
[...]

PR #192, opened 2020-03-25
Use same directory for Dockerfile and app

rynowak:
Fixes: #189
Fixes: #153

This change rejiggers the docker build infrastructure to publish the project
and generate a Dockerfile in the same directory. This way they can't be on
separate drives!

Also added some focused tests for single-phase docker build.

### DIFF
--- a/src/Microsoft.Tye.Core/DockerContainerBuilder.cs
+++ b/src/Microsoft.Tye.Core/DockerContainerBuilder.cs
@@ -34,55 +34,93 @@ namespace Microsoft.Tye
                 throw new ArgumentNullException(nameof(container));
             }
 
-            using var tempFile = TempFile.Create();
-
-            var dockerFilePath = Path.Combine(project.ProjectFile.DirectoryName, "Dockerfile");
-            if (File.Exists(dockerFilePath))
-            {
-                output.WriteDebugLine($"Using existing Dockerfile '{dockerFilePath}'.");
-            }
-            else
-            {
-                await DockerfileGenerator.WriteDockerfileAsync(output, application, project, container, tempFile.FilePath);
-                dockerFilePath = tempFile.FilePath;
-            }
-
-            // We need to know if this is a single-phase or multi-phase Dockerfile because the context directory will be
-            // different depending on that choice.
             string contextDirectory;
-            if (container.UseMultiphaseDockerfile ?? true)
+            var dockerFilePath = Path.Combine(project.ProjectFile.DirectoryName, "Dockerfile");
+
+            TempFile? tempFile = null;
+            TempDirectory? tempDirectory = null;
+
+            try
             {
-                contextDirectory = ".";
-            }
-            else
-            {
-                var publishOutput = project.Outputs.OfType<ProjectPublishOutput>().FirstOrDefault();
-                if (publishOutput is null)
+                // We need to know if this is a single-phase or multi-phase Dockerfile because the context directory will be
+                // different depending on that choice.
+                //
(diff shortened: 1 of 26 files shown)


test/E2ETest/testassets/projects/multi-phase-dockerfile/Program.cs
```diff
@@ -0,0 +1,26 @@
+using System;
```
rynowak: Everything past this point is just scaffolded apps.

test/E2ETest/TyeBuildTests.Dockerfile.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Tye;
+using Xunit;
+using static E2ETest.TestHelpers;
+
+namespace E2ETest
+{
+    // Tests permutations of our Dockerfile-related behaviors.
+    public partial class TyeBuildTests
```
davidfowl: Can these tests run in parallel?

src/Microsoft.Tye.Core/DockerContainerBuilder.cs
```diff
@@ -34,55 +34,93 @@ public static async Task BuildContainerImageAsync(OutputContext output, Applicat
...
+                    contextDirectory = ".";
+
+                    if (File.Exists(dockerFilePath))
+                    {
+                        output.WriteDebugLine($"Using existing Dockerfile '{dockerFilePath}'.");
+                    }
+                    else
+                    {
+                        // We need to write the file, let's stick it under obj.
+                        Directory.CreateDirectory(Path.Combine(project.ProjectFile.DirectoryName, "obj"));
```
davidfowl: Instead of hard coding obj use the variable from msbuild `IntermediateOutputPath` (I think that's what it's called).

src/Microsoft.Tye.Core/DockerContainerBuilder.cs
```diff
@@ -34,55 +34,93 @@ public static async Task BuildContainerImageAsync(OutputContext output, Applicat
...
+                    {
+                        throw new InvalidOperationException("We should have published the project for a single-phase Dockerfile.");
+                    }
+
+                    contextDirectory = publishOutput.Directory.FullName;
+
+                    // Clean up directory when done building image
+                    tempDirectory = new TempDirectory(publishOutput.Directory);
+
+                    if (File.Exists(dockerFilePath) & container.UseMultiphaseDockerfile == false)
```
davidfowl: && or &?

test/E2ETest/TyeBuildTests.Dockerfile.cs
rynowak: Nope. Can't test docker in parallel.

src/Microsoft.Tye.Core/DockerContainerBuilder.cs
rynowak: ooh wow, that's a rare one. Should be `&&`

src/Microsoft.Tye.Core/DockerContainerBuilder.cs
rynowak: Adding it. FYI this code path doesn't have a UX right now, so you have no way to hit it. I'm just maintaining the code because we'll want to expose it eventually.

rynowak:
> I should normalize the other directories here as well. I did it later.

Ya I saw that after you merged. It's best if we do all of the data-cleaning before stuff makes it to `ApplicationBuilder`.

PR #193, opened 2020-03-25
Print error if there are no runnable services in the application.

davidfowl:


src/tye/Program.DeployCommand.cs
```diff
@@ -39,6 +39,12 @@ public static Command CreateDeployCommand()
 
                 var output = new OutputContext(console, verbosity);
                 var application = await ApplicationFactory.CreateAsync(output, path);
+
+                if (application.Services.Count == 0)
+                {
+                    throw new CommandException($"No services found in \"{application.Source.Name}\"");
```
jkotalik: Why not throw in the hosts instead of the commands like you did with the build command?

davidfowl: I just threw in the thing that calls `ApplicationFactory.CreateAsync`

Issue #197, opened 2020-03-25
Add Voting Sample from Community Standup Intro


spboyer:
https://www.youtube.com/watch?v=yzNEHrWM78o

jkotalik:
@glennc cool if I port it over?

PR #201, opened 2020-03-25
Add voting sample 

jkotalik:
Fixes https://github.com/dotnet/tye/issues/197

@davidfowl what do I need to do to convert the ingress.yaml to the tye format?


samples/voting/tye.yaml
```diff
@@ -0,0 +1,25 @@
...
+  bindings:
+    - port: 5007
+- name: redis
+  dockerImage: redis
+  bindings:
+    - port: 6379
+- name: worker
+  project: worker/worker.csproj
+  bindings:
+    - port: 5000
```
davidfowl: Is the worker HTTP based?

samples/voting/tye.yaml
```diff
@@ -0,0 +1,25 @@
+name: VotingSample
```
davidfowl: Remove all of the ports for individual projects. It should autogenerate ports.

samples/voting/tye.yaml
```diff
@@ -0,0 +1,25 @@
...
+services:
+- name: vote
```
davidfowl:
```yml
ingress:
  rules: 
     - path: /vote
       service: vote
     - path: /results
       service: results
```

jkotalik: Is this only for local for now?

PR #203, opened 2020-03-26
Remove reading of launch settings

davidfowl:
- If bindings are specified, add an http and https binding with autodetect port to true.

### DIFF
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -97,7 +97,7 @@ namespace Microsoft.Tye
                 builder.Services.Add(service);
 
                 // If there are no bindings and we're in ASP.NET Core project then add an HTTP and HTTPS binding
-                if (configService.Bindings.Count == 0 && 
+                if (configService.Bindings.Count == 0 &&
                     service is ProjectServiceBuilder project2 &&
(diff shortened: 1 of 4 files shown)


src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -102,21 +102,19 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
...
-                        Name = "https",
-                        AutoAssignPort = true,
-                        Protocol = "https"
-                    };
+                    });
 
-                    service.Bindings.Add(httpBinding);
-                    service.Bindings.Add(httpsBinding);
+                    // TODO: Uncomment when the CI is unblocked
+                    //service.Bindings.Add(new BindingBuilder()
```
jkotalik: What's the problem with this?

davidfowl: dev cert broken on the CI (expired on windows and untrusted on osx/linux)

davidfowl: https://dev.azure.com/dnceng/public/_build/results?buildId=574074&view=ms.vss-test-web.build-test-results-tab&runId=18003516&paneView=attachments&resultId=100013

PR #211, opened 2020-03-26
Two minor cleanup items

jkotalik:
- Adds auto assign port to the voting sample (was accidentally stashed when I was testing)
- Increase docker stop/remove timeouts and log a warning when we failed to timeout. I frequently hit the 5 second timeout locally, sometimes taking 5 to 10 seconds to stop docker.

### DIFF
--- a/samples/voting/tye.yaml
+++ b/samples/voting/tye.yaml
@@ -8,8 +8,6 @@ services:
     - port: 6379
 - name: worker
   project: worker/worker.csproj
-  bindings:
-    - autoAssignPort: true
 - name: postgres
   image:  postgres
   env:
--- a/samples/voting/tye.yaml
+++ b/samples/voting/tye.yaml
@@ -8,6 +8,8 @@ services:
     - port: 6379
 - name: worker
   project: worker/worker.csproj
+  bindings:
+    - autoAssignPort: true
 - name: postgres
   image:  postgres
   env:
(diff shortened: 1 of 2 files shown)


samples/voting/tye.yaml
```diff
@@ -8,6 +8,8 @@ services:
     - port: 6379
 - name: worker
   project: worker/worker.csproj
+  bindings:
```
davidfowl: Undo this change, it's no longer required.

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -207,12 +208,17 @@ async Task RunDockerContainer(IEnumerable<(int Port, int? ContainerPort, int Bin
...
-                var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(5));
+                var timeoutCts = new CancellationTokenSource(DockerStopTimeout);
 
                 _logger.LogInformation("Stopping container {ContainerName} with ID {ContainerId}", replica, shortContainerId);
 
                 result = await ProcessUtil.RunAsync("docker", $"stop {containerId}", throwOnError: false, cancellationToken: timeoutCts.Token);
 
+                if (timeoutCts.IsCancellationRequested)
+                {
+                    _logger.LogWarning($"Failed to stop container after {DockerStopTimeout.Seconds} seconds, container will most likely be running.", replica, shortContainerId);
```
rynowak: 👏

Issue #208, opened 2020-03-26
Yaml image property not found on type 'Microsoft.Tye.ConfigModel.ConfigService' at `tye run` time


luisgizirian:
### Describe the bug
While running the Getting Started, I came to the redis piece (https://github.com/dotnet/tye/blob/master/docs/redis.md).

### To Reproduce
When trying `tye run` an error appears. I double checked indentation and copy/paste potential origins.

tye.yaml contents:
```
# tye application configuration file
# read all about it at https://github.com/dotnet/tye
#
# when you've given us a try, we'd love to know what you think:
#    https://aka.ms/AA7q20u
#
name: tye-microservice
services:
- name: frontend
  project: frontend/frontend.csproj
- name: backend
  project: backend/backend.csproj
- name: redis
  image: redis
```
Exception:

An unhandled exception has occurred, how unseemly: 
YamlDotNet.Core.YamlException: (Line: 13, Col: 3, Idx: 324) - (Line: 13, Col: 3, Idx: 324): Exception during deserialization
 ---> System.Runtime.Serialization.SerializationException: Property 'image' not found on type 'Microsoft.Tye.ConfigModel.ConfigService'.
   at YamlDotNet.Serialization.TypeInspectors.TypeInspectorSkeleton.GetProperty(Type type, Object container, String name, Boolean ignoreUnmatched)
   at YamlDotNet.Serialization.NodeDeserializers.ObjectNodeDeserializer.YamlDotNet.Serialization.INodeDeserializer.Deserialize(IParser parser, Type expectedType, Func`3 nestedObjectDeserializer, Object& value)
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   --- End of inner exception stack trace ---
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.ValueDeserializers.AliasValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.<>c__DisplayClass3_0.<DeserializeValue>b__0(IParser r, Type t)
[...]

Issue #205, opened 2020-03-26
`tye run` runs the incorrect build version.


stevejgordon:
### Description

I attempted to use `tye run` against an existing sample solution of mine which I recently upgraded to .NET 5 from 3.1. The solution contains a front end application (Razor Pages) which calls a backend API service, similar to the sample.

`tye run` attempts to run the incorrect version.

### To Reproduce
I updated the front end to use the service discovery configuration.

When I execute `tye run`, the API project starts successfully. The web front end fails repeatedly (see output below).

The build produces output in the bin\Debug\netcoreapp5.0

The run step, however, seems to pick up an older build from the bin\Debug\netcoreapp3.0 folder, which exists from historical builds.

`dotnet "E:\Projects\Pluralsight\Worker Services Course\Tye Playground\src\TennisBookings.Web\bin\Debug\netcoreapp3.0\TennisBookings.Web.dll"`

**Full Output**
```
dotnet build "E:\Projects\Pluralsight\Worker Services Course\Tye Playground\src\TennisBookings.Web\TennisBookings.Web.csproj" /nologo
Restore completed in 70.83 ms for E:\Projects\Pluralsight\Worker Services Course\Tye Playground\src\TennisBookings.Web\TennisBookings.Web.csproj.
You are using a preview version of .NET Core. See: https://aka.ms/dotnet-core-preview
TennisBookings.Web -> E:\Projects\Pluralsight\Worker Services Course\Tye Playground\src\TennisBookings.Web\bin\Debug\netcoreapp5.0\TennisBookings.Web.dll
TennisBookings.Web -> E:\Projects\Pluralsight\Worker Services Course\Tye Playground\src\TennisBookings.Web\bin\Debug\netcoreapp5.0\TennisBookings.Web.Views.dll
Build succeeded.
0 Warning(s)
0 Error(s)
Time Elapsed 00:00:01.55
[tennisbookings.web_3504bce3-8]:dotnet "E:\Projects\Pluralsight\Worker Services Course\Tye Playground\src\TennisBookings.Web\bin\Debug\netcoreapp3.0\TennisBookings.Web.dll"
[tennisbookings.web_3504bce3-8]: Could not execute because the specified command or file was not found.
[tennisbookings.web_3504bce3-8]: Possible reasons for this include:
[tennisbookings.web_3504bce3-8]: * You misspelled a built-in dotnet command.
[...]

PR #213, opened 2020-03-26
Change version in getting_started for tye

jkotalik:
I made sure that tye run and tye deploy work, just in case. There a few bugs (https://github.com/dotnet/tye/issues/208, https://github.com/dotnet/tye/issues/205) that have already been fixed or have had docs that are updated that don't work against previous tye versions.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -8,20 +8,20 @@ Tye is a tool that makes developing, testing, and deploying microservices and di
 2. Install tye via the following command:
 
     ```text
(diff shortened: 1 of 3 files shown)


davidfowl:
Oh wait, did we update the doc sytax as well? image vs dockerImage etc?

jkotalik:
Yes, but I think we need to update the schema and a doc. I'll do that here.

davidfowl:
@jkotalik The schema is already updated. It has both old and new though.

jkotalik:
I don't think it's updated for the schema.

https://github.com/dotnet/tye/pull/213/files#diff-9762849daa5c598c4b9dc9e7130e0b92R175-R207.

image

davidfowl:
Ah you're right

PR #221, opened 2020-03-26
Split out port assignment from the proxy

davidfowl:


src/Microsoft.Tye.Hosting/PortAssigner.cs
```diff
@@ -0,0 +1,104 @@
+﻿using System.Collections.Generic;
```
jkotalik: copyright.

davidfowl:
> I mean sure? Like I'm all for code cleanup, just not sure what the motivation is here.

Splitting things out that don't belong together.

PR #227, opened 2020-03-27
do dispose async

SimonCropp:
i assume you would prefer to do dispose in an async manner where possible?

### DIFF
--- a/src/Microsoft.Tye.Core/ProcessUtil.cs
+++ b/src/Microsoft.Tye.Core/ProcessUtil.cs
@@ -109,7 +109,7 @@ namespace Microsoft.Tye
             process.BeginErrorReadLine();
 
             var cancelledTcs = new TaskCompletionSource<object?>();
-            using var _ = cancellationToken.Register(() => cancelledTcs.TrySetResult(null));
+            await using var _ = cancellationToken.Register(() => cancelledTcs.TrySetResult(null));
 
             var result = await Task.WhenAny(processLifetimeTask.Task, cancelledTcs.Task);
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Tye.Core/ProcessUtil.cs
```diff
@@ -109,7 +109,7 @@ public static class ProcessUtil
             process.BeginErrorReadLine();
 
             var cancelledTcs = new TaskCompletionSource<object?>();
-            using var _ = cancellationToken.Register(() => cancelledTcs.TrySetResult(null));
+            await using var _ = cancellationToken.Register(() => cancelledTcs.TrySetResult(null));
```
rynowak: Wow. I learned something new today thanks to this PR!

@davidfowl did you know about this? https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Threading/CancellationTokenRegistration.cs#L48

rynowak:
CI failures look unrelated to your changes. Rerunning.

src/Microsoft.Tye.Core/ProcessUtil.cs
SimonCropp: do u want a PR that does CancellationTokens?

davidfowl: Yah

davidfowl:
There’s a machine in the ci pool with busted dev certs

rynowak:
> There’s a machine in the ci pool with busted dev certs 

Is there a way to repair it? We could run the dev cert tool as part of our CI.

src/Microsoft.Tye.Core/ProcessUtil.cs
SimonCropp: name for CancellationToken variables? my preference is `cancellation`, but you already have some with `cancellationToken`

davidfowl:
I’m going to do that the next time I can repro it

PR #228, opened 2020-03-27
changed UI colors for dashboard

amarie401:
Nothing fancy, updated the UI colors for the dashboard. Thoughts??

tye


rynowak:
Ship it!

PR #233, opened 2020-03-27
add cancellation tokens

SimonCropp:
ok i need help with this one. not sure when we should do "throwifcancelrequested"


rynowak:
What's the goal here? We don't expect most of these APIs to be consumed outside of what they're used for today. Are there specific things that have bad cancellation behavior today?

SimonCropp:
it was initially in response to this https://github.com/dotnet/tye/pull/227#discussion_r399037857

but i think this would be better done by peeps who better understand the codebase. so will close for now

PR #234, opened 2020-03-27
Improve container base image detection

davidfowl:
- Improve ASP.NET project detection to work with netcoreapp2.1 applications
- Grab the target framework name and version from the project separately so we can use the vesion to determine the base image
- Updated Dockerfile generation and tye run --docker to work with different container versions based on the TFM version.

### DIFF
--- a/src/Microsoft.Tye.Core/ProjectReader.cs
+++ b/src/Microsoft.Tye.Core/ProjectReader.cs
@@ -219,9 +219,17 @@ namespace Microsoft.Tye
             project.TargetFramework = targetFramework;
             output.WriteDebugLine($"Found target framework: {targetFramework}");
 
+            // TODO: Parse the name and version manually out of the TargetFramework field if it's non-null
+            project.TargetFrameworkName = projectInstance.GetPropertyValue("_ShortFrameworkIdentifier");
+            project.TargetFrameworkVersion = projectInstance.GetPropertyValue("_ShortFrameworkVersion") ?? projectInstance.GetPropertyValue("_TargetFrameworkVersionWithoutV");
+
             var sharedFrameworks = projectInstance.GetItems("FrameworkReference").Select(i => i.EvaluatedInclude).ToList();
             project.Frameworks.AddRange(sharedFrameworks.Select(s => new Framework(s)));
             output.WriteDebugLine($"Found shared frameworks: {string.Join(", ", sharedFrameworks)}");
+
+            project.IsAspNet = project.Frameworks.Any(f => f.Name == "Microsoft.AspNetCore.App") ||
+                               projectInstance.GetPropertyValue("MicrosoftNETPlatformLibrary") == "Microsoft.AspNetCore.App" ||
+                               projectInstance.GetPropertyValue("_AspNetCoreAppSharedFxIsEnabled") is string s && s != null && bool.Parse(s);
 
             output.WriteDebugLine($"Evaluation Took: {sw.Elapsed.TotalMilliseconds}ms");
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs
```diff
@@ -83,10 +83,9 @@ private async Task TransformProjectToContainer(Service service, ProjectRunInfo p
             serviceDescription.RunInfo = dockerRunInfo;
         }
 
-        private static string DetermineContainerImage(string targetFramework)
+        private static string DetermineContainerImage(ProjectRunInfo project)
         {
-            // TODO: Determine the base image from the tfm
-            return "mcr.microsoft.com/dotnet/core/sdk:3.1-buster";
+            return $"mcr.microsoft.com/dotnet/core/sdk:{project.TargetFrameworkVersion}";
```
spboyer: This pattern would assume that all development, if using Linux containers, would be done on Debian based containers given the TFM base image is always such. How would we specify an alpine or other flavor? Should we consider looking leveraging  [Container MS Build Properties](https://docs.microsoft.com/en-us/visualstudio/containers/container-msbuild-properties?view=vs-2019) like `DockerfileTag` here?

davidfowl: Not possible to override it today. I’d imagine we’ll eventually let you control the base image or docker file completely

davidfowl: Seems reasonable to support these in msbuild for alignment anyways.

davidfowl: @spboyer can you file an issue for this?

spboyer: Done - https://github.com/dotnet/tye/issues/240

src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -219,10 +219,18 @@ private static void EvaluateProject(OutputContext output, ProjectServiceBuilder
             project.TargetFramework = targetFramework;
             output.WriteDebugLine($"Found target framework: {targetFramework}");
 
+            // TODO: Parse the name and version manually out of the TargetFramework field if it's non-null
+            project.TargetFrameworkName = projectInstance.GetPropertyValue("_ShortFrameworkIdentifier");
+            project.TargetFrameworkVersion = projectInstance.GetPropertyValue("_ShortFrameworkVersion") ?? projectInstance.GetPropertyValue("_TargetFrameworkVersionWithoutV");
+
```
rynowak: This really seems suspect :( do we know that 3.1 and 5.0 SDKS behavre the same way?

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -99,7 +99,8 @@ async Task RunDockerContainer(IEnumerable<(int ExternalPort, int Port, int? Cont
                     // Default to development environment
                     ["DOTNET_ENVIRONMENT"] = "Development",
                     // Remove the color codes from the console output
-                    ["DOTNET_LOGGING__CONSOLE__DISABLECOLORS"] = "true"
+                    ["DOTNET_LOGGING__CONSOLE__DISABLECOLORS"] = "true",
+                    ["ASPNETCORE_LOGGING__CONSOLE__DISABLECOLORS"] = "true"
```
rynowak: OOOOH yeah!

src/Microsoft.Tye.Core/ProjectReader.cs
davidfowl: Yes, they do, at least by default they do. I still think we should do the janky fallback to parsing the tfm.

PR #236, opened 2020-03-27
remove redundant state machines

SimonCropp:


jkotalik:
How are you finding these? Do you have an analyzer that you are running or just manually combing through to find these?

I'd like to have some of these locally if so 😄

Issue #207, opened 2020-03-26
Tye init --force omit defaults behavior when tye.yaml already exist


unaizorrilla:
### Describe the bug
**Tye init --force** does not omit default values on serialization when tye.yaml already exist. 

When tye init is executed at the first time the yaml  result is:

```yaml
# tye application configuration file
# read all about it at https://github.com/dotnet/tye
#
# when you've given us a try, we'd love to know what you think:
#    https://aka.ms/AA7q20u
#
name: tye-samples
services:
- name: frontend
  project: FrontEnd/FrontEnd.csproj
- name: api
  project: Api/Api.csproj
```
If tye init --force is executed again, some default values are not omited  and result file include default values:

```yaml
# tye application configuration file
# read all about it at https://github.com/dotnet/tye
#
# when you've given us a try, we'd love to know what you think:
#    https://aka.ms/AA7q20u
#
name: tye-samples
services:
- name: frontend
  project: FrontEnd/FrontEnd.csproj
  bindings:
  - autoAssignPort: true
    protocol: https
  - autoAssignPort: true
    protocol: http
  volumes: []
  env: []
- name: api
  project: Api/Api.csproj
  bindings:
  - autoAssignPort: true
    protocol: https
  - autoAssignPort: true
    protocol: http
  volumes: []
  env: []
ingress: []
```

The serializer is configured correctly in both cases:
 
**ConfigureDefaultValuesHandling(DefaultValuesHandling.OmitDefaults)**

But the application object from ConfigFactory.FromFile

https://github.com/dotnet/tye/blob/2a1b22d4bd911ef4ce28845ac49cbd620727be1d/src/tye/InitHost.cs#L60 

have null values for defaults when tye.yaml does not exist and not null,empty arrays, when tye.yaml exist. For yamldotnet serializer empty array is not a default value and is not omited.

### To Reproduce
tye init
tye init --force

unaizorrilla:
Hi @davidfowl 

The label "good first issue" is like "up for grabs" ? can i do a pr for this?

davidfowl:
Yep, please send a PR 😄

rynowak:
Yes, please do :)

unaizorrilla:
Fixed with #237

Issue #235, opened 2020-03-27
Tests fails depending on Environment.NewLine


unaizorrilla:
### Describe the bug

At this moment, a lot of test are failing depending on Environment.NewLine, all tests that compare strings like 

https://github.com/dotnet/tye/blob/067d1efac75cef82d0777a077616bf9d6536e718/test/E2ETest/TyeInitTests.cs#L79

are failing if Environment.NewLine is Windows ( EOL = \r\n) because expected string is created with Linux based OS (EOL = \n)

YamlDotNet use Environment.NewLine on serialization and I can't see any way to specify this

### To Reproduce

Run tests on windows machine!

We can create some  string extension method on Tests project to fix expected string with current OS EOL like

 ```csharp
        public static string NormalizeNewLines(this string value)
        {
            return value
                .Replace("\r\n", "\n")
                .Replace("\n", Environment.NewLine);
        }
```

And fix tests 

 ```csharp
Assert.Equal(expectedContent, content.NormalizeNewLines());
 ```

If this solutions is ok I can contribute this!

unaizorrilla:
Fixed with #237

PR #237, opened 2020-03-27
Fix for issue #207

unaizorrilla:
Created a new object graph visitor for yaml dotnet to ignore default and also empty collections. Unifiy creation of serializer / deserializer well configured.

### DIFF
--- a/src/tye/InitHost.cs
+++ b/src/tye/InitHost.cs
@@ -1,14 +1,11 @@
 ﻿using System.IO;
 using Microsoft.Tye.ConfigModel;
-using Microsoft.Tye.Hosting.Dashboard;
-using YamlDotNet.Serialization;
-using YamlDotNet.Serialization.NamingConventions;
+using Microsoft.Tye.Serialization;
 
 namespace Microsoft.Tye
 {
     public class InitHost
     {
-
         public static string CreateTyeFile(FileInfo? path, bool force)
         {
             var (content, outputFilePath) = CreateTyeFileContent(path, force);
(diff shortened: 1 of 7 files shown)


src/tye/InitHost.cs
```diff
@@ -58,10 +55,7 @@ public static (string, string) CreateTyeFileContent(FileInfo? path, bool force)
             if (path is FileInfo && path.Exists)
             {
                 var application = ConfigFactory.FromFile(path);
-                var serializer = new SerializerBuilder()
-                    .WithNamingConvention(CamelCaseNamingConvention.Instance)
-                    .ConfigureDefaultValuesHandling(DefaultValuesHandling.OmitDefaults)
-                    .Build();
+                var serializer = YamlSerializer.CreateSerializer();
```
jkotalik: Hm, seems like the tye init tests are failing due to a new line difference (Expected \r\n, got \n). Not sure why that would change. 

I added a work item to do tests with yaml comparison rather than file comparison as new line differences like these aren't important to test.

unaizorrilla: I recently fill a issue about this https://github.com/dotnet/tye/issues/235

davidfowl: May as well do it in this change since it's failing tests

unaizorrilla: ok

unaizorrilla: done!!

jkotalik:
LGTM.

Issue #29, opened 2020-03-02
tye package references


davidfowl:
The idea behind this command is to enable sharing of tye manifests. `tye pack` would convert projects to container images and would put that in an archive (maybe a nuget package).  `tye.yml` would also be extended to reference these manifests as dependencies for consumption. The idea is that I can share a set of services so that another team can depend on them. 

Imagine the following:

Team B works on the backend API which relies on redis for caching.

```yml
# tye application configuration file
# read all about it at https://github.com/dotnet/tye
name: backend-services
services:
- name: backend
  project: backend\backend.csproj
- name: redis
  dockerImage: redis
  bindings:
    - port: 6379
```

They run the `tye pack` command and it produces a package that can be pushed to a feed to share the manifest. It also containerizes the backend project, pushes it to a registry (which maybe should be burned into the manifest)

```
tye pack
Building container image backend for backend\backend.csproj
Pushed docker image: 'davidfowl/backend:1.0.0'
Package 'backend\backend-services.1.0.0.nupkg'
```

Now team A (the front end team) can reference backend as a package in their tye manifest.

```yml
name: frontend
services:
- name: frontend
  project: frontend\frontend.csproj
- name: backend
  package: backend/1.0.0  # package reference
```

`tye run` will download this package from the feed (using feeds specified in `nuget.config`) and will grab the `tye.yml` file from that package and will run it as part of the services locally.

davidfowl:
I think we need to front load this to see if it can "solve" some of the problems with multi repo development https://github.com/dotnet/tye/issues/239.

davidfowl:
# Tye Packages

Consider the following application broken into 2 repositories:

**Frontend Team**
- MyApp.Frontend

**Backend Team**
- MyApp.Backend
- MyApp.Worker 
- RabbitMq

## Option 1

Package each application as nuget packages with the publish contents and push them to a nuget feed.

FrontEnd Team makes:

- **MyApp.Frontend.1.0.0.nupkg**

Backend Team makes:

- **MyApp.Backend.1.0.0.nupkg**
- **MyApp.Worker.1.0.0.nupkg**

Frontend team's `tye.yaml`

```yaml
name: multi-project
services:
- name: frontend
  project: frontend/frontend.csproj
  replicas: 2
  bindings:
  - port: 8000
- name: backend
  package: MyApp.Backend
  version: 1.0.0
  bindings:
  - port: 7000
- name: worker
  package: MyApp.Worker
  version: 1.0.0
[...]

PR #244, opened 2020-03-28
parse secret from dotnet user-secrets file

teabaggs:
Hi,
I hacked on this a bit as a way to get to know the codebase. 
I was thinking that this feature could rely on dontnet user secrets since that is a thing already and is cross plat. Users will somehow have to specify a secret ID so we can get a handle on the secrets.json file, and the name of the config value in the secrets file. Tye can read the secrets file and pull out the named value and inject into into the env of the container. 
Obviously my code needs work but if this seems like a reasonable approach please gimme some pointers and I'd be happy to work on this feature.

tye.yaml
```yaml
- name: redis
  image: redis
  bindings:
    - port: 6379
  env:
    - name: SUPER_SECRET
      value: $tye/Password
```
secrets file
```json
{
  "Password": "SuperDawg"
}
```
`tye run` outputs
```
Running docker command run -d  -v C:\Users\X\AppData\Roaming\Microsoft\UserSecrets:/root/.microsoft/usersecrets:ro  ... ASPNETCORE_URLS=http://*:6379 -e PORT=6379 -e SUPER_SECRET=SuperDawg ...
```
The rest of the line is omitted for brevity. 

I noticed that the entire secrets folder is mounted. If there was an option to specify a secret ID for a container then that single secrets file could be mounted. At my early stage of involvement I'm not sure what value is there in mounting secrets for container services anyway.

Thanks,
Phil


davidfowl:
Thanks! This did cross my mind when filing that issue, let me share some thoughts:
- We should have a first class secret value for env variables, not a template.
- Re-using the user secret id storage doesn't really have any benefits:
   - The secret id wouldn't be passed into each of the projects for app usage so there's no sharing between app and orchestrator (maybe that's by design though).
   - We need to pick a secret id somehow, does this happen on first run? Where is it persisted? tye.yaml?

davidfowl:
tye.yaml

```yaml
- name: redis
  image: redis
  bindings:
    - port: 6379
  env:
    - name: SUPER_SECRET
      secretName: Pwd
      secretStoreId: somethingunique 
```
secrets file
```json
{
  "Password": "SuperDawg"
}

File: src/Microsoft.Tye.Core/UserSecretsBuilder.cs [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+        public static string GetSecretValue(string envSource)
+        {
+            var sep = '/';
+            var index = envSource.IndexOf(sep);
+            var secretId = envSource[1..index];
+            var secretName = envSource[(index + 1)..];
+            return GetSecretValue(secretId, secretName);
+        }
+
+        private static string GetSecretValue(string secretId, string secretName)
```
davidfowl: This needs to be cross platform.

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -149,7 +149,33 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
...
+                         *   env:
+                         *     - name: POSTGRES_PASSWORD
+                         *       source: Postgres_Password
+                         *       type: secret
+                         */
+                        // then at the top service config provide the secret id                        
+                        /*
+                         * - name: postgres
+                         * - image: postgres
+                         * - secrets: <dotnet user-secrets guid>
```
davidfowl: Yep. something like this, maybe per value, per service and per file?

davidfowl:
I think we should maybe have a tye add/remove/list secret command as well

src/Microsoft.Tye.Core/UserSecretsBuilder.cs
rynowak: Do you have to give some suggestions? Isn't there another place we do this correctly in the codebase today?

davidfowl: https://github.com/dotnet/tye/blob/77e01e4b9167db457e720abe3504faebc3a35c6b/src/Microsoft.Tye.Hosting/DockerRunner.cs#L349

teabaggs: I can lean on [this](https://github.com/dotnet/tye/blob/master/src/Microsoft.Tye.Hosting/DockerRunner.cs#L349) and possibly extract that to some public or internal static class.

davidfowl: Yep extract it and share it

src/tye/tye.csproj
```diff
@@ -4,7 +4,7 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <RootNamespace>Microsoft.Tye</RootNamespace>
-    <AssemblyName>tye</AssemblyName>
+    <AssemblyName>stye</AssemblyName>
```
davidfowl: Undo

src/Microsoft.Tye.Core/ConfigModel/ConfigConfigurationSource.cs
```diff
@@ -13,5 +13,11 @@ public class ConfigConfigurationSource
 
         [Required]
         public string Value { get; set; } = default!;
+
+        [Required]
+        public string SecretName { get; set; } = default!;
```
davidfowl: These can’t be required unconditionally

src/Microsoft.Tye.Core/ConfigModel/ConfigConfigurationSource.cs
```diff
@@ -13,5 +13,11 @@ public class ConfigConfigurationSource
...
+
+        [Required]
+        public string SecretStoreId { get; set; } = default!;
```
davidfowl: This can’t be required either.

src/Microsoft.Tye.Core/ConfigModel/ConfigConfigurationSource.cs
```diff
@@ -13,5 +13,11 @@ public class ConfigConfigurationSource
 
         [Required]
         public string Value { get; set; } = default!;
```
davidfowl: Value can no longer be required.

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -149,7 +149,15 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
                     }
                     else if (service is ContainerServiceBuilder container)
                     {
-                        container.EnvironmentVariables.Add(envVar);
+                        if (configEnvVar.Value is null)
+                        {
+                            var value = UserSecretsBuilder.GetSecretValue(configEnvVar.SecretStoreId, configEnvVar.SecretName);
```
davidfowl: I think this should be resolved later, preserve the secret ness until the runner or deployment model needs it. @rynowak ideas here?

src/tye/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "tye": {
+      "commandName": "Project",
```
davidfowl: Delete this file

src/Microsoft.Tye.Core/ApplicationFactory.cs
davidfowl: Yea I like this idea. There should be a secret resolver component that runs later and does the right thing. It shouldn’t happen at the configuration layer

teabaggs: I feel like I want to add a new file to manage secrets and secret provider info so `tye.yaml` doesn't blow up. This would leave the config model untouched. Since the connection settings for something like a key vault provider might contain some sensitive information itself having those settings in a separate file would also allow that file to be easily excluded from source control.
 
In the hosting layer, right before a service is started, the secret resolver would check for any secrets and add new env vars. If the data is in tye.yaml, then that file should either be re-read to get secret info (so the config model is not dealing with any secret stuff) or some info about the secrets (but not the actual secret value) should be added to the config model and passed through up to the hosting layer so that the value can be resolved before running the service.

Pros for a secret file:
- keep secrets totally separate from config
- might also be useful to help tooling generate k8 manifests
- keep sensitive info out of SCM easily
- secret management command(s) you mention like `tye secret add ...` could work in this file
- [package references](https://github.com/dotnet/tye/issues/29) could strip out sensitive info and pack the secrets file so that consumers could just fill in the sensitive info and get up and running.

Cons for a secret file:
- extra file to parse and manage schema for
-  invites the possibility of having an env in `tye.yaml` and a secret with the same env var name so would require some kind of conflict resolution
- I'm sure there are others

For example:
```
name: microservice
secretProviders:
  - name: user-secrets
    type: user-secrets
    settings:
      secretId: ABC
  - name: other-user-secrets
    type: user-secrets
    settings:
      secretId: XYZ
  - name: key-vault
    type: key-vault
    settings:
      vaultName: MyAzureKeyVault
services:
  - name: redis    
    env:    
      - name: SECRET_1
        secretProvider: user-secrets
        secretKey: REDIS_SECRET
      - name: SECRET_2
        secretProvider: other-user-secrets
        secretKey: REDIS_SECRET
      - name: SECRET_3
        secretProvider: key-vault
        secretKey: REDIS_SECRET
```

would result in 3 env vars being added to the redis container
```
SECRET_1=<value of REDIS_SECRET from user secrets ABC>
SECRET_2=<value of REDIS_SECRET from user secrets XYZ>
SECRET_3=<value of REDIS_SECRET from key vault MyAzureKeyVault>
```

jkotalik:
We should have an issue to design this feature based on https://github.com/dotnet/tye/pull/244#discussion_r401754079. 

Going to close this for now as it's stale.

Issue #239, opened 2020-03-27
per project tye.yaml & merge on tye deploy


spboyer:
Pattern similar to helm charts where each service can have its own and apply on deployment.

rynowak:
I've done a little bit of thinking about this.

## Summary

The main issue to solve is the impedence mismatch inherent between development and deployment in the case where you have one service per git repo (which is pretty common), so that you can have a CI/CD pipeline per service:

- You want **all** of the deployment details of each service to live in its repo so that services are individually and repeatedly deployable
- You need to be able to reason about other services (not in your repo) for development

---

Some other tools kinda dodge this problem, or don't face it because they aren't doing the same things we are.

- docker-compose doesn't have deployment
- skaffold has dev and deployment for multiple services, but doesn't really try to abstract over dependencies or k8s

[Garden](https://garden.io) is pretty similar to Tye in goals (it is also  much more mature). Garden has a per-repo/per-application configuration file (`garden.yaml`), and per-container/pre-project config files. *Note that I am using our terminology to describe them in order to draw parallels.*

Garden also has a fairly sophisticated system for working with multiple repositories: https://docs.garden.io/advanced/using-remote-sources#importing-remote-repositories

## Rationale (why this matters)

CI/CD systems that people use are either, *a:* really advanced and able to figure out what part of a source tree needs to build/deploy, or *b:* not really advanced at all 😆.

One of the core value propositions of microservices is to deploy separately, and today creating multiple repositories is a really straightforward way to achieve. Part of our community achieves success today by scripting the creation of new repos with all of the associated boilerplate and CI definitions when they want to add a service. This usually leads to the kind of development pain that Project Tye is trying to solve. 

Some set of developers is using single-repo strategies to achieve success today, and they have either invested a lot of hard work into CI/CD customization, or are just deploying everything on every build. That's fine too, but we need to set our bar at meeting the more complex set of requirements. 

## Goals

Some goals that hopeful we agree upon for what success means for a feature like this:

- Support multi-repo development and deployment scenarios as first-class
- Don't make support for single-repo worse
- Keep things simple (don't require submodules or worse for basic scenarios)
- Make the experience match between multi-repo and single-repo (same commands, properties, etc) 

## Future considerations

Right now we have **minimal** configuration for what we do in deployment. We're missing all kinds of knobs and dials that might apply. I'm going to call out some features that might impact deployment in the future that are cross-cutting (apply to the application instead of a single service).

- environments/profiles (where are you deploying to)
[...]

PR #245, opened 2020-03-29
Added concept of dependencies to tye.yaml

davidfowl:
- It's now possible to reference other tye.yaml files to launch the services.
- This allows for splitting yaml files and composing them separately.

Fleshing out ideas from https://github.com/dotnet/tye/issues/29


src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -21,208 +21,229 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
...
+                            Host = configRule.Host,
+                            Path = configRule.Path,
+                            Service = configRule.Service,
+                        };
+                        ingress.Rules.Add(rule);
+                    }
                 }
 
-                foreach (var configRule in configIngress.Rules)
+                foreach (var configDependency in config.Dependencies)
```
davidfowl: This is the core of the change.

samples/multi-project/tye.yaml
```diff
@@ -4,16 +4,14 @@ name: multi-project
...
   replicas: 2
-  bindings:
-  - port: 8000
 - name: worker
   project: worker/worker.csproj
 - name: rabbit
   image: rabbitmq:3-management
   bindings:
     - port: 5672
+dependencies:
```
davidfowl: Eventually this will support packages and other things (git repositories?)

samples/multi-yaml/fe/frontend/Order.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Shared
+{
+    [DataContract]
```
rynowak: What's using DCS?

davidfowl: ProtobufNet (the code first one)

src/Microsoft.Tye.Core/ApplicationBuilder.cs
```diff
@@ -13,12 +13,15 @@ public ApplicationBuilder(FileInfo source, string name)
...
             Source = source;
             Name = name;
+            Application = name;
         }
 
         public FileInfo Source { get; set; }
 
         public string Name { get; set; }
 
+        public string? Application { get; set; }
```
rynowak: What is this for?

samples/multi-yaml/fe/tye.yaml
```diff
@@ -0,0 +1,8 @@
+app: multi-tye # name of the bigger application
+name: frontend # name of this application
+services:
+- name: frontend
+  project: frontend/frontend.csproj
+  replicas: 2
+dependencies:
+- path: ../be/tye.yaml
```
rynowak: Related to #239 - this is an improvement, but doesn't solve most of the really valuable scenarios.

The thing that gives me pause, I'm not sure why you'd want to break up tye.yaml into multiple files if you have to have access to them all at once.

src/Microsoft.Tye.Core/ConfigModel/ConfigApplication.cs
```diff
@@ -21,10 +21,15 @@ public class ConfigApplication
 
         public string? Name { get; set; }
 
+        [YamlMember(Alias = "app")]
+        public string? Application { get; set; }
```
rynowak: Let's talk about this. I had intended `name` to be the application name. I'm not sure what this is for.

samples/multi-yaml/fe/frontend/Order.cs
rynowak: That's pretty dope

src/Microsoft.Tye.Core/ApplicationBuilder.cs
davidfowl: From #29

> What does the name section of the manifest mean? Is that the overall application name or the name of this slice of the multi-service application?

When you "pack" your application the name is the package name, where app would be the overall application (used for tags etc)

samples/multi-yaml/fe/tye.yaml
davidfowl: This is just a sample to make sure it works as advertised.

PR #250, opened 2020-03-29
Add Dapr integration to Tye

rynowak:
Adds a new `extensions` integration point. Currently all extensbility has
to be inside the Tye codebase. There's no functionality for loading or distributing
plugins.

Enable dapr for an application like:

```yaml
name: test_app
extensions:
- name: dapr
```

The dapr extension currently accepts and requires no additional options, and applies
to all services defined in the application.

What it does:

- In local run: starts a dapr sidecar for each project that does http
- In local run: sidecars start in the directory of tye.yaml, so dapr will look for
  component manifests in ./components (relative to tye.yaml
- In deployment: adds required annotations to deployments

Some new features that were added to tye to enable this:

- Config: Ability to parse and conditionally run extensions
- Run: Ability to token replace env-vars into command line args
- Deploy: Ability to configure labels and annotations


src/Microsoft.Tye.Extensions/Microsoft.Tye.Extensions.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
rynowak: New project to keep us from leaking "extensions" code into the core.

samples/dapr/.tye/process_store
```diff
@@ -0,0 +1,6 @@
+{"pid":"80819"}
```
davidfowl: Remove? We should add .tye to .gitignore

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
```diff
@@ -0,0 +1,152 @@
...
+                    var metrics = new BindingBuilder()
+                    {
+                        AutoAssignPort = true,
+                        Name = "metrics",
+                        Protocol = "http",
+                    };
+                    proxy.Bindings.Add(metrics);
+
+                    // Set APP_PORT based on the project's assigned port for http
+                    if (httpBinding == null)
```
davidfowl: Didn't we already check this for null?

samples/dapr/.tye/process_store
rynowak: yes. We should consider moving this to user profile if we have to as well.

davidfowl:
Small nits but LGTM

samples/dapr/.tye/process_store
davidfowl: Sure the only thing I’m worried about is that it needs to be keyed by something so we can find it on the next run. If we don’t multiple runs by the same user can conflict

rynowak: A pretty simple way to do that is to key based on PID of the Host. So you get a hierarchy like:

```
$Home/.tye/pid/12345/<data we capture today>
$Home/.tye/pid/56789/<data we capture today>
```

Then the host can look for orphaned processed from another run by considering:

- pid of other running host process
- folders that exist
- (optional) timestamps of files

The only case where you'd have a problem is if the OS decided to reuse a PID for another tye host process.

PR #252, opened 2020-03-29
process needs to be disposed of

SimonCropp:


src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -347,6 +347,12 @@ private void WriteReplicaToStore(string pid)
             return null;
         }
 
+        private static string GetEntryPointFilePath()
+        {
+            using var process = Process.GetCurrentProcess();
```
davidfowl: This doesn't seem right.

davidfowl: Does this return a new instance every time?

SimonCropp: yep

```
public static Process GetCurrentProcess()
{
	return new Process(".", isRemoteMachine: false, NativeMethods.GetCurrentProcessId(), null);
}
```

SimonCropp: i think "returns new instance" is meant to be implied by the fact that it is a method and not a property

Issue #249, opened 2020-03-29
Pass additional arguments to docker run


tomasherceg:
### What should we add or change to make your life better?

I tried to add Mongo DB as a Docker container, but I couldn't mount the volume correctly on Windows using `tye.yml`. 

The problem is a combination of the following things:
* Mongo doesn't work with volumes mapped to Windows filesystem (`-v c:/something:/data/db`). It needs the named volume (which is mapped to a filesystem in the host VM) - I need to run `docker volume create --name=mongodata` (I can do it manually, it's a one-time operation), and then `docker run ... -v mongodata:/data/db`.
* When I use `mongodata` as `source` in `tye.yml`, it thinks that it's a relative path and make a full path for it - so `c:/something/mongodata` is passed)
* I haven't found a way how to pass additional arguments to `docker run` using `tye.yml` or how to work with named volumes

### Why is this important to you?

I can work-around this by running the container on my own, but I think that Mongo is quite a popular database so more people can run into this issue, and I also think that `tye.yml` could offer more flexibility because many containers need some special configuration.

Anyway, GREAT JOB! I love Tye so far. ❤

rynowak:
Thanks! Recategorizing this as a bug since something doesn't work.

davidfowl:
> When I use mongodata as source in tye.yml, it thinks that it's a relative path and make a full path for it - so c:/something/mongodata is passed)

I think this might be the think to fix here. We can support a -name on volumes instead of source.

> I haven't found a way how to pass additional arguments to docker run using tye.yml or how to work with named volumes

This is also something we should support.

davidfowl:
The PR for named volumes is up

davidfowl:
I've fixed the named volumes, support, will leave this open for docker run command support. I'm a bit hesitant to give that amount of power for now.

tomasherceg:
@davidfowl Wow, that was quick! The named volume support is great - this should help everyone who is using Mongo. I agree that giving the option to specify any docker run args can be dangerous.

rynowak:
Closing this because we don't plan to directly support configuring the args of `docker run` right now. If you have something you need, please ask, we'd love to talk to you.

PR #253, opened 2020-03-30
Add support for named docker volumes

davidfowl:
- Added tests

Contributes to #249

### DIFF
--- a/src/Microsoft.Tye.Core/ProjectReader.cs
+++ b/src/Microsoft.Tye.Core/ProjectReader.cs
@@ -211,9 +211,9 @@ namespace Microsoft.Tye
             output.WriteDebugLine($"IntermediateOutputPath={project.IntermediateOutputPath}");
 
             // Normalize directories to their absolute paths
-            project.IntermediateOutputPath = Path.Combine(project.ProjectFile.DirectoryName, project.IntermediateOutputPath);
-            project.TargetPath = Path.Combine(project.ProjectFile.DirectoryName, project.TargetPath);
-            project.PublishDir = Path.Combine(project.ProjectFile.DirectoryName, project.PublishDir);
+            project.IntermediateOutputPath = Path.Combine(project.ProjectFile.DirectoryName, project.IntermediateOutputPath.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar));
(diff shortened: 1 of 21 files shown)


test/E2ETest/TyeRunTests.cs
```diff
@@ -138,6 +139,101 @@ public async Task FrontendBackendRunTest()
...
+                Assert.Equal("Things saved to the volume!", await client.GetStringAsync(serviceUri));
+            });
+
+            // Delete the volume
+            await ProcessUtil.RunAsync("docker", $"volume rm {volumeName}");
+        }
+
+        [ConditionalFact]
+        [SkipIfDockerNotRunning]
+        public async Task DockerHostVolumeTest()
```
davidfowl: Good thing I wrote a test, `tye run --docker` didn't work on linux 😮

rynowak: We have a bug for that already. Did you fix it?

src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -258,5 +259,14 @@ private static void EvaluateProject(OutputContext output, ProjectServiceBuilder
...
             }
         }
+
+        private static string NormalizePath(string path)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return path.Replace('/', '\\');
+            }
+            return path.Replace('\\', '/');
```
rynowak: note: this isn't super right bcause `\` is a legal filename character in POSIX. There isn't really a good option unfortunately. I guess the answer is don't do that.

test/E2ETest/TyeRunTests.cs
davidfowl: Not yet, the host.docker.internal one was about communication between services in containers. This was because of path normalization.

test/E2ETest/TyeRunTests.cs
```diff
@@ -269,8 +361,51 @@ public async Task NullDebugTargetsDoesNotThrow()
 
             await host.StopAsync();
         }
+        private async Task<string> GetServiceUrl(HttpClient client, Uri serviceApi, string serviceName)
```
rynowak: needs more newline

PR #254, opened 2020-03-30
Work around host.docker.internal not working on linux

### LINKED ISSUE #176: We can't rely on host.docker.internal
### Describe the bug
The hostname `host.docker.internal` is provided via tye service discovery to applications when doing a tye run, so they can communicate with the *load balancer* rather than a specific replica of another service.

The problem is that the port/url in question is running inside the host, not inside the container. The hostname `host.docker.internal` is part of docker for windows/mac and is not there on linux.

See: https://github.com/docker/cli/issues/2290

### To Reproduce

Run the `FrontendBackendRunTestWithDocker` - it's skipped on linux.

### Options

Some options to fix this. It's possible to get the IP of the host machine and pass it in to docker using `--add-host`. I've tried a few flavours of this, but It's getting late, and I desparately want to get the other 10 tests passing and checked in.

davidfowl:
- Use the machine ip instead of host.docker.internal

PS: Will look into overwriting the host file in the container but its a tad bit hacky.

Fixes #176

### DIFF
--- a/test/E2ETest/TyeRunTests.cs
+++ b/test/E2ETest/TyeRunTests.cs
@@ -300,7 +300,6 @@ namespace E2ETest
 
         [ConditionalFact]
         [SkipIfDockerNotRunning]
-        [SkipOnLinux]
         public async Task FrontendBackendRunTestWithDocker()
         {
             var projectDirectory = new DirectoryInfo(Path.Combine(TestHelpers.GetSolutionRootDirectory("tye"), "samples", "frontend-backend"));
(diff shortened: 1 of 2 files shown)


test/E2ETest/TyeRunTests.cs
```diff
@@ -300,7 +300,6 @@ public async Task IngressRunTest()
 
         [ConditionalFact]
         [SkipIfDockerNotRunning]
-        [SkipOnLinux]
```
rynowak: 👏

PR #255, opened 2020-03-30
Make ports go away

davidfowl:
- Make AutoAssignPort true by default for binding and ingress. If a binding is defined auto assign the port. This will of course be ignored if there's a port defined.
- Don't auto assign a port if it's there's a connection string defined.

Allows things like this:

```yaml
services:
- name: frontend
  executable: node
  args: frontend\app.js
  bindings:
    - protocol: http
- name: backend
  executable: go
  args: run backend\main.go
  bindings:
    - protocol: http
```

### DIFF
--- a/src/Microsoft.Tye.Hosting/TyeDashboardApi.cs
+++ b/src/Microsoft.Tye.Hosting/TyeDashboardApi.cs
@@ -2,17 +2,17 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Text.Json;
+using System.Text.Json.Serialization;
 using System.Threading.Tasks;
-using Microsoft.Tye.Hosting.Model;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
-using System.Text.Json.Serialization;
-using System.Collections.Generic;
+using Microsoft.Tye.Hosting.Model;
 using Microsoft.Tye.Hosting.Model.V1;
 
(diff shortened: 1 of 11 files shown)


rynowak:
> Make AutoAssignPort true by default for binding and ingress. If a binding is defined auto assign the port.

Do we need the setting at all then? Why not take it out?

> Don't auto assign a port if it's there's a connection string defined.

We should start thinking about how we want these things to work in deployment when we change them. I'm worried that we're making too many small changes that oscillate around the right design.

Context: I was thinking about how this change impacts deployment, and realized that we're way too hardcoded today. We assume that a connectionString should be present for every container (non-.NET) project, and there's no way to control the behavior.

davidfowl:
> Context: I was thinking about how this change impacts deployment, and realized that we're way too hardcoded today. We assume that a connectionString should be present for every container (non-.NET) project, and there's no way to control the behavior.

I see, this doesn't impact that behavior though right?

> Do we need the setting at all then? Why not take it out?

Good call, I was thinking the same thing.

> We should start thinking about how we want these things to work in deployment when we change them. I'm worried that we're making too many small changes that oscillate around the right design.

We need some deployment tests... I agree though, most of it has been looking for a sweet spot in how much to describe. We have a mix of implicit and explicit behaviors:

- For ASP.NET Core projects, add 2 default bindings (http and https) if none specified.
- For projects, we assume that if there's no connection string on a binding and there is a binding then it must be http.
- (new) Assume that ports are autogenerated if there's no connection string for a binding, rather than error.

src/Microsoft.Tye.Hosting/TyeDashboardApi.cs
```diff
@@ -174,6 +173,7 @@ private static V1Service CreateServiceJson(Model.Service service)
                 {
                     Name = replica.Value.Name,
                     Ports = replica.Value.Ports,
+                    Environment = replica.Value.Environment
```
davidfowl: hmmm?

PR #269, opened 2020-03-30
remove tyeDir from PurgeHostAndWaitForGivenReplicasToStop

SimonCropp:
Note this is a breaking change. 

i cant tell if this is a bug in that it isnt used. or it is future proofing for when it will be needed?


rynowak:
> Note this is a breaking change. 

I'm not sure how this can be a breaking change, we don't have any APIs for public consumption 😆 yet. This is just test code.

My guess is that this changed when @davidfowl was doing some cleanup.

PR #270, opened 2020-03-30
Revert changes to Dockerfile step

jkotalik:
Reverting a part of #236

### DIFF
--- a/src/Microsoft.Tye.Core/GenerateDockerfileStep.cs
+++ b/src/Microsoft.Tye.Core/GenerateDockerfileStep.cs
@@ -13,16 +13,16 @@ namespace Microsoft.Tye
 
         public bool Force { get; set; }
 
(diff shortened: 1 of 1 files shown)


SimonCropp:
@jkotalik sorry. what did i break?

davidfowl:
@SimonCropp moving where things get printed. Before VS after. These sorts of micro-optimizations don't really move the needle for a tool like this.

SimonCropp:
@davidfowl ok. sorry about that. i was just reading the codebase, and submitting random things as i found them. i will stop doing the PR part now

davidfowl:
I don't want to discourage contribution but I would refrain from micro-optimizations at this point. We have bigger performance things to work on at the moment. Like running msbuild 😢

jkotalik:
Yeah, we encourage the contributions, this kind of optimization is very small when running and building takes over 5 seconds.

PR #272, opened 2020-03-30
optimise IsEmptyArray

SimonCropp:
This is a minor perf tweek. but i figured it might be in the hot path, so i would submit it anyway

 * combines an `is` and a cast into a pattern match
 * which means the `value is object` should also be redundant?
 * also the type param was never used


SimonCropp:
BTW. please tell me if these PRs are getting annoying

davidfowl:
@SimonCropp only a little. I would prefer if you could take a look at some of the filed issues. That would help us make progress. 

Also if you're interested here's the 0.1 board https://github.com/dotnet/tye/projects/3

PR #275, opened 2020-03-31
Pull docker images before running

### LINKED ISSUE #222: Pull the docker image before running
Today run implicitly pulls the image but the ux isn't great. See https://github.com/dotnet/tye/issues/200. Instead, we should pull the image explicitly before running.

davidfowl:
- This step also validates that docker is installed before trying to pull and before running so that logic can be removed from the DockerRunner.
- Let the application exit if it tye run fails to start.

Fixes #222


src/Microsoft.Tye.Hosting/DockerImagePuller.cs
```diff
@@ -0,0 +1,87 @@
+﻿using System;
```
rynowak: copyright

PR #276, opened 2020-03-31
Adds a very mad-katz metrics dashboard

rynowak:
image

### DIFF
--- a/src/Microsoft.Tye.Hosting/Dashboard/Pages/Index.razor
+++ b/src/Microsoft.Tye.Hosting/Dashboard/Pages/Index.razor
@@ -88,8 +88,6 @@
         {
             _subscriptions.Add(a.ReplicaEvents.Subscribe(OnReplicaChanged));
         }
-
-        base.OnInitialized();
     }
 
     private void OnReplicaChanged(ReplicaEvent replicaEvent)
--- a/src/Microsoft.Tye.Hosting/Dashboard/Shared/MetricsDisplay.razor
+++ b/src/Microsoft.Tye.Hosting/Dashboard/Shared/MetricsDisplay.razor
@@ -1,0 +1,74 @@
+@implements IDisposable
+@using System.Threading;
+
+<h4>Metrics</h4>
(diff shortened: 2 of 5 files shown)


src/Microsoft.Tye.Hosting/Dashboard/Pages/Index.razor
```diff
@@ -88,8 +88,6 @@
         {
             _subscriptions.Add(a.ReplicaEvents.Subscribe(OnReplicaChanged));
         }
-
-        base.OnInitialized();
```
rynowak: This is never needed in Blazor. We do good design :)

src/Microsoft.Tye.Hosting/Dashboard/Shared/MetricsDisplay.razor
```diff
@@ -0,0 +1,74 @@
...
+        replica = Service?.Replicas.FirstOrDefault().Value;
+        OnReplicaChanged(replica);
+    }
+
+    void OnReplicaChanged(ReplicaStatus? replica)
+    {
+        this.replica = replica;
+        if (this.replica == null && timer is object)
+        {
+            timer.Dispose();
```
rynowak: this would be cleaner with an observable, but it would require some new plumbing and probably use of immutabledictionary (or at least immutable snapshots as a pattern).

davidfowl: Instead of that can we move this logic elsewhere and hide it behind an IObservable/Subject?

rynowak: I did most of this but ran into a problem with only has solutions I hate...

The problem is that event counters fire once per-counter per-tick (the counters are not batched. 

So the solution could be to use immutable snapshots (`ImmutableDictionary<>`, and then fire once each time each counter updates. Then you'd want an observable that does buffering in between the producer and consumer. This would have more moving parts and use more resources than what we're doing.

If we scale up the number of consumers of the metrics then it makes sense to move to a push-based model.

Or if you know of a way to have eventcounters do batching 😆

PR #278, opened 2020-03-31
Add support for container networking

davidfowl:
- This adds all containers in the tye.yaml to the same container network.
- Container networking only works with a single replica today since the containers are named after their replicas.
- Use the container host name and port when injecting env variables. This handles the multiple replica case by falling back to the host ip and port.
- This cleans up container to container communication when migrating docker-compose files.
- Don't override the container port if already set.
- Remove StopAsync from TyeHost and exposed DisposeAsync. This patterns removes common clean up and hanging issues that occur when errors happen on start/stop
- Cleaned up run tests to use similar code to run, clean up and capture logs.

### DIFF
--- a/docs/redis.md
+++ b/docs/redis.md
@@ -95,7 +95,7 @@ We just showed how `tye` makes it easier to communicate between 2 applications r
       - port: 6379
     - name: redis-cli
       image: redis
-      args: "redis-cli -h host.docker.internal MONITOR"
+      args: "redis-cli -h redis MONITOR"
    ```
 
     We've added 2 services to the `tye.yaml` file. The `redis` service itself and a `redis-cli` service that we will use to watch the data being sent to and retrieved from redis.
--- a/samples/mongo-sample/tye.yaml
+++ b/samples/mongo-sample/tye.yaml
@@ -1,0 +1,21 @@
+services:
+- name: mongo
+  image: mongo
+  env:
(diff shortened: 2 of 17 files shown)


samples/mongo-sample/tye.yaml
```diff
@@ -0,0 +1,21 @@
...
+  image: mongo
+  env:
+    - name: ME_CONFIG_MONGODB_ADMINUSERNAME
+      value: root
+    - name: ME_CONFIG_MONGODB_ADMINPASSWORD
+      value: example
+  bindings:
+    - port: 27017
+- name: mongo-express
+  image: mongo-express
```
davidfowl: The mongo-express container image points to the host name mongo by default https://github.com/mongo-express/mongo-express#usage-docker

alexfdezsauco:
Here is the feature: https://github.com/dotnet/tye/issues/282

BTW, I noticed that also mix changes in this pull request from other. In case of feature approval I will create a new one.

docs/redis.md
```diff
@@ -95,7 +95,7 @@ We just showed how `tye` makes it easier to communicate between 2 applications r
       - port: 6379
     - name: redis-cli
       image: redis
-      args: "redis-cli -h host.docker.internal MONITOR"
+      args: "redis-cli -h redis MONITOR"
```
davidfowl, line 98: Cleans this up nicely!

PR #279, opened 2020-03-31
Network configuration for Docker services

alexfdezsauco:
Some docker container configuration requires a network. In this pull request a network parameter is added to the ConfigService and used to create the docker container.


davidfowl:
See https://github.com/dotnet/tye/pull/278.

alexfdezsauco:
Looks like #278 include network creation. This pull request only include the usage of an existing network.

alexfdezsauco:
@davidfowl is possible to use an existing network with #278?

davidfowl:
Nope. If you have a scenario for that. Please file an issue so we can better understand the need

alexfdezsauco:
Here is the feature: #282

BTW, I noticed that also mix changes in this pull request from other. In case of feature approval I will create a new one.

PR #281, opened 2020-03-31
DockerImagePuller checks the installed images first

alexfdezsauco:
DockerImagePuller checks if the image is already installed before pull it

### DIFF
--- a/src/Microsoft.Tye.Hosting/DockerImagePuller.cs
+++ b/src/Microsoft.Tye.Hosting/DockerImagePuller.cs
@@ -75,7 +75,7 @@ namespace Microsoft.Tye.Hosting
 
             bool installed = false;
             string output = string.Empty;
-            var processResult = await ProcessUtil.RunAsync(
+            await ProcessUtil.RunAsync(
                                     "docker",
                                     $"images --filter \"reference={name}:{version}\" --format \"{{{{.ID}}}}\"",
                                     outputDataReceived: data => installed = true,
--- a/src/Microsoft.Tye.Hosting/DockerImagePuller.cs
+++ b/src/Microsoft.Tye.Hosting/DockerImagePuller.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Microsoft.Tye.Hosting.Model;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Tye.Hosting/DockerImagePuller.cs
```diff
@@ -66,16 +67,39 @@ private async Task PullContainerAsync(string image)
...
+            }
+
+            string output = string.Empty;
+            await ProcessUtil.RunAsync(
+                             "docker",
+                             "images",
+                             outputDataReceived: data => output += data + '\n',
+                             throwOnError: false);
+
+            var matchCollection = Regex.Matches(output, @"([^\s]+)\s+([^\s]+).+");
```
davidfowl: Instead of this, use the --format and --filter parameters.

alexfdezsauco: Fixed

src/Microsoft.Tye.Hosting/DockerImagePuller.cs
```diff
@@ -66,16 +64,39 @@ private async Task PullContainerAsync(string image)
         {
             await Task.Yield();
 
+            string name = image;
+            string version = "latest";
+            var idx = image.LastIndexOf(":");
```
davidfowl, line 69: Do you need to do this? Doesn't reference handle it?

src/Microsoft.Tye.Hosting/DockerImagePuller.cs
```diff
@@ -66,16 +64,39 @@ private async Task PullContainerAsync(string image)
...
+            if (image.IndexOf("/") < idx)
+            {
+                name = image.Substring(0, idx);
+                version = image.Substring(idx + 1);
+            }
+
+            bool installed = false;
+            string output = string.Empty;
+            await ProcessUtil.RunAsync(
```
davidfowl, line 78: What does this return? What happens if the exit code is non zero?

src/Microsoft.Tye.Hosting/DockerImagePuller.cs
alexfdezsauco: No, also fixed. I just mixed the previous approach, it is not longer required.

src/Microsoft.Tye.Hosting/DockerImagePuller.cs
alexfdezsauco: Throws CommandException

src/Microsoft.Tye.Hosting/DockerImagePuller.cs
```diff
@@ -66,16 +64,34 @@ private async Task PullContainerAsync(string image)
...
+                                    "docker",
+                                    $"images --filter \"reference={image}\" --format \"{{{{.ID}}}}\"",
+                                    outputDataReceived: data => installed = true,
+                                    throwOnError: false);
+
+            if (result.ExitCode != 0)
+            {
+                throw new CommandException("Docker images command failed");
+            }
+
```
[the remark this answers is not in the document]
alexfdezsauco: When the image is not installed it will continue to pull the images. This command

    images --filter "reference={image} --format "{{.ID}}"

only output the single line that match with the filter condition. In such case I set the installed flag to true.

     outputDataReceived: data => installed = true,

for instance:

    >docker images --filter "reference=jboss/keycloak:6.0.1" --format "{{.ID}}"
    3a6718ca4ee0

I don't print the output, I just catch it to set the flag to true.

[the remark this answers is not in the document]
davidfowl: Though this is probably fine, it'd be more clear if we did a few things:
- Printed the output of this command images in case of failure.
- Get the output in the result and check if it's non-empty to determine if it's installed (remove the outputDataReceived callback and use result.Output)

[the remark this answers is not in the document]
alexfdezsauco: Fixed

davidfowl:
Thanks!

Issue #282, opened 2020-04-01
Network configuration for Docker services and usage of existing networks


alexfdezsauco:
### What should we add or change to make your life better?

For a reason I just created a docker container and a network that its life-cycle is not controlled by Tye.  For instance I have a postgres container that live forever on my development computer with a lot of databases.

But at some point I want to integrate a Tye hosted solution to an existing network. For instance, the solution includes a keycloak but I want to use an existing database that is actually in the postgres container not hosted by Tye.  

keycloak and postgres are a good scenario because,  at least on docker for windows, both containers have to use the same network.  So, I created a network outside Tye and I want launch a container with Tye that reference such network.

### Why is this important to you?

Because my development machine could be a mess and I want to maximize the options.

davidfowl:
I'd prefer it if containers weren't individually configurable on a specific network but I'm find connecting all containers to an existing network that can be specified. Ideally tye wouldn't end up being just as noisy as docker-compose (that's actually the point) but letting you specify the network is a fine way to allow containers created by tye to integrate with another container network.

How does that sound?

alexfdezsauco:
It sounds good.

davidfowl:
Done!

PR #287, opened 2020-04-01
Network configuration for Docker services and usage of existing networks

alexfdezsauco:
#282 Add configuration option to enable the containers usage of an existing network


src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -51,8 +51,32 @@ public async Task StartAsync(Application application)
 
             string? dockerNetwork = null;
 
+            if (!string.IsNullOrWhiteSpace(application.Network))
```
davidfowl, line 54: IsNullOrWhiteSpace -> IsNullOrEmpty

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -94,7 +118,7 @@ public async Task StopAsync(Application application)
 
             await Task.WhenAll(tasks);
 
-            if (application.Items.TryGetValue("dockerNetwork", out var dockerNetwork))
+            if (string.IsNullOrWhiteSpace(application.Network) && application.Items.TryGetValue("dockerNetwork", out var dockerNetwork))
```
davidfowl, line 121: Same

davidfowl:
This change also needs a test.

src/Microsoft.Tye.Hosting/DockerRunner.cs
alexfdezsauco: Fixed

src/Microsoft.Tye.Hosting/DockerRunner.cs
alexfdezsauco: Fixed

src/Microsoft.Tye.Core/ConfigModel/ConfigApplication.cs
```diff
@@ -21,6 +21,8 @@ public class ConfigApplication
 
         public string? Registry { get; set; }
 
+        public string? Network { get; set; }
```
davidfowl, line 24: @rynowak how do you feel about this? I don't love it being so top level and being specific to containers but I don't have an alternative yet.

alexfdezsauco:
> Can you also write a test?

Yes, I can. I'm getting related with Tye testing approach (kind of integration tests). 

Basically I have to test if a network exists the docker execution doesn't create a new one and doesn't dispose it at the end. I'm thinking to create a docker only yaml file but it will require at least one image (or to download such image) to run.

davidfowl:
> Basically I have to test if a network exists the docker execution doesn't create a new one and doesn't dispose it at the end. I'm thinking to create a docker only yaml file but it will require at least one image (or to download such image) to run.

Model it after the volume test https://github.com/dotnet/tye/blob/6671da72c0f7ef83ddabcd8d44e2e0ebb7d0e3f2/test/E2ETest/TyeRunTests.cs#L193.

Create a network at the beginning of the test with a random name and programmatically change the model to use it. Then after it works, make a rest API call to get the container id and then use docker inspect to make sure the container is on the right network (or maybe use the REST API that tye exposes).

test/E2ETest/E2ETest.csproj
```diff
@@ -40,6 +40,10 @@
     <None Remove="testassets\generate\single-project.yaml" />
     <None Remove="testassets\init\frontend-backend.yaml" />
     <None Remove="testassets\init\multi-project.yaml" />
+    <None Remove="testassets\projects\network-test\appsettings.Development.json" />
```
davidfowl, line 43: Remove?

test/E2ETest/testassets/projects/network-test/Startup.cs
```diff
@@ -0,0 +1,54 @@
...
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapGet("/", async context =>
```
davidfowl, line 34: Change the sample so it isn't using volumes. That isn't what's being tested here.

test/E2ETest/TyeRunTests.cs
```diff
@@ -240,6 +244,80 @@ public async Task DockerNamedVolumeTest()
             await ProcessUtil.RunAsync("docker", $"volume rm {volumeName}");
         }
 
+        [ConditionalFact]
+        [SkipIfDockerNotRunning]
+        public async Task DockerNetworkAssignmentTest()
+        {
+            using var projectDirectory = CopyTestProjectDirectory("network-test");
```
davidfowl, line 251: I would actually use the frontend-backend example. That does container to container communication and is written for you already. Forgot to mention that earlier.

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -51,8 +51,32 @@ public async Task StartAsync(Application application)
...
+
+                if (!string.IsNullOrWhiteSpace(dockerNetworkResult.StandardOutput))
+                {
+                    _logger.LogInformation("The specified network {Network} exists", application.Network);
+
+                    dockerNetwork = application.Network;
+                }
+                else
+                {
+                    _logger.LogInformation("The specified network {Network} doesn't exist.", application.Network);
```
davidfowl, line 72: LogError?

test/E2ETest/TyeRunTests.cs
```diff
@@ -21,6 +21,10 @@
 
 namespace E2ETest
 {
+    using Newtonsoft.Json;
+
+    using JsonSerializer = System.Text.Json.JsonSerializer;
```
davidfowl, line 26: Undo?

test/E2ETest/TyeRunTests.cs
davidfowl: Let me be more specific, there's a test called `FrontendBackendRunTestWithDocker`. Copy that test and add the custom code to create a network.

alexfdezsauco:
Fix according with the revision. But there is an issue with whitespace formatting, but not sure why.

alexfdezsauco:
Fixed

davidfowl:
Did you remove the check to make sure the app is working? Keep the original assertions...

alexfdezsauco:
Fixed

davidfowl:
I’d recommend pinging the backend before the frontend to reduce potential flakiness. This was an intentional change that was made in previous PRs because of random failures

davidfowl:
Thanks!

PR #289, opened 2020-04-02
Add a library for `AddTyeSecrets`

rynowak:
This removes the need to copy-paste code that reads a specific directory
in Program.cs.

Updated docs/samples and tests.

We need a workaround in tests to be able to use a P2P for the library. What
I did here is the same trick we do in Razor.

- Force everyone to use an MSBuild variable to locate the library
- Set that variable in a Directory.Build.props for the normal build
- Drop a special Directory.Build.props for testing

Updates a bunch of test code to use new helper methods for copying
stuff, so we can correctly do this.


rynowak:
Currently working on the second part of this - getting rid of the copy-paste for reading config

PR #292, opened 2020-04-02
Added container proxy for talking to the host

davidfowl:
- This change introduces a container proxy which makes it possible for docker containers can talk to host services using container networking. These proxies will not show up in the dashboard as they are "infrastructure" containers.
- Added Private and NetworkAlias to DockerRunInfo but did not expose these configuration.

This is what the topology looks like:

Without replicas
image

With replicas

image

### DIFF
--- a/samples/nginx-ingress/ApplicationB/Startup.cs
+++ b/samples/nginx-ingress/ApplicationB/Startup.cs
@@ -1,0 +1,40 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace ApplicationB
+{
+    public class Startup
+    {
+        // This method gets called by the runtime. Use this method to add services to the container.
+        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
+        public void ConfigureServices(IServiceCollection services)
+        {
+        }
+
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+        {
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapGet("/", async context =>
+                {
+                    await context.Response.WriteAsync("Hello from Application B " + Environment.GetEnvironmentVariable("APP_INSTANCE") ?? Environment.GetEnvironmentVariable("HOSTNAME"));
+                });
+            });
+        }
+    }
+}
--- a/src/Microsoft.Tye.Hosting/AggregateApplicationProcessor.cs
+++ b/src/Microsoft.Tye.Hosting/AggregateApplicationProcessor.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
 using Microsoft.Tye.Hosting.Model;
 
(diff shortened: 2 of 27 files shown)


src/Microsoft.Tye.Hosting/AggregateApplicationProcessor.cs
```diff
@@ -28,10 +29,28 @@ public async Task StartAsync(Application application)
 
         public async Task StopAsync(Application application)
         {
+            var exceptions = new List<Exception>();
```
rynowak, line 32: SPICY but...

This is why the composite pattern is the worst. We probably would have already had this code in place if not for the design *feeling simple* when it's not.

samples/nginx-ingress/ApplicationB/Startup.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace ApplicationB
+{
+    public class Startup
+    {
+        // This method gets called by the runtime. Use this method to add services to the container.
+        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
+        public void ConfigureServices(IServiceCollection services)
```
jkotalik, line 17: Why do we need another sample?

davidfowl: As opposed to?

src/Microsoft.Tye.Hosting/AggregateApplicationProcessor.cs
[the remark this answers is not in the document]
davidfowl: Yeahhh...

rynowak, line 32: SPICY but...

This is why the composite pattern is the worst. We probably would have already had this code in place if not for the design *feeling simple* when it's not.

samples/nginx-ingress/ApplicationB/Startup.cs
rynowak: oh snap

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -49,6 +49,49 @@ public async Task StartAsync(Application application)
...
+            foreach (var service in application.Services.Values)
+            {
+                if (service.Description.RunInfo is DockerRunInfo || service.Description.Bindings.Count == 0)
+                {
+                    continue;
+                }
+
+                // Inject a proxy per non-container service. This allows the container to use normal host names within the
+                // container network to talk to services on the host
+                var proxyContanier = new DockerRunInfo($"mcr.microsoft.com/dotnet/core/sdk:3.1", "dotnet Microsoft.Tye.Proxy.dll")
```
rynowak, line 62: This is real spicy 👍

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -49,6 +49,49 @@ public async Task StartAsync(Application application)
...
+                        Host = binding.Host,
+                        ContainerPort = binding.ContainerPort,
+                        Name = binding.Name,
+                        Port = binding.Port,
+                        Protocol = binding.Protocol
+                    };
+                    b.ReplicaPorts.Add(b.Port.Value);
+                    proxyDescription.Bindings.Add(b);
+                }
+                var proxyContanierService = new Service(proxyDescription);
```
rynowak, line 90: oops

src/Microsoft.Tye.Proxy/Microsoft.Tye.Proxy.csproj
```diff
@@ -0,0 +1,22 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Bedrock.Framework" Version="0.1.38-alpha.gd25d5b37ad" />
+    <PackageReference Include="System.IO.Pipelines" Version="4.7.0" />
+  </ItemGroup>
+
+  <!-- Include *.deps.json and *.runtimeconfig.json in ContentWithTargetPath so they will be copied to the output folder of projects
+       that reference this one. -->
+  <Target Name="AddRuntimeDependenciesToContent" Condition=" '$(TargetFrameworkIdentifier)' == '.NETCoreApp'" BeforeTargets="GetCopyToOutputDirectoryItems">
```
rynowak, line 15: This is a cool trick that I haven't seen before. We did something much more complex in Razor SDK :-|

src/Microsoft.Tye.Proxy/Program.cs
```diff
@@ -0,0 +1,135 @@
+﻿using System;
```
rynowak, line 1: copyright bruh

davidfowl: Gah every time

PR #293, opened 2020-04-02
Link Code of Conduct

terrajobst:
See [this policy rule](https://github.com/dotnet/org-policy/blob/master/doc/PR15.md) for more details.


jkotalik:
We currently have the code of conduct in the README: https://github.com/dotnet/tye#net-foundation. If it needs to be in a separate file, that's fine.

The aspnetcore repo doesn't have this file either, as an FYI.

jkotalik:
Nvrm, looks like a PR has already been made there as well: https://github.com/dotnet/aspnetcore/pull/20471

Issue #232, opened 2020-03-27
Configuration input and model validation


davidfowl:
Today invalid configuration show awful hard to read de-serialization exceptions in the console. Once that succeeds, it's possible to have an invalid configuration model and we don't have much validation in place today (besides very basic required fields, like the name of services).

We need to do both input validation and model validation:
- **Input validation** - Making sure the configuration schema is correct (type checking).
- **Model validation** - Making sure the configuration is valid after it is correctly parsed.

**Input Validation**
- [x] - Improve the input validation we need to write custom de-serialization logic and give nice errors with the line number in the file etc.
- [x] - Replicas can't be negative

**Model Validation**
- [x] - https://github.com/dotnet/tye/issues/106 Tye should throw when multiple bindings are specified without names 
- [x] - All services should have a name (it's required)
- [x] - Ingress protocol should be http or https
- [x] - Ingress must point to a valid service
- [x] - Multiple bindings cannot have the same name
- [x] - Services cannot have the same name
- [x] - Services cannot be bound to the same port
- [x] - For services, image, project and executable are mutually exclusive

davidfowl:
Did you implement all of the model validation rules?

jkotalik:
Thx.

rynowak:
bop it.

PR #294, opened 2020-04-03
Implement custom deserialization and validation logic

jkotalik:
Fixes https://github.com/dotnet/tye/issues/232.


src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -244,84 +243,5 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
 
             return builder;
         }
-
-        private static void ValidateConfigApplication(ConfigApplication config)
```
davidfowl, line 248: Why did you remove this?

src/Microsoft.Tye.Core/ConfigModel/ConfigApplication.cs
```diff
@@ -28,5 +33,139 @@ public class ConfigApplication
...
+                throw new TyeYamlException(
+                    "Configuration validation failed." + Environment.NewLine +
+                    string.Join(Environment.NewLine, results.Select(r => r.ErrorMessage)));
+            }
+
+            foreach (var extension in config.Extensions)
+            {
+                if (!extension.TryGetValue("name", out var name) || string.IsNullOrWhiteSpace(name as string))
+                {
+                    throw new TyeYamlException(CoreStrings.ExtensionMustProvideAName);
```
rynowak, line 55: oh no resources. lol

Do you think there's a really important reason we use resources?

src/Microsoft.Tye.Core/ConfigModel/ConfigApplication.cs
```diff
@@ -28,5 +33,139 @@ public class ConfigApplication
...
+                            $"Volume '{volume.Source}' of service '{service.Name}' validation failed." + Environment.NewLine +
+                            string.Join(Environment.NewLine, results.Select(r => r.ErrorMessage)));
+                    }
+                }
+            }
+
+            foreach (var ingress in config.Ingress)
+            {
+                // We don't currently recurse into ingress rules or ingress bindings right now.
+                // There's nothing to validate there.
```
rynowak, line 120: Comment seems stale now

src/Microsoft.Tye.Core/ConfigModel/ConfigIngress.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Tye.ConfigModel
     public class ConfigIngress
     {
         [Required]
-        public string Name { get; set; } = default!;
+        public string? Name { get; set; } = default!;
```
rynowak, line 13: Remove `default!` if you're going to make these nullable. 

The point of how this worked in the past:

The nullability contract + `[Required]` reflect the **valid** state of the object. That's what `default!` is used. If you're going to treat the properties as nullable, there's no need to initialized them explicitly.

src/Microsoft.Tye.Core/Microsoft.Tye.Core.csproj
```diff
@@ -32,6 +32,12 @@
     <Content Include="Templates\**" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
 
+  <ItemGroup>
+    <EmbeddedResource Update="CoreStrings.resx" EmitFormatMethods="true">
+      <Generator></Generator>
```
rynowak, line 37: Isn't this supposed to be set to something?

src/Microsoft.Tye.Core/Serialization/TyeYamlException.cs
```diff
@@ -0,0 +1,32 @@
...
+            : this(start,  message, null)
+        {
+        }
+
+        public TyeYamlException(Mark start, string message, Exception? innerException)
+            : base($"Error parsing tye.yaml: ({start.Line}, {start.Column}): {message}", innerException)
+        {
+        }
+
+        public TyeYamlException(string message, Exception inner)
```
rynowak, line 27: `Exception?`

src/Microsoft.Tye.Core/Serialization/YamlConfigApplicationHelpers.cs
```diff
@@ -0,0 +1,65 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Tye.ConfigModel;
+using YamlDotNet.RepresentationModel;
+
+namespace Tye.Serialization
+{
+    public static class YamlConfigApplicationHelpers
+    {
+
```
rynowak, line 12: @jamesnk

src/Microsoft.Tye.Core/Serialization/YamlConfigApplicationHelpers.cs
```diff
@@ -0,0 +1,65 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Tye.ConfigModel;
+using YamlDotNet.RepresentationModel;
+
+namespace Tye.Serialization
+{
+    public static class YamlConfigApplicationHelpers
```
rynowak, line 10: Please no.

src/Microsoft.Tye.Core/Serialization/YamlConfigApplicationHelpers.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        public static void HandleConfigApplication(YamlMappingNode? yamlMappingNode, ConfigApplication app)
+        {
+            foreach (var child in yamlMappingNode!.Children)
+            {
+                string? key;
+                switch (child.Key.NodeType)
+                {
+                    case YamlNodeType.Scalar:
+                        key = (child.Key as YamlScalarNode)!.Value;
```
rynowak, line 21: Don't combine an `as` cast with `!`. You want this to be a c-style cast so it will fail with a good error instead of a nullref.

src/Microsoft.Tye.Core/Serialization/YamlConfigApplicationHelpers.cs
jkotalik: What do you mean?

rynowak: A good rule. You can remove the word `Helpers` and it's suddenly a better name.

rynowak: `ConfigApplicationParser` would be a good name

src/Microsoft.Tye.Core/Microsoft.Tye.Core.csproj
jkotalik: Directory.Build.targets sets stuff and it's filled in with arcade

src/Microsoft.Tye.Core/Serialization/YamlExtensionHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using YamlDotNet.RepresentationModel;
+
+namespace Tye.Serialization
+{
+    public static class YamlExtensionHelper
+    {
+        internal static void HandleExtensionsMapping(YamlSequenceNode yamlSequenceNode, List<Dictionary<string, object>> extensions)
```
rynowak, line 12: Can the class be internal instead of the method? It's pretty rare to want a public type with internal methods.

src/Microsoft.Tye.Core/Serialization/YamlExtensionHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+                            {
+                                case YamlNodeType.Scalar:
+                                    key = (mapping.Key as YamlScalarNode)!.Value;
+                                    break;
+                                default:
+                                    throw new TyeYamlException(mapping.Key.Start,
+                                        CoreStrings.FormatUnexpectedType(YamlNodeType.Scalar.ToString(), mapping.Key.NodeType.ToString()));
+                            }
+
+                            if (key != null)
```
rynowak, line 33: why  do we need a null check here? This seems like it should either not happen or be an error if it does.

src/Microsoft.Tye.Core/ConfigModel/ConfigApplication.cs
jkotalik: I like them for having consistent strings and sharing strings between src and test.

What other mechanism would you prefer?

src/Microsoft.Tye.Core/Serialization/YamlIngressHelpers.cs
```diff
@@ -0,0 +1,189 @@
...
+                    default:
+                        throw new TyeYamlException(child.Start,
+                            CoreStrings.FormatUnexpectedType(YamlNodeType.Mapping.ToString(), child.NodeType.ToString()));
+                }
+            }
+        }
+
+        private static void HandleIngressMapping(YamlMappingNode? yamlMappingNode, ConfigIngress configIngress)
+        {
+            foreach (var child in yamlMappingNode!.Children)
```
rynowak, line 33: nullability tip. If you find you find yourself using defining a nullable parameter and then immediately asserting it's non-null then the parameter should not be nullable. The caller needs to do the checking.

jkotalik: I think this is just inconsistent. I have other places where I removed the nullability check. I need to do a second pass of this when I have a fresh mind as there is some low hanging fruit.

src/Microsoft.Tye.Core/Serialization/YamlParser.cs
```diff
@@ -0,0 +1,97 @@
...
+            {
+                throw new TyeYamlException(node.Start, CoreStrings.FormatExpectedYamlScalar(key));
+            }
+
+            return (node as YamlScalarNode)!.Value;
+        }
+
+        public void Dispose()
+        {
+            _reader.Dispose();
```
rynowak, line 94: nitpick: This isn't really right. The problem is that you have two constructors where **we** create the Reader, and thus **own** it, and one constructor where it's passed in, and thus we don't **own** it. 

The best thing to do IMO is not be disposable. If you want to support passing in a string for tests, you don't need dispose for that :)

jkotalik:
> This is great. I left some comments about ideas to clean this up, but they aren't really major, more like whines 👍 I'll leave it up to you which ones you want to deal with.

I like the wines 🍷

src/Microsoft.Tye.Core/ConfigModel/ConfigApplication.cs
rynowak: I hadn't thought about sharing with tests. Makes good sense 👍

src/Microsoft.Tye.Core/Microsoft.Tye.Core.csproj
rynowak: OK. I'm just not sure why we need the attribute and for it to be empty.

src/Microsoft.Tye.Core/Serialization/YamlConfigApplicationHelpers.cs
rynowak: > What do you mean?

No helpers. Pick better names or don't make classes 😆

src/Microsoft.Tye.Core/Microsoft.Tye.Core.csproj
davidfowl: Agree it’s triggering me

jkotalik: I see now, let me try and remove it.

src/Microsoft.Tye.Core/Serialization/YamlParser.cs
jkotalik: The constructor which takes in a TextReader was supposed to be internal/private, thus the parser owns the TextReader.

PR #297, opened 2020-04-03
Centralize some property and fix Extensions.Config

rynowak:


rynowak:
M.T.E.Config wasn't getting published because were manually managing `IsPackable` on a per-project basis.

Issue #296, opened 2020-04-03
How can I inject dapr.io/config: "tracing" using tye?


thangchung:
Currently, I use `tye` to deploy my application to AKS. Everything is fine now. But with Dapr in-place we can do the tracing out of the box. See at https://github.com/dapr/docs

So I have a quick question that how can we inject `dapr.io/config: "tracing"` into the deployment to enable the tracing with Open Telemetry using `tye`?

davidfowl:
Are you talking about adding annotations for deployment using tye?

thangchung:
Yes, it is. Just adding more like below

```yaml
annotations:
	...
    dapr.io/config: "tracing"
    dapr.io/log-level: "debug"
```

davidfowl:
cc @rynowak

rynowak:
@thangchung - thanks for [blogging us](https://dev.to/thangchung/debugging-dapr-application-using-tye-tool-1djb), I enjoyed the read, and I'm glad you find our project interesting. 

---

We haven't done anything to enable this yet. 

Some ideas:

- We will eventually add some features to customize the k8s manifests annotations and labels, but this doesn't seem like a great case for that because....
- We can add per-extension settings that don't necessary map 1:1 to k8s, or have a different meaning in local development

Example:

```yaml
extensions:
- name: dapr
  config: myconfig
  log-level: debug
```

This would map to the annotations you defined above when deployed, and could map to `config myconfig -log-level=debug` when running locally. 

This seems like it should *just work* would all you to have access to all of dapr's config settings in local dev. 

If you're interested in sending a PR - all the changes would go into [DaprExtension](https://github.com/dotnet/tye/blob/master/src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs). Otherwise, I can add this sometime in the next few days.

rynowak:
Keeping open because I need to fix something with the local dev part of this. Deployment support is merged. Thanks @thangchung 👍

PR #298, opened 2020-04-04
Add debug and tracing annotations

thangchung:
This PR adds `dapr.io/config` and `dapr.io/log-level` into the annotations section in dapr.yaml. Related to https://github.com/dotnet/tye/issues/296

### DIFF
--- a/src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
+++ b/src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
@@ -138,19 +138,14 @@ namespace Microsoft.Tye.Extensions.Dapr
                         deployment.Annotations.Add("dapr.io/id", project.Name);
                         deployment.Annotations.Add("dapr.io/port", (httpBinding.Port ?? 80).ToString(CultureInfo.InvariantCulture));
 
-                        if (!config.Data.Any())
+                        if (config.Data.TryGetValue("config", out var daprConfig) && daprConfig is object)
                         {
-                            continue;
+                            deployment.Annotations.TryAdd("dapr.io/config", daprConfig.ToString());
                         }
 
-                        if (config.Data.TryGetValue("config", out var daprConfig))
+                        if (config.Data.TryGetValue("log-level", out var logLevel) && logLevel is object)
                         {
-                            deployment.Annotations.Add("dapr.io/config", (daprConfig.ToString() ?? "tracing").ToString(CultureInfo.InvariantCulture));
-                        }
-
-                        if (config.Data.TryGetValue("log-level", out var logLevel))
-                        {
-                            deployment.Annotations.Add("dapr.io/log-level", (logLevel.ToString() ?? "info").ToString(CultureInfo.InvariantCulture));
+                            deployment.Annotations.TryAdd("dapr.io/log-level", logLevel.ToString());
                         }
                     }
                 }
--- a/src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
+++ b/src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Tye.Extensions.Dapr
 
                         // These environment variables are replaced with environment variables
                         // defined for this service.
-                        Args = $"-app-id {project.Name} -app-port %APP_PORT% -dapr-grpc-port %DAPR_GRPC_PORT% --dapr-http-port %DAPR_HTTP_PORT% --metrics-port %METRICS_PORT% --placement-address localhost:50005",
+                        Args = $"-app-id {project.Name} -app-port %APP_PORT% -dapr-grpc-port %DAPR_GRPC_PORT% --dapr-http-port %DAPR_HTTP_PORT% --metrics-port %METRICS_PORT% -log-level %LOG_LEVEL% -config %CONFIG% --placement-address localhost:50005",
                     };
                     context.Application.Services.Add(proxy);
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
```diff
@@ -137,6 +137,21 @@ public override Task ProcessAsync(ExtensionContext context, ExtensionConfigurati
...
                         deployment.Annotations.Add("dapr.io/port", (httpBinding.Port ?? 80).ToString(CultureInfo.InvariantCulture));
+
+                        if (!config.Data.Any())
+                        {
+                            continue;
+                        }
+
+                        if (config.Data.TryGetValue("config", out var daprConfig))
+                        {
+                            deployment.Annotations.Add("dapr.io/config", (daprConfig.ToString() ?? "tracing").ToString(CultureInfo.InvariantCulture));
```
rynowak, line 148: `daprConfig.ToString() ?? "tracing"` - this doesn't seem like a very good default because it's not something someone could guess. 

What would you think about putting this into the `if` so that we don't add any annotation when this is `null`?

```C#
if (config.Data.TryGetValue("config", out var daprConfig) && daprConfig is object)
{
...
}
```

Same for the block below, I'd rather not put anything in the yaml if it's the default.

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
```diff
@@ -137,6 +137,21 @@ public override Task ProcessAsync(ExtensionContext context, ExtensionConfigurati
                         deployment.Annotations.Add("dapr.io/enabled", "true");
                         deployment.Annotations.Add("dapr.io/id", project.Name);
                         deployment.Annotations.Add("dapr.io/port", (httpBinding.Port ?? 80).ToString(CultureInfo.InvariantCulture));
+
+                        if (!config.Data.Any())
+                        {
+                            continue;
+                        }
```
rynowak, line 144: nitpick: this `if` isn't really doing anything useful. If the dictionary is empty both of the following blocks will be skipped anyway because they won't find the value they're looking for.

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
```diff
@@ -137,6 +137,21 @@ public override Task ProcessAsync(ExtensionContext context, ExtensionConfigurati
                         deployment.Annotations.Add("dapr.io/enabled", "true");
                         deployment.Annotations.Add("dapr.io/id", project.Name);
                         deployment.Annotations.Add("dapr.io/port", (httpBinding.Port ?? 80).ToString(CultureInfo.InvariantCulture));
+
```
rynowak, line 140: Do you have any interest in making this work for local dev as well? You could add the same thing to the command line here: https://github.com/dotnet/tye/pull/298/files#diff-e1a701ef2a3ac88566fe0e681b1e224fR40

If you don't feel like doing this I'll do it in a follow-up after this PR is merged.

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
thangchung: @rynowak Are you happy if I add as below

```csharp
if (config.Data.TryGetValue("config", out var daprConfig) && daprConfig is object)
{
    deployment.Annotations.Add("dapr.io/config", daprConfig.ToString()?.ToString(CultureInfo.InvariantCulture));
}

if (config.Data.TryGetValue("log-level", out var logLevel) && logLevel is object)
{
    deployment.Annotations.Add("dapr.io/log-level", logLevel.ToString()?.ToString(CultureInfo.InvariantCulture));
}
```

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
thangchung: I will remove it

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
thangchung: Could you clarify more? I don't get the point of `making this work for local dev`

thangchung: Please help me with it. Thank you.

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
thangchung: Currently, we have checked `is object` in the if statement, but the `daprConfig` and `logLevel` is possible null and now allow to build. Could I put `#pragma warning disable CS8604 // Possible null reference argument.` to disable this check?

rynowak: I think the reason why you're getting that warning is because you're calling `ToString()` twice.

You can simplify `logLevel.ToString()?.ToString(CultureInfo.InvariantCulture)` to `logLevel.ToString()`.

If that doesn't fix it, feel free to suppress the warning because that seems like a bug.

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
rynowak: local dev => `tye run`. You'd need to pass the commandline arguments to `daprd` - ex: `-log-level info -config myConfig`

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
thangchung: After I do a suppress that warning then everything built okay, except for the format tool :|

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
```diff
@@ -138,15 +138,17 @@ public override Task ProcessAsync(ExtensionContext context, ExtensionConfigurati
                         deployment.Annotations.Add("dapr.io/id", project.Name);
                         deployment.Annotations.Add("dapr.io/port", (httpBinding.Port ?? 80).ToString(CultureInfo.InvariantCulture));
 
+#pragma warning disable CS8604 // Possible null reference argument.
```
davidfowl, line 141: That's usally done by doing this:

```
daprConfig!.ToString()
```

thangchung: I did it with your suggestion, but no luck. Could anyone tell me how can I overcome this check?

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
thangchung: Did that. Please check it

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
thangchung: I did `deployment.Annotations.TryAdd("dapr.io/config", daprConfig!.ToString() ?? string.Empty);`, and it can overcome this check but not sure it is correctly

rynowak:
I'll clone this and take a look. let's try to get this done 👍

rynowak:
This looks close enough and you've got the basics done. I'm going to merge this and then I'll fix the formatting.

PR #299, opened 2020-04-05
Implement library support for service discovery

rynowak:
Implements three flavor of library support for service-discovery:

- GetConnectionString (arbitrary strings) augmenting existing
  functionality already in asp.net core
- GetServiceUri (uris) can be set via connection string or combined
  from a protocol/host/port triple
- GetServiceHost (host/authority) can be set via connection string or
  combined from a host/port pair

This uses the term `host` even though the correct technical term is
`authority`. There are two good reasons for this:

- you should always question authority
- people don't really know that term xD


rynowak:
Work in progress while I do some:
- [ ] cleanup on samples
- [ ] double-check the tutorial
- [ ] create an issue for breaking changes
- [ ] test that the samples work 👍 
- [ ] write the service discovery docs

docs/redis.md
```diff
@@ -64,7 +64,7 @@ We just showed how `tye` makes it easier to communicate between 2 applications r
 
        services.AddStackExchangeRedisCache(o =>
        {
-            var connectionString = Configuration["connectionstring:redis"];
+            var connectionString = Configuration["connectionstrings:redis"];
```
davidfowl, line 67: Use the extension method.

src/Microsoft.Tye.Extensions.Configuration/TyeConfigurationExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+namespace Microsoft.Extensions.Configuration
+{
+    public static class TyeConfigurationExtensions
+    {
+        public static Uri? GetServiceUri(this IConfiguration configuration, string name, string? binding = null)
+        {
+            var key = GetKey(name, binding);
+
+            var connectionString = configuration[$"connectionstrings:{key}"];
+            if (!string.IsNullOrEmpty(connectionString))
```
davidfowl, line 16: Is this weird? Feels weird.

docs/redis.md
rynowak: image

src/Microsoft.Tye.Extensions.Configuration/TyeConfigurationExtensions.cs
rynowak: which part? 

I haven't written the docs yet, but here's what I was going to explain.

> Use the method (`GetConnectionString`, `GetServiceUri`, `GetServiceHost`) **that best describes the kind of config/secret you need to access**.

We need **something** that does this, because in the deployment flow we don't know what code you wrote in your app. Ex: the configuration used to access redis is commonly a hostname/port in development, and commonly a connection string (arbitrary text) in production.

To me it doesn't seem weird once you have the realization that:

- We don't/can't know what code is in your app
- We don't know how to decompose arbitrary text (secrets) into host/port/protocol

---

The potential for mistakes does exist here btw, because imagine that you used `GetServiceHost` for redis. That won't work in production. So we probably need more guidance.

docs/redis.md
davidfowl: https://github.com/dotnet/tye/pull/299#issuecomment-609501437

src/Microsoft.Tye.Extensions.Configuration/TyeConfigurationExtensions.cs
rynowak: We had a discussion about this more in teams, and ended up with this as the best proposal:

- Don't allow connection strings to override GetServiceUri
- Get rid of GetServiceHost (use GetConnectionString instead for things like redis)
- In dev, we populate CONNECTIONSTRING_REDIS to be the host:port (prepend the protocol if specified)

samples/voting/worker/Extensions.cs
```diff
@@ -10,7 +10,7 @@ public static class Extensions
     {
         public static string GetSqlConnectionString(this IConfiguration configuration)
         {
-            return configuration["connectionstring:postgres"] ?? $"Server={configuration["service:postgres:host"]};Port={configuration["service:postgres:port"]};User Id=postgres;Password=pass@word1;";
+            return configuration["connectionstrings:postgres"] ?? $"Server={configuration["service:postgres:host"]};Port={configuration["service:postgres:port"]};User Id=postgres;Password=pass@word1;";
```
rynowak, line 13: @davidfowl - this is something we should decide on...

----

Let me propose an ideal

I think the ideal thing here would be to use templating in the YAML for this.

So you'd write:

```yaml
services:
- name: postgres
  image:  postgres
  env:
    - name: POSTGRES_PASSWORD
      value: "pass@word1"
  bindings:
  - connectionString: "Server=%SERVICE__POSTGRES__HOST%;Port=%SERVICE__POSTGRES__PORT%;User Id=postgres;Password=pass@word1"
```
This doesn't require any tricky code in libraries or your app. It's roughly as simple as what this code is doing, and things like a hardcoded username/password move out of your code. It got **more** maintainable because the hardcoded password isn't used across multiple files.

You could even do better:

```yaml
services:
- name: postgres
  image:  postgres
  env:
    - name: POSTGRES_PASSWORD
      value: "pass@word1"
  bindings:
  - connectionString: "Server=%SERVICE__POSTGRES__HOST%;Port=%SERVICE__POSTGRES__PORT%;User Id=postgres;Password=%POSTGRES_PASSWORD%"
```

We could also enlighten this to use secrets.

---

I think you were onto something with the user having to do some work re: understanding connection strings, but I think it's better if it works in this direction.

src/Microsoft.Tye.Hosting/Model/Application.cs
```diff
@@ -97,10 +97,32 @@ void SetBinding(ServiceDescription targetService, ServiceBinding b)
                     envName = $"{serviceName.ToUpper()}_{b.Name.ToUpper()}";
                 }
 
+                // Use the container name as the host name if there's a single replica (current limitation)
+                var host = b.Host ?? (service.Description.RunInfo is DockerRunInfo ? targetService.Name : defaultHost);
+
+                // Review: should we split this codepath based on the KIND of service that 
+                // provides the binding?
```
rynowak, line 104: @davidfowl  Let's discuss this.

I don't really have  strong feeling. What I wonder is if we should **avoid** populating the HOST/PORT env-vars in dev because they won't be there in production. However, the example that I liked about with connection string uses token replacement based on the env-vars, so..... 🤣

samples/voting/worker/Extensions.cs
davidfowl: Yea, I'm a fan of this. It might be better if we have a more native syntax for the various pieces:

- Services `${service:name}`
- Env variables `${env:SERVICE__POSTGRES__HOST}`

etc.

src/Microsoft.Tye.Hosting/Model/Application.cs
davidfowl: I think it's worth a shot, though if we do container and not container, it'll break in the case where you run with docker.

samples/voting/worker/Extensions.cs
rynowak: yeah I agree, that's nice.

jkotalik: Pardon my ignorance, but are we doing anything to consolidate how environment variables value are obtained between windows and linux? Like if I have the string: `Server=%SERVICE__POSTGRES__HOST%`, how would this value be obtained on linux?

davidfowl:
This is looking like a clean coherent story!

JamesNK:
Spicy idea:

Has any thought been given to code generation of known environment variables? Or even just service names. At design/build time we have the yaml file with that info. Some codegen would eliminate magic strings:

```cs
var uri = Configuration.GetServiceUri("backend");

// vs

var uri = Configuration.GetServiceUri(ServiceEnvironments.Backend);
```

davidfowl:
That seems like a fine value add once we have source generators

samples/multi-project/tye.yaml
```diff
@@ -16,4 +16,5 @@ services:
 - name: rabbit
   image: rabbitmq:3-management
   bindings:
-    - port: 5672
+    - protocol: amqp
```
davidfowl, line 19: This isn’t required right?

test/Microsoft.Tye.Extensions.Configuration.Tests/TyeConfigurationExtensionsTest.cs
```diff
@@ -104,21 +104,21 @@ public void GetConnectionString_WithBinding_DoesNotCombineHostAndPort()
...
+                { "service:myservice:protocol", "http" },
+                { "service:myservice:host", "expected.example.com" },
+                { "service:myservice:port", "80" },
             });
 
             var configuration = builder.Build();
 
             var result = configuration.GetServiceUri("myservice");
-            Assert.Equal(new Uri("http://expected.example.com"), result);
+            Assert.Equal(new Uri("http://expected.example.com:80"), result);
```
davidfowl, line 121: Why append 80? Should we have smarts around 443 and 80?

samples/mongo-sample/tye.yaml
```diff
@@ -1,3 +1,4 @@
+name: mongo-sample
```
davidfowl, line 1: @jkotalik should this be required?

src/Microsoft.Tye.Hosting/PortAssigner.cs
```diff
@@ -44,12 +44,6 @@ static int GetNextPort()
 
                 foreach (var binding in service.Description.Bindings)
                 {
-                    if (binding.ConnectionString != null)
```
davidfowl, line 47: Why this change?

src/Microsoft.Tye.Hosting/Model/Application.cs
```diff
@@ -80,78 +83,80 @@ string GetValueFromBinding(EnvironmentVariableSource source)
...
-                set($"SERVICE__{configName}__HOST", host);
-                set($"{envName}_SERVICE_HOST", host);
+                set($"SERVICE__{configName}__HOST", binding.Host);
+                set($"{envName}_SERVICE_HOST", binding.Host);
             }
+        }
+
+        // Compute the list of bindings visible to `service`. This is contextual because details like
+        // whether `service` is a container or not will change the result.
+        private List<EffectiveBinding> ComputeBindings(Service service, string defaultHost)
```
davidfowl, line 131: not a fan of local functions yet?

davidfowl:
@rynowak are you planning to update the dapr replacement support that was added with the new syntax?

docs/frontend_backend_run.md
```diff
@@ -134,7 +123,13 @@ Now that we have two applications running, let's make them communicate. By defau
     }
    ```
 
-4. Now register this client in `Startup.cs` class in `ConfigureServices` of the `frontend` project:
+4. Add a reference to the `Microsoft.Tye.Extensions.Configuration` package.
+
+    ```txt
+    dotnet add package Microsoft.Tye.Extensions.Configuration
```
davidfowl, line 129: Do we need to add the feed until we ship to nuget?

src/Microsoft.Tye.Hosting/PortAssigner.cs
davidfowl: Ah, found it.

src/Microsoft.Tye.Hosting/TokenReplacement.cs
```diff
@@ -0,0 +1,111 @@
...
+                binding = bindings.FirstOrDefault(b => b.Service == keys[1] && b.Name == null);
+                return GetValueFromBinding(binding, keys[2]);
+            }
+            else if (keys.Length == 4 && keys[0] == "service")
+            {
+                binding = bindings.FirstOrDefault(b => b.Service == keys[1] && b.Name == keys[2]);
+                return GetValueFromBinding(binding, keys[3]);
+            }
+
+            return null;
```
davidfowl, line 91: Throw? Unsupported? We may actually want to validate this syntax during parsing of config so we can give nice line errors (not required for this PR).

rynowak:
> That seems like a fine value add once we have source generators

Yes, this is something we've talked about 👍

samples/mongo-sample/tye.yaml
rynowak: Right now the schema says its required, and deployment relies on it. If we made it optional, what would the defualt be? the directory name converted to lower?

samples/multi-project/tye.yaml
rynowak: It is if you want to use `GetServiceUri` for rabbit. Per our discussion:

- We treat everything as URI-like by default
- We use connection strings only when you provided one

The URL that would be returned by `GetServiceUri` would then be `http://localhost:5672` which is not what they accept.


So we could either:

- Show this with a connection string in our examples
- Specify the protocol
- Introduce other methods `GetServiceHost` and `GetServicePort`. 

Obviously I prefer 2, or I would have done one of those other things. Your thoughts?

test/Microsoft.Tye.Extensions.Configuration.Tests/TyeConfigurationExtensionsTest.cs
rynowak: we could, none of the code we have today does this.

src/Microsoft.Tye.Hosting/PortAssigner.cs
rynowak: yeah, we need to assign ports for bindings even when they have a connection string.

src/Microsoft.Tye.Hosting/Model/Application.cs
rynowak: This could be a local function but it would close over `this` and so couldn't be static. I wasn't sure how the compiler represents that I just used private.

rynowak:
> are you planning to update the dapr replacement support that was added with the new syntax?

I wasn't planning to make it part of this change, because it's not user-facing, and because I'm running out of time.

src/Microsoft.Tye.Hosting/TokenReplacement.cs
rynowak: This throws in the caller, which has more context.

docs/frontend_backend_run.md
rynowak: Yeah I need to do that in general. Now that we've got a library we need to include that in the instructions everywhere. Good catch.

davidfowl:
Np, make sure you file an issue to follow up.

test/Microsoft.Tye.Extensions.Configuration.Tests/TyeConfigurationExtensionsTest.cs
rynowak: This doesn't actually matter. URI will deal with this for us. If you look at teh codez, you can see that it's wrapped in `new Uri`. I'll change the test.

src/Microsoft.Tye.Extensions.Configuration/TyeSecretsConfigurationBuilderExtensions.cs
```diff
@@ -17,34 +17,26 @@ public static class TyeSecretsConfigurationBuilderExtensions
...
         /// When the environment variable is specified, and the specified directory exists, then the value of
         /// <see cref="TyeSecretsConfigurationSource.FileProvider" /> will be non-null.
         /// </remarks>
-        public static IConfigurationBuilder AddTyeSecrets(this IConfigurationBuilder builder, Action<TyeSecretsConfigurationSource>? configure = null)
+        public static IConfigurationBuilder AddTyeSecrets(this IConfigurationBuilder builder)
         {
-            TyeSecretsConfigurationSource source;
             var secretsDirectory = Environment.GetEnvironmentVariable(TyeSecretsConfigurationSource.TyeSecretsPathEnvironmentVariable);
-            if (secretsDirectory == null || !Directory.Exists(secretsDirectory))
+            if (Directory.Exists(secretsDirectory))
```
davidfowl, line 29: ?

rynowak: This was straight up broken. The config provider was totally busto, because the KPF-file provider doesn't search recursively.

What's here now matches the copy-paste code that we deleted a week agao.

rynowak:
@davidfowl @jkotalik - any more feedback on this one (see the new doc?). I'm trying to wrap up the last few things...

samples/mongo-sample/tye.yaml
davidfowl: I just want to make sure it fails if it's not there.

samples/multi-project/tye.yaml
davidfowl: Option 4, default to TCP and the caller would have to know they have to change it to amqp. Either way it seems like its not required but nice to do it here so that the APIs work.

samples/mongo-sample/tye.yaml
rynowak: Ok so the answer here is that `ApplicationFactory` will default to the directory name. So it's not required. I need to update the docs and schema.

Do you see a drawback to it being optional? I'm cool with it.

samples/multi-project/tye.yaml
rynowak: It's still important to try and agree on what we're showing. If we do what you're suggesting then when we prompt users at the commandline to get a URI during deployment, they will put in *something* 🤷‍♂. I think it's better in that regard to show that you have control over the URI - which is why I ended up here.

samples/mongo-sample/tye.yaml
jkotalik: I'm fine with it being optional and using the directory name as a default.

Issue #14, opened 2020-03-01
Improve UI for dashboard on shutdown


rynowak:
Blazor supports custom UI for when your reconnection is rejected by the server (translation: when the server is gone). We should do something nicer than the default - it's meant to be replaced.

![image](https://user-images.githubusercontent.com/1430011/75616981-6b3bb900-5b0d-11ea-8c2b-43ae493bfcac.png)

Issue #303, opened 2020-04-06
Flaky test: MultiProjectPurgeTest


jkotalik:
https://dnceng.visualstudio.com/internal/_build/results?buildId=586921&view=ms.vss-test-web.build-test-results-tab&runId=18492844&resultId=100014&paneView=debug

```
System.Threading.Tasks.TaskCanceledException : A task was canceled.

at E2ETest.TestHelpers.StartHostAndWaitForReplicasToStart(TyeHost host) in /_/test/E2ETest/TestHelpers.cs:line 162
at E2ETest.TestHelpers.StartHostAndWaitForReplicasToStart(TyeHost host) in /_/test/E2ETest/TestHelpers.cs:line 162
at E2ETest.TyePurgeTests.MultiProjectPurgeTest() in /_/test/E2ETest/TyePurgeTests.cs:line 87
at E2ETest.TyePurgeTests.MultiProjectPurgeTest() in /_/test/E2ETest/TyePurgeTests.cs:line 108
--- End of stack trace from previous location where exception was thrown ---
```

```

Target ResolveProjectReferences:
    __________________________________________________
    Project "/tmp/hh5dcpa4.1sk/worker/worker.csproj" is building "/home/vsts/work/1/s/src/Microsoft.Tye.Extensions.Configuration/Microsoft.Tye.Extensions.Configuration.csproj" (default targets):

    Target GenerateTargetFrameworkMonikerAttribute:
      Skipping target "GenerateTargetFrameworkMonikerAttribute" because all output files are up-to-date with respect to the input files.
    Target _BatchGenerateResxSource:
      Skipping target "_BatchGenerateResxSource" because it has no outputs.
    Target CoreGenerateAssemblyInfo:
      Skipping target "CoreGenerateAssemblyInfo" because all output files are up-to-date with respect to the input files.
    Target CoreCompile:
      Skipping target "CoreCompile" because all output files are up-to-date with respect to the input files.
    Target GenerateBuildDependencyFile:
        /home/vsts/work/1/s/.dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(192,5): error MSB4018: The "GenerateDepsFile" task failed unexpectedly.
[...]

PR #305, opened 2020-04-06
add a custom reconnect dialog

anurse:
Fixes https://github.com/dotnet/tye/issues/14

I wanted to play with Tye and figured this was a quick fix to contribute ;).

I couldn't resist a little nerd jokery too, but I'm happy to edit it to be more "serious" if preferred:

The 'tye' dashboard, with a dark grey overlay and super-imposed text "Tye has exited. It is now safe to close this browser window."

### DIFF
--- a/src/Microsoft.Tye.Hosting/Dashboard/Pages/_Host.cshtml
+++ b/src/Microsoft.Tye.Hosting/Dashboard/Pages/_Host.cshtml
@@ -15,6 +15,33 @@
     <base href="~/" />
     <link rel="stylesheet" href="css/bootstrap/bootstrap.min.css" />
     <link href="css/site.css" rel="stylesheet" />
+    <style>
+        #components-reconnect-modal {
+            font-family: 'Cascadia Code', Consolas, monospace;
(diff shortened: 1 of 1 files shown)


anurse:
I presume the failure is #303 ;)

jkotalik:
I'm not a fan of the color scheme, it looks scary. I don't think we need to adjust the theme right now, @amarie401 thoughts?

I also think the text theme should match whatever the dashboard has.

What is the nerdy joke here?

anurse:
Oof, you hit me right in the olds. (https://www.howtogeek.com/181295/yes-its-okay-to-shut-down-your-computer-with-the-power-button/)

I’m fine switching it to the standard styling, if that’s the preference :).

anurse:
Alright, now that I'm done feeling super old, I swapped back to the previous color scheme. Same text though, so it still has that little Windows 95 inside joke for those who remember ;).

The 'tye' dashboard, with a light grey overlay and super-imposed text "Tye has exited. It is now safe to close this browser window."

Issue #262, opened 2020-03-30
Add scripts for inner loop on tye


rynowak:
Make it convenient to use a local build of tye inside the repo

davidfowl:
The fix I'd like to see here is a bit convoluted but leads to a better dev experience. In development, we should be using the source files directly. To achieve this we need to burn the physical project path into the assembly via some msbuild magic and use that at runtime to add a file provider that points to the project's wwwroot.

PR #306, opened 2020-04-06
Add 'dev-tye' func to `activate` scripts to improve local dev flow

anurse:
So the idea here is that running `. .\activate.ps1` will also add a `dev-tye` function that automatically runs the `src\tye\tye.csproj` from the clone in which you ran `activate`. Because it's a global func, it works wherever you are. Running the `deactivate` function removes the `dev-tye` function.

Might be related to #262 ? @davidfowl had some comments there that seemed to be looking for more complicated stuff, but this seemed like something useful to me and might be nice for other folks in the repo.

I'm happy to just have this in my own scripts if people don't think this fits in the main `activate` script, or should be somewhere else. Just sketched it out while doing some exploration in teh repo and though it could be c o o l.

### DIFF
--- a/activate.ps1
+++ b/activate.ps1
@@ -11,6 +11,11 @@ if ($MyInvocation.CommandOrigin -eq 'runspace') {
     exit 1
 }
 
+function dev-tye() {
+    $ProjectPath = Join-Path (Join-Path $PSScriptRoot "src") "tye"
+    dotnet run --project "$ProjectPath" @args
+}
+
 function deactivate ([switch]$init) {
 
     # reset old environment variables
--- a/activate.ps1
+++ b/activate.ps1
@@ -32,8 +32,9 @@ function deactivate ([switch]$init) {
     Remove-Item env:DOTNET_ROOT -ea ignore
     Remove-Item env:DOTNET_MULTILEVEL_LOOKUP -ea ignore
     if (-not $init) {
-        # Remove the deactivate function
+        # Remove the deactivate and dev-tye functions
         Remove-Item function:deactivate
+        Remove-Item function:dev-tye
     }
 }
 
--- a/activate.ps1
+++ b/activate.ps1
@@ -11,7 +11,7 @@ if ($MyInvocation.CommandOrigin -eq 'runspace') {
     exit 1
 }
 
-function dev-tye() {
+function tye() {
     $ProjectPath = Join-Path (Join-Path $PSScriptRoot "src") "tye"
     dotnet run --project "$ProjectPath" @args
 }
@@ -34,7 +34,7 @@ function deactivate ([switch]$init) {
     if (-not $init) {
         # Remove the deactivate and dev-tye functions
         Remove-Item function:deactivate
-        Remove-Item function:dev-tye
+        Remove-Item function:tye
     }
 }
 
--- a/activate.sh
+++ b/activate.sh
@@ -61,6 +61,7 @@ deactivate () {
     if [ ! "${1:-}" = "init" ] ; then
         # Remove the deactivate function
         unset -f deactivate
+        unset -f dev-tye
     fi
 }
 
--- a/activate.sh
+++ b/activate.sh
@@ -32,6 +32,10 @@ if [ $sourced -eq 0 ]; then
     exit 1
 fi
 
+dev-tye () {
+    dotnet run --project "$THIS_SCRIPT/src/tye" "$@"
+}
+
 deactivate () {
 
     # reset old environment variables
--- a/activate.sh
+++ b/activate.sh
@@ -32,7 +32,7 @@ if [ $sourced -eq 0 ]; then
     exit 1
 fi
 
-dev-tye () {
+tye () {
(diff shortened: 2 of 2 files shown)


jkotalik:
Thanks for doing this. We chatted about this in our sync today and this is exactly what we came up with.

activate.ps1
```diff
@@ -11,6 +11,11 @@ if ($MyInvocation.CommandOrigin -eq 'runspace') {
     exit 1
 }
 
+function dev-tye() {
```
jkotalik, line 14: Thought on just making the name “tye” here instead of dev-tye? Would that work? 

It more similarly matches what activate does with dotnet. We don’t call dev-dotnet.

activate.ps1
```diff
@@ -11,6 +11,11 @@ if ($MyInvocation.CommandOrigin -eq 'runspace') {
...
+    $ProjectPath = Join-Path (Join-Path $PSScriptRoot "src") "tye"
+    dotnet run --project "$ProjectPath" @args
```
jkotalik, line 16: There is a performance cost for doing dotnet run vs running the executable directly if it’s already built and doesn’t need update.

anurse: Correct, but it's also the only way to ensure that if you made a change to tye, you're getting it. If we use the exe directly, it could be out-of-date relative to the code.

activate.ps1
anurse: Personally, I didn't want it to conflict with my globally installed "released" version of tye. Given that I have to have "activated" it though, I could be OK switching to `tye` if people prefer.

anurse:
> Thanks for doing this. We chatted about this in our sync today and this is exactly what we came up with.

Great minds ;).

activate.ps1
jkotalik: Yeah, I'd say conflict with what is installed if it works.

activate.ps1
jkotalik: I don't disagree, but it's going to feel significantly slower, especially if we want to use a dev version of tye for demoing. 

For example, try running tye --version with dotnet run vs running the exe. It's going to feel much better with the exe, even though it doesn't build

Maybe we need two options here. One that builds and one that doesn't build.

activate.ps1
anurse: So you'd prefer that it *does* conflict? I can live with that I suppose.

activate.ps1
anurse: It's "a *lot* slower" when it needs to rebuild, and just "slower" when there hasn't been a code change. To do something like this, you have to compare two time losses:

1. Time spent determining if a rebuild is needed (which is what `dotnet run` does)
2. The expected time spent by engineers using the script and not realizing it's using an out-of-date build.

In my extremely anecdotal past history, I actually find the latter more costly than the former. Even if it doesn't actually waste more time, it builds up my frustration level by making me say "**why isn't my code working!?!?!**" more often :).

So from my perspective, I'm OK losing a bit of time with each command. It's also a relatively constant time increase. So if you're doing a `dev-tye run`, for example, the percent of time lost to `dotnet run` is much smaller than when you're doing `dev-tye --version`, because the `tye run` command takes longer to run anyway and the build is a relatively-constant time operation.

> Maybe we need two options here. One that builds and one that doesn't build.

I'm fine with that, but I'd always use the one that builds 😄. Either way, it's y'alls repo and I'm quite happy to defer to you for preferred behavior (I can always have my own script 😉)

jkotalik: @rynowak @davidfowl thoughts?

davidfowl:
I don’t think it should build. It should just use the already built version

anurse:
🤷‍♀, so be it. I just won't use it :). 

That does raise a concern about putting it in the `activate` script though since now it will also override the existing `tye` command and would mess with other command resolution. The `activate` script is necessary in order to use `dotnet build` or launch VS, so you can't really avoid using it if you're command-line based, and now it would replace your `tye` command with one that may not reflect the current state of the code and hides your currently-installed `tye`.

davidfowl:
I won’t use it either if it builds because that’s way to slow to do any dev work or demos 😬

davidfowl:
@rynowak and @jkotalik if you think it's useful I'm fine merging if we rename tye to something else so it won't conflict with my existing dev flow and won't overwrite the global tool.

activate.ps1
rynowak: I would prefer this do `dotnet run`. It's trivial to install a private build if you need *teh fast*.

rynowak: I can tell you that if there's a flavor of this script that doesn't do a build, I will never never never never never use it. All of the productivity of running 1 second faster is erased the first time you have to stop and wonder if it's actually using the right build 😆

rynowak:
I was commenting on this and saw that everyone else disagrees with me. If the script doesn't build I won't use it. I don't care about being 2 sec faster, I want to know that it is always running the latest code. 

I already have a workflow for this, so I'm not impacted either way. I thought the point of this was to make it easier to *other people* work on the repo, and it's not clear how that should work.

PR #307, opened 2020-04-07
fix launch path

staff0rd:


davidfowl:
This is how you can tell nobody is using vscode with f5

rynowak:
yeah, I just have scripts that do this and I attach.

```sh
dotnet run -p ~/github.com/dotnet/tye/src/tye -- "[debug]" "$@"
```

PR #311, opened 2020-04-07
Rename tye to dtye

davidfowl:


activate.ps1
```diff
@@ -11,7 +11,7 @@ if ($MyInvocation.CommandOrigin -eq 'runspace') {
     exit 1
 }
 
-function tye() {
+function dtye() {
```
jkotalik, line 14: dtye or dev-tye?

rynowak: `dtye` is better because then I don't have to rename my script.

Issue #206, opened 2020-03-26
`tye run` attempts to run test projects


stevejgordon:
`tye run` on a solution with an API, web frontend and a test project, attempts to run the test project. Would it be possible to detect test projects and exclude them?

I worked around this by using `tye init` to create the yaml file and then removed the test project.

davidfowl:
I wonder if we can detect a test project...

jkotalik:
Yeah I started brainstorming about this, and I actually got quite nervous about detecting test projects. I think if we want `tye run` to work without a tye.yaml file, we need to figure out a way to exclude test projects, but adding heuristics for checking if a project is a test project (ex folder path includes "test", project name includes "test", etc.), it can lead to false positives. 

TL;DR I think we _need_ to fix this, but not sure how to sniff for test projects, as David suggests.

rynowak:
We can check for the presence of the test SDK, I think. I'm sure there's an MSBuild way to do this, and *I have the power of MSBuild and anime on my side*.

rynowak:
Ok so the right way to detect this is with `IsTestProject` = true. That's set by the test SDK.

The thing we might want to *instead* is look for projects that produce an EXE. Since we're no longer looking for launch settings we will now also trigger on class libraries.

davidfowl:
> Since we're no longer looking for launch settings we will now also trigger on class libraries.

Eeek

ArieJones:
While I see the IsTestProject in the Tye E2ETest project I don't see it in projects that we have recently upgraded to 3.1 ... for example 

```<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
    <IsPackable>false</IsPackable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="AutoMapper" Version="9.0.0" />
    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="7.0.0" />
    <PackageReference Include="FakeItEasy" Version="6.0.0-beta.1" />
    <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="3.1.0" />
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0-preview-20191115-01" />
    <PackageReference Include="Shouldly" Version="4.0.0-beta0002" />
[...]

PR #314, opened 2020-04-07
Exclude projects without launchSettings.json

rynowak:
Fixes: #206

This change only applies to the logic that we use to look at solution
files. This adds a heuristic that we only include projects with a
launchSettings.json for these operations (when considering a solution).
This will exclude unit test projects and class libraries.

In this case where the user has a `tye.yaml` then whatever's in there
will win. Likewise if a user passes a path to a project directly then
we'll try to run it without asking questions.

I chose `launchSettings.json` for this because it's a decision we can
make really quickly without having to run any MSBuild code. When we
think about operating on a solution with 50 projects with `tye run` we
really really don't want to have to evaluate each project with MSBuild
to figure out what kind of project it is (500ms).

### DIFF
--- a/test/E2ETest/TyeInitTests.cs
+++ b/test/E2ETest/TyeInitTests.cs
@@ -74,5 +74,24 @@ namespace E2ETest
 
             Assert.Equal(expectedContent.NormalizeNewLines(), content.NormalizeNewLines());
         }
+
+        // Tests our logic that excludes non-applications (unit tests, classlibs, etc)
+        [Fact]
+        public void Init_ProjectKinds()
+        {
+            using var projectDirectory = CopyTestProjectDirectory("project-types");
+
+            // delete already present yaml
+            File.Delete(Path.Combine(projectDirectory.DirectoryPath, "tye.yaml"));
+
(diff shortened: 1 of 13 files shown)


test/E2ETest/TyeInitTests.cs
```diff
@@ -74,5 +74,24 @@ public void FrontendBackendTest()
 
             Assert.Equal(expectedContent.NormalizeNewLines(), content.NormalizeNewLines());
         }
+
+        // Tests our logic that excludes non-applications (unit tests, classlibs, etc)
+        [Fact]
+        public void Init_ProjectKinds()
+        {
+            using var projectDirectory = CopyTestProjectDirectory("project-types");
```
jkotalik, line 82: Do you think having a tye run test is necessary here to verify that the application doesn't run any test projects? I don't think it does, but just wanted to pick your brain.

rynowak: I like where your head is at, but I don't really think so. If this were a shipping product maybe 👍 I'm optimizing for current/future productivity and the risk of a regression is low.

PR #315, opened 2020-04-08
Use environment variables for secrets in prod

### LINKED ISSUE #313: Figure out how to get rid of `AddTyeSecrets`
We should make this go away as a required piece of code, but keep the functionality somehow.

Ideas:

- [ ] Don't use files on disk (has drawbacks: less secure and can't update in place)
- [ ] Build this into ASP.NET Core somehow
- [ ] Replace CreateDefaultBuilder with our own thing
- [ ] Inject a hosting startup in production
- [ ] Inject a hosting startup and dependencies during publishing

davidfowl:
- Updating in place does not apply to connection strings specifically since they are usually configured at startup as singletons.
- This brings consistency with the development experience.
- Gets rid of AddTyeBindings
- "Less secure", sure but to be pedantic env variables are stored in virtual file on disk while the process is running. I'd argue if you want full security then use keyvault/vault/name your secret store.

Fixes #313


src/Microsoft.Tye.Core/ComputedBindings.cs
```diff
@@ -43,33 +43,33 @@ protected SecretInputBinding(string name, ServiceBuilder service, BindingBuilder
         public BindingBuilder Binding { get; }
     }
 
-    public sealed class SecretConnctionStringInputBinding : SecretInputBinding
+    public sealed class SecretConnectionStringInputBinding : SecretInputBinding
```
rynowak, line 46: <3

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
```diff
@@ -211,33 +211,84 @@ internal static class KubernetesManifestGenerator
...
+                        foreach (var binding in bindings.Bindings.OfType<SecretInputBinding>())
                         {
-                            env.Add(new YamlMappingNode()
+                            /*
+                             - name: SECRET_USERNAME
+                               valueFrom:
+                                 secretKeyRef:
+                                   name: mysecret
+                                   key: username
+                           */
```
rynowak, line 222: eww no block comments

docs/service_discovery.md
```diff
@@ -286,22 +286,4 @@ stringData:
...
-        })
-        .ConfigureWebHostDefaults(web =>
-        {
-            web.UseStartup<Startup>();
-        });
-```
-
-Tye will also inject the `TYE_SECRETS_PATH` environment variable with the value `/var/tye/bindings/`. This is done to avoid hardcoding a file-system path in the configuration provider.
\ No newline at end of file
+To get these values into the application, Tye will use environment variables that reference the Kubernetes secrets described above and will use the environment-variable naming scheme described above.
```
rynowak, line 289: The redis deployment part of the tutorial needs to be updated.

rynowak: Do a `CTRL+F` to make sure we don't have any lingering references since we're removing it.

davidfowl: done

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
davidfowl: Fixed

davidfowl:
@rynowak good to merge?

Issue #316, opened 2020-04-08
Don't require a name in tye.yaml and default to the current directory


jkotalik:
We currently validate that a name is present in the tye.yaml, however a good default would be the directory name of the current directory (ToLower).

rynowak:
I'm pretty sure we actually **don't** validate this. I think the work that's needed will be to update the schema.

PR #321, opened 2020-04-08
Update schema and docs to describe name as optional

rynowak:
Fixes: #316


jkotalik:
Ah, I got confused between service names and the top level name for where we have validation.

PR #322, opened 2020-04-08
More yaml verification for a few scenarios

jkotalik:
Fixes https://github.com/dotnet/tye/issues/232

### DIFF
--- a/src/Microsoft.Tye.Core/ConfigModel/ConfigApplication.cs
+++ b/src/Microsoft.Tye.Core/ConfigModel/ConfigApplication.cs
@@ -56,6 +56,7 @@ namespace Microsoft.Tye.ConfigModel
                 }
             }
 
+
             foreach (var service in config.Services)
             {
                 context = new ValidationContext(service);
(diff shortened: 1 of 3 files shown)


src/Microsoft.Tye.Core/ConfigModel/ConfigApplication.cs
```diff
@@ -83,11 +98,17 @@ public void Validate()
 
                     if (string.IsNullOrEmpty(binding.Name) && service.Bindings.Count > 1)
                     {
-                        throw new TyeYamlException(CoreStrings.MultipleServiceBindingsWithoutName);
+                        throw new TyeYamlException(CoreStrings.FormatMultipleBindingWithoutName("service"));
                     }
+
                     if (service.Bindings.Where(o => o.Name == binding.Name).Count() > 1)
```
davidfowl, line 104: nit:

```C#
service.Bindings.Where(o => o.Name == binding.Name).Count() > 1
```

Can be written as:

```C#
service.Bindings.Count(o => o.Name == binding.Name) > 1
```

jkotalik: oh sick linq is awesome.

Issue #285, opened 2020-04-01
Update Dapr sample to latest release


rynowak:
Dapr released an update that resolves one of the problems users are likely to hit with our sample. https://github.com/dapr/dotnet-sdk/releases/tag/v0.6.0-preview01

PR #326, opened 2020-04-08
Update dapr version

jkotalik:
Fixes https://github.com/dotnet/tye/issues/285.

Having trouble running this locally at the moment due to the error:

```
"failed to start metrics server: listen tcp :49398: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted.
```

### DIFF
--- a/src/Microsoft.Tye.Hosting/Model/Application.cs
+++ b/src/Microsoft.Tye.Hosting/Model/Application.cs
@@ -38,6 +38,10 @@ namespace Microsoft.Tye.Hosting.Model
                 // Inject normal configuration
                 foreach (var pair in service.Description.Configuration)
                 {
+                    if (pair.Name == "METRICS_PORT")
+                    {
+                        Console.WriteLine("wow");
+                    }
                     if (pair.Value is object)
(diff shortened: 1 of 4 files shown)


jkotalik:
Figured it out, we had a bad check for the service name. This is working now.

src/Microsoft.Tye.Hosting/Model/Application.cs
```diff
@@ -38,6 +38,10 @@ public void PopulateEnvironment(Service service, Action<string, string> set, str
                 // Inject normal configuration
                 foreach (var pair in service.Description.Configuration)
                 {
+                    if (pair.Name == "METRICS_PORT")
+                    {
+                        Console.WriteLine("wow");
+                    }
```
rynowak, line 44: uhhhhh

rynowak: wow such metrics.

jkotalik: How did this get here????? 😉

Issue #295, opened 2020-04-03
Print output for msbuild pre-amble


davidfowl:
### What should we add or change to make your life better?
We need to print out the something while we're building the tye.yaml model internally since it involves msbuild and could be slow. Today it looks as though the tool itself is just really slow to boot on commands.

### Why is this important to you?
It'll be a better user experience.

rynowak:
We shouldn't print MSBuild's header because that will be one-per-project. We should just print a progress message like

```
Loading Application Details...
```

davidfowl:
Yea the title is wrong but the issue text is correct

PR #328, opened 2020-04-09
Show some progress while we're loading tye.yaml

rynowak:
Fixes: #295

### DIFF
--- a/src/tye/Program.RunCommand.cs
+++ b/src/tye/Program.RunCommand.cs
@@ -72,18 +72,21 @@ namespace Microsoft.Tye
                     throw new CommandException("No project or solution file was found.");
                 }
 
-                var output = new OutputContext(console, Verbosity.Quiet);
+                var output = new OutputContext(console, Verbosity.Info);
+
+                output.WriteInfoLine("Loading Application Details...");
                 var application = await ApplicationFactory.CreateAsync(output, path);
-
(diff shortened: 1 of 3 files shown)


src/tye/Program.RunCommand.cs
```diff
@@ -72,18 +72,21 @@ private static Command CreateRunCommand(string[] args)
...
+                var output = new OutputContext(console, Verbosity.Info);
 
+                output.WriteInfoLine("Loading Application Details...");
+                var application = await ApplicationFactory.CreateAsync(output, path);
                 if (application.Services.Count == 0)
                 {
                     throw new CommandException($"No services found in \"{application.Source.Name}\"");
                 }
 
+                await application.ProcessExtensionsAsync(ExtensionContext.OperationKind.LocalRun);
```
rynowak, line 84: Moved the validation to be right after the application is read.

src/tye/Program.RunCommand.cs
```diff
@@ -72,18 +72,21 @@ private static Command CreateRunCommand(string[] args)
...
+
+                InitializeThreadPoolSettings(application.Services.Count);
+
+                output.WriteInfoLine("Launching Tye Host...");
+                output.WriteInfoLine(string.Empty);
```
rynowak, line 89: Blank line here is because we're changing from one logging format to another. Makes it look better, and makes it clear that the host is taking over.

PR #329, opened 2020-04-09
Some minor fixes

rynowak:


restore.cmd
```diff
@@ -1,3 +1,3 @@
 @ECHO OFF
 SETLOCAL
-PowerShell -NoProfile -NoLogo -ExecutionPolicy ByPass -Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = '';& '%~dp0build.ps1' -all -nobuild -restore %*; exit $LASTEXITCODE"
\ No newline at end of file
+PowerShell -NoProfile -NoLogo -ExecutionPolicy ByPass -Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = '';& '%~dp0eng\common\Build.ps1' -all -nobuild -restore %*; exit $LASTEXITCODE"
```
rynowak, line 3: This was an incorect copy-paste from dotnet/aspnetcore. We don't have a build.ps1 script in the root.

src/Microsoft.Tye.Core/PublishProjectStep.cs
```diff
@@ -51,7 +51,7 @@ public override async Task ExecuteAsync(OutputContext output, ApplicationBuilder
                 throw new CommandException("'dotnet publish' failed.");
             }
 
-            output.WriteDebugLine($"Created Publish Output: '{outputDirectory}'");
+            output.WriteDebugLine($"Created Publish Output: '{outputDirectory.DirectoryPath}'");
```
rynowak, line 54: Found while trying to repro another issue. This was just printing the class name, `TempDirectory`.

PR #330, opened 2020-04-09
WIP local development docs

rynowak:


rynowak:
Sharing my progress so you can comment on it :)

rynowak:
There's a rough TOC at the bottom of the additional points to discuss. Please point it out if I missed anything.

davidfowl:
This is a really great start and I think we should merge it and tbd the rest. The only feedback I have is that I wonder if docs like this get too long. Maybe we need to break it up but I don’t have suggestions on how yet

docs/reference/local_development.md
```diff
@@ -0,0 +1,198 @@
...
+
+- Load the application (`tye.yaml`)
+- Build all projects 
+- Pull all container images
+- Assign ports and bindings
+- Launch and monitor all services
+
+## Loading the application
+
+A Tye application is specified using `tye.yaml`. You can find more information about the Tye application format [here](/docs/reference/schema.md).
```
jkotalik, line 21: Let's figure out a way to share this info with tye deploy. I would end up duplicating most of this.

docs/reference/local_development.md
```diff
@@ -0,0 +1,198 @@
...
+Tye acts as a *local orchestrator* for development by building and launching multiple services, managing and monitoring services, and enabling network communication between services.
+
+This document will describe how these features work, as well as the implications and behaviors of various optional settings.
+
+## Executing `tye run`
+
+When executing `tye run` for local development, Tye goes through the following steps:
+
+- Load the application (`tye.yaml`)
+- Build all projects
```
jkotalik, line 14: Should we mention extensions here?

rynowak: I wasn't planning to yet. That seems like a level of detail we don't need.

PR #337, opened 2020-04-09
update versions in getting started

rynowak:


rynowak:
Will repoint this at nuget.org one the packages make it there.

rynowak:
docs only.

PR #338, opened 2020-04-09
connectionstring lowercase for kubernetes

jkotalik:
Ran into during verification. connectionString vs connectionstring in kubernetes.

If we prefer connectionString, I can change the other place instead.

### DIFF
--- a/src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
+++ b/src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
@@ -229,7 +229,7 @@ namespace Microsoft.Tye
                                             { "secretKeyRef", new YamlMappingNode()
                                                 {
                                                     { "name", new YamlScalarNode(binding.Name) { Style = ScalarStyle.SingleQuoted } },
-                                                    { "key", new YamlScalarNode("connectionString") { Style = ScalarStyle.SingleQuoted } },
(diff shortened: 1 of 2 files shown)


jkotalik:
I've verified this fix works, need confirmation on what syntax you prefer for connectionString and connectionstring and will update tests.

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
```diff
@@ -229,7 +229,7 @@ internal static class KubernetesManifestGenerator
                                             { "secretKeyRef", new YamlMappingNode()
                                                 {
                                                     { "name", new YamlScalarNode(binding.Name) { Style = ScalarStyle.SingleQuoted } },
-                                                    { "key", new YamlScalarNode("connectionString") { Style = ScalarStyle.SingleQuoted } },
+                                                    { "key", new YamlScalarNode("connectionstring") { Style = ScalarStyle.SingleQuoted } },
```
rynowak, line 232: cross-checked this with: https://github.com/dotnet/tye/blob/master/src/Microsoft.Tye.Core/ValidateSecretStep.cs#L113

rynowak: Looks good 👍

PR #342, opened 2020-04-10
Update branding to 0.2

rynowak:


davidfowl:
Wait, you need to update the getting started docs for using ci builds

rynowak:
Good catch, will do that!

PR #347, opened 2020-04-10
Use take instead of two commands

khalidabuhakmeh:
the `take` command will call `mkdir` and `cd` in one command.

### DIFF
--- a/docs/tutorials/hello-tye/00_run_locally.md
+++ b/docs/tutorials/hello-tye/00_run_locally.md
@@ -7,8 +7,7 @@ This tutorial will demonstrate how to use [`tye run`](/docs/reference/commandlin
 1. Make a new folder called `microservice` and navigate to it:
 
     ```text
-    mkdir microservice
-    cd microservice
+    take microservice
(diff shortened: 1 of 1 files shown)


docs/tutorials/hello-tye/00_run_locally.md
```diff
@@ -7,8 +7,7 @@ This tutorial will demonstrate how to use [`tye run`](/docs/reference/commandlin
 1. Make a new folder called `microservice` and navigate to it:
 
     ```text
-    mkdir microservice
-    cd microservice
+    take microservice
```
jkotalik, line 10: I haven't heard of this command before, and I don't see it available on both ubuntu or powershell. 

The commands above are generally available on most shells, so I'd prefer to keep it as is.

davidfowl: Agree

jkotalik:
Thanks for the suggestion, but we are going to close this for now as we believe `take` isn't generally available.

PR #349, opened 2020-04-10
Correct FAQ link

mdekrey:
While Windows is case-insensitive, it appears GitHub nees the appropriate casing for the link to work.


davidfowl:
Thanks!

PR #350, opened 2020-04-10
add dashboard option to auto launch

spboyer:
added `--dashboard` to launch the dashboard on `run`

Test is also added but should likely comment this out due to the need for selenium at this point.

### DIFF
--- a/src/Microsoft.Tye.Hosting/TyeHost.cs
+++ b/src/Microsoft.Tye.Hosting/TyeHost.cs
@@ -4,10 +4,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.CommandLine.Invocation;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using System.Net;
 using System.Net.Sockets;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.Extensions.Configuration;
@@ -97,6 +100,11 @@ namespace Microsoft.Tye.Hosting
             _logger.LogInformation("Dashboard running on {Address}", app.Addresses.First());
 
             await _processor.StartAsync(_application);
+
+            if (_args.Contains("--dashboard"))
+            {
+                OpenDashboard(app.Addresses.First());
+            }
 
             return app;
         }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -298,6 +313,32 @@ private async Task StopAsync()
...
+                    System.Diagnostics.Process.Start("xdg-open", url);
+                }
+                else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                {
+                    System.Diagnostics.Process.Start("open", url);
+                }
+            }
+            catch
+            {
+                throw;
```
davidfowl, line 338: Why throw here instead of logging directly?

spboyer: method was static, _logger not available will make the change

test/E2ETest/TyeRunTests.cs
```diff
@@ -71,6 +71,33 @@ public async Task SingleProjectRunTest()
             });
         }
 
+        [Fact]
+        public async Task SingleProjectRunTestWithDashboard()
```
jkotalik, line 75: Agreed with your comment, right now this test isn't useful. Let's just delete it for now and verify manually that this works on linux, mac, and windows.

src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -298,6 +306,31 @@ private async Task StopAsync()
...
+            {
+                //still not addressed https://github.com/dotnet/corefx/issues/10361
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    url = url.Replace("&", "^&");
+                    System.Diagnostics.Process.Start(new ProcessStartInfo("cmd", $"/c start {url}") { CreateNoWindow = true });
+                }
+                else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                {
+                    System.Diagnostics.Process.Start("xdg-open", url);
```
rynowak, line 321: TIL about `xdg-open`. Nice!

src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -298,6 +306,31 @@ private async Task StopAsync()
...
 
+        private void OpenDashboard(string url)
+        {
+            try
+            {
+                //still not addressed https://github.com/dotnet/corefx/issues/10361
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    url = url.Replace("&", "^&");
+                    System.Diagnostics.Process.Start(new ProcessStartInfo("cmd", $"/c start {url}") { CreateNoWindow = true });
```
rynowak, line 317: Do we need quotes around these URLs?

src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -298,6 +306,31 @@ private async Task StopAsync()
             _processor = null;
         }
 
+        private void OpenDashboard(string url)
+        {
+            try
+            {
+                //still not addressed https://github.com/dotnet/corefx/issues/10361
```
rynowak, line 313: It sounds like from here: https://github.com/dotnet/runtime/issues/17938#issuecomment-380770589 that this does work now if you use `UseShellExecute = true`. I don't really mind the code here because I think it's super clear what's happening longform.

Either way could you change this comment to be less negative?

spboyer: I have had random NoOp using ShellExec on other command line apps  [dotnet azure deploy]. This works 100%. Removing comment.

src/Microsoft.Tye.Hosting/TyeHost.cs
spboyer: yes, powershell works fine. However if zsh or other shell add-ins. Quotes ensure success.

davidfowl:
Thanks!

Issue #15, opened 2020-03-01
Support JSON in addition to YAML


rynowak:
Some people have hate in their hearts for YAML. It's really cheap for us to support JSON as well as YAML, since YAML can be converted to YAML. https://github.com/aaubry/YamlDotNet/wiki/Samples.ConvertYamlToJson

Additionally, JSON has better support for schema/intellisense

We could do this by making JSON the source of truth and then converting YAML to JSON (always)

PureKrome:
> since YAML can be converted to YAML.

I think you ment "`since YAML can be converted to JSON`". 

And yes, would love json as the source of truth. But it's now 6 weeks after this issue, so i'm guessing this boat has sailed ... ⛵

jkotalik:
No boats have sailed here! This project is just starting, and the last 6 weeks we honestly have been mostly working on documentation and paper cuts that people are running into, no major feature work besides core fundamentals (ex service discovery).

PureKrome:
woot! and cheers @spboyer for the PR. Respect, mate 👍 

Respect to the entire team, actually 👍

spboyer:
@rynowak @jkotalik - can we close? Did a spike on this with #355 and we didn't merge as YAML was the better choice moving forward.

rynowak:
I think @davidfowl will be really happy to see this closed.

davidfowl:
😄

PureKrome:
Ok - so summary/answer is: YAML config only. No JSON config support.

yes?

PR #355, opened 2020-04-10
Support json as config as well as yaml

spboyer:
Addresses #15

### DIFF
--- a/src/Microsoft.Tye.Core/ConfigModel/ConfigFactory.cs
+++ b/src/Microsoft.Tye.Core/ConfigModel/ConfigFactory.cs
@@ -3,8 +3,13 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections.Generic;
+using System.Dynamic;
 using System.IO;
+using k8s;
+using k8s.Models;
 using Microsoft.Tye.Serialization;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
 using Tye.Serialization;
 
 namespace Microsoft.Tye.ConfigModel
@@ -16,6 +21,9 @@ namespace Microsoft.Tye.ConfigModel
             var extension = file.Extension.ToLowerInvariant();
             switch (extension)
             {
+                case ".json":
+                    return FromJson(file);
+
                 case ".yaml":
                 case ".yml":
                     return FromYaml(file);
@@ -86,6 +94,17 @@ namespace Microsoft.Tye.ConfigModel
         private static ConfigApplication FromYaml(FileInfo file)
         {
             using var parser = new YamlParser(file);
+            return parser.ParseConfigApplication();
+        }
+
+        private static ConfigApplication FromJson(FileInfo file)
+        {
+            var converter = new ExpandoObjectConverter();
+            dynamic obj = JsonConvert.DeserializeObject<ExpandoObject>(file.OpenText().ReadToEnd(), converter);
+            var serializer = new YamlDotNet.Serialization.Serializer();
+            string yamlContent = serializer.Serialize(obj);
+
+            using var parser = new YamlParser(yamlContent);
             return parser.ParseConfigApplication();
         }
     }
--- a/src/Microsoft.Tye.Core/Serialization/YamlSerializer.cs
+++ b/src/Microsoft.Tye.Core/Serialization/YamlSerializer.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using k8s;
+using System.Linq;
 using YamlDotNet.Serialization;
 using YamlDotNet.Serialization.NamingConventions;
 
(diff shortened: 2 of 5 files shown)


tye.sln
```diff
@@ -4,6 +4,9 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 VisualStudioVersion = 16.0.29613.14
 MinimumVisualStudioVersion = 15.0.26124.0
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{8C662D59-A3CB-466F-8E85-A8E6BA5E7601}"
+	ProjectSection(SolutionItems) = preProject
+		src\tye\InitHost.cs = src\tye\InitHost.cs
```
davidfowl, line 8: What is this?

spboyer: welcome to the Internal preview?

src/Microsoft.Tye.Core/ConfigModel/ConfigFactory.cs
```diff
@@ -88,5 +96,16 @@ private static ConfigApplication FromYaml(FileInfo file)
...
+
+        private static ConfigApplication FromJson(FileInfo file)
+        {
+            var converter = new ExpandoObjectConverter();
+            dynamic obj = JsonConvert.DeserializeObject<ExpandoObject>(file.OpenText().ReadToEnd(), converter);
+            var serializer = new YamlDotNet.Serialization.Serializer();
+            string yamlContent = serializer.Serialize(obj);
+
+            using var parser = new YamlParser(yamlContent);
+            return parser.ParseConfigApplication();
```
rynowak, line 108: Is this step even needed? (would YamlParser just parse the JSON document since JSON is YAML).

I'm thinking about things like line-numbers matching up if we do a conversion step. We just did a bunch of work in the yaml parser/validator to get better error messages with line numbers.

src/Microsoft.Tye.Core/Serialization/YamlSerializer.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using k8s;
```
rynowak, line 5: Does this class actually use the k8s OM for anything?

src/tye/InitHost.cs
```diff
@@ -6,21 +6,32 @@ namespace Microsoft.Tye
...
-        public static (string, string) CreateTyeFileContent(FileInfo? path, bool force)
+        public static (string, string) CreateTyeFileContent(FileInfo? path, bool force, bool json = false)
         {
             if (path is FileInfo && path.Exists && !force)
             {
                 ThrowIfTyeFilePresent(path, "tye.yml");
                 ThrowIfTyeFilePresent(path, "tye.yaml");
+
+                if (json)
+                    ThrowIfTyeFilePresent(path, "tye.json");
```
rynowak, line 34: This check should probably happen regardless of `json` being true/false. We don't want to create a `tye.yaml` if you already have `tye.json` without `force`.

src/tye/InitHost.cs
```diff
@@ -77,7 +88,7 @@ public static (string, string) CreateTyeFileContent(FileInfo? path, bool force)
                     }
 
                     // If the input file is a sln/project then place the config next to it
-                    outputFilePath = Path.Combine(directory.FullName, "tye.yaml");
+                    outputFilePath = Path.Combine(directory.FullName, outputFilePath);
```
rynowak, line 91: What does this do with our comments? https://github.com/dotnet/tye/pull/355/files#diff-03e018baedc83b3a7dd33c431001ccfdR48 when you do JSON?

src/Microsoft.Tye.Core/Serialization/YamlSerializer.cs
```diff
@@ -17,5 +18,19 @@ public static ISerializer CreateSerializer()
...
+            var deserializer = new DeserializerBuilder().Build();
+            var yamlObject = deserializer.Deserialize<string>(yamlContent);
+
+            var serializer = new SerializerBuilder()
+                .JsonCompatible()
+                .Build();
+
+            var json = serializer.Serialize(yamlObject);
+
+            return json;
```
rynowak, line 33: It would be really nice if this could also inject `$schema` pointing at: https://raw.githubusercontent.com/dotnet/tye/master/src/schema/tye-schema.json

src/tye/InitHost.cs
spboyer: Comments are not supported in JSON.  Also consider the following note from http://www.yaml.org/spec/1.2/spec.html section 3.2.3.3. Comments

> Comments are a presentation detail and must not have any effect on the serialization tree or representation graph...

src/Microsoft.Tye.Core/Serialization/YamlSerializer.cs
spboyer: No, removed.

src/Microsoft.Tye.Core/ConfigModel/ConfigFactory.cs
spboyer: Good question, would have to test. First though is no. Yaml is format (space) specific where JSON can be one line no spaces and still be valid.

src/Microsoft.Tye.Core/Serialization/YamlSerializer.cs
spboyer: added.

spboyer:
This was  a good spike. Moving away in favor of current configuration option,
@davidfowl

PR #359, opened 2020-04-13
Infrastructure and test cleanup

jkotalik:
- Moves or copies any samples that are used for tests into testassets (aka stop testing samples).
  - A few apps have been kept in the samples folder (apps-with-ingress, nginx) as they demonstrate interesting scenarios.
  - I've kept the frontend-backend app in samples for now as it's a basic app many of us like to run locally for sanity checking with low friction. 
  - I've moved the single-project and multi-project apps to the test folder, aka they don't exist in samples anymore.
- For tye init tests, stop doing file comparison and compare with yaml. I'm going to make a follow up PR to polish this more with a visitor for comparison, but this was a fairly cheap win.


test/E2ETest/TestHelpers.cs
```diff
@@ -221,5 +197,106 @@ await using (cancellation.Token.Register(() => stoppedTask.TrySetCanceled()))
                 }
             }
         }
+
+        public static void CompareConfigApplications(ConfigApplication expected, ConfigApplication actual)
```
rynowak, line 201: pref: some other class named `TyeAssert` or something like that is better than putting a bunch of random stuff in *helpers*

test/E2ETest/TyeInitTests.cs
```diff
@@ -15,64 +18,32 @@ public class TyeInitTests
...
-            // delete already present yaml
-            File.Delete(Path.Combine(projectDirectory.DirectoryPath, "tye.yaml"));
-
-            var projectFile = new FileInfo(Path.Combine(projectDirectory.DirectoryPath, "frontend-backend.sln"));
-
-            var (content, _) = InitHost.CreateTyeFileContent(projectFile, force: false);
-            var expectedContent = File.ReadAllText("testassets/init/frontend-backend.yaml");
-
-            output.WriteLine(content);
+            var expectedContent = File.ReadAllText("testassets/init/multi-project.yaml");
```
rynowak, line 43: I feel like we should keep **ONE** of these as a text-bases comparison. There is some stuff we care about in there that's non-semantic (comments, empty properties vs the properties not present).

test/E2ETest/TestHelpers.cs
jkotalik: I see the vision, and it's definitely better.

Issue #261, opened 2020-03-30
Test yaml with a semantic diff instead of string comparison


rynowak:

jkotalik:
This is done.

PR #361, opened 2020-04-13
Use node comparison for generate and init comparison

jkotalik:
For https://github.com/dotnet/tye/issues/261.

I actually recognized that YamlNode implements Equals correctly and recurisvely, however it doesn't give great error messages as it only returns true/false. So I still went ahead and made a small visitor pattern for each node type.

Example of output:

image

We also output the the yaml if they don't match in the output.


test/E2ETest/TestHelpers.cs
```diff
@@ -221,5 +222,37 @@ await using (cancellation.Token.Register(() => stoppedTask.TrySetCanceled()))
                 }
             }
         }
+
+        public static void AssertYamlContentEqual(string expected, string actual, ITestOutputHelper output = null!)
```
davidfowl, line 226: Shouldn't this be ITestOutputHelper? output = null

test/E2ETest/Infrastructure/EqualityYamlNodeVisitor.cs
```diff
@@ -0,0 +1,136 @@
+﻿using System;
```
rynowak, line 1: copyright

test/E2ETest/Infrastructure/EqualityYamlNodeVisitor.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                VisitSequence((YamlSequenceNode)node, (YamlSequenceNode)otherNode);
+            }
+        }
+
+        public void VisitSequence(YamlSequenceNode node, YamlSequenceNode otherNode)
+        {
+            if (otherNode == null)
+            {
+                throw new TyeYamlException($"Expected sequence to be not null, difference starting at Expected: ({node.Start.Line}, {node.Start.Column})");
```
rynowak, line 33: Should we track the context (path to current node) for use in error messages? We could include the current context, and since you'd have that you could also track the parent node of `otherNode` which would be able to give you line number info of both expected and actual.

test/E2ETest/Infrastructure/EqualityYamlNodeVisitor.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        public EqualityYamlNodeVisitor()
+        {
+        }
+
+        public void Visit(YamlNode node, YamlNode otherNode)
+        {
+            if (node.NodeType == YamlNodeType.Mapping)
+            {
+                VisitMapping((YamlMappingNode)node, (YamlMappingNode)otherNode);
```
rynowak, line 17: We should have a good error for the case where the cast fails on `otherNode`.

rynowak: Put another way, if `node` and `otherNode` aren't the same type, you'll get a garbo error.

test/E2ETest/Infrastructure/EqualityYamlNodeVisitor.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            if (node.Tag?.Equals(otherNode.Tag) == false)
+            {
+                throw new TyeYamlException($"Expected tags to be equal, difference starting at Expected: ({node.Start.Line}, {node.Start.Column}) Actual: ({otherNode.Start.Line}, {otherNode.Start.Column})");
+            }
+
+            if (node.Children.Count != otherNode.Children.Count)
+            {
+                throw new TyeYamlException($"Number of children differ for sequence, Expected: ({node.Start.Line}, {node.Start.Column}) Actual: ({otherNode.Start.Line}, {otherNode.Start.Column})");
+            }
```
rynowak, line 44: It might be better to put this at the end, and then you could include things like the text of the *extra* children afterwards.

test/E2ETest/Infrastructure/EqualityYamlNodeVisitor.cs
```diff
@@ -0,0 +1,136 @@
...
+                    VisitMapping((YamlMappingNode)childNode, (YamlMappingNode)otherChildNode);
+                }
+                else if (childNode.NodeType == YamlNodeType.Scalar)
+                {
+                    VisitScalar((YamlScalarNode)childNode, (YamlScalarNode)otherChildNode);
+                }
+                else if (childNode.NodeType == YamlNodeType.Sequence)
+                {
+                    VisitSequence((YamlSequenceNode)childNode, (YamlSequenceNode)otherChildNode);
+                }
```
rynowak, line 67: call `Visit(...)` here

test/E2ETest/Infrastructure/EqualityYamlNodeVisitor.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            if (node.Children.Count != otherNode.Children.Count)
+            {
+                throw new TyeYamlException($"Number of children differ for mapping, Expected: ({node.Start.Line}, {node.Start.Column}) Actual: ({otherNode.Start.Line}, {otherNode.Start.Column})");
+            }
+
+            foreach (var child in node.Children)
+            {
+                var childValue = child.Value;
+                var otherChildValue = otherNode.Children[child.Key];
```
rynowak, line 91: This will also blow up with a bad error in the event that the expected mapping isn't present in `otherNode`. Can we do a `TryGetValue` instead?

test/E2ETest/Infrastructure/EqualityYamlNodeVisitor.cs
```diff
@@ -0,0 +1,136 @@
...
+                    VisitMapping((YamlMappingNode)childValue, (YamlMappingNode)otherChildValue);
+                }
+                else if (childValue.NodeType == YamlNodeType.Scalar)
+                {
+                    VisitScalar((YamlScalarNode)childValue, (YamlScalarNode)otherChildValue);
+                }
+                else if (childValue.NodeType == YamlNodeType.Sequence)
+                {
+                    VisitSequence((YamlSequenceNode)childValue, (YamlSequenceNode)otherChildValue);
+                }
```
rynowak, line 109: call `Visit(...)`

test/E2ETest/Infrastructure/EqualityYamlNodeVisitor.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            if (node.Tag?.Equals(otherNode.Tag) == false)
+            {
+                throw new TyeYamlException($"Expected tags to be equal. " +
+                    $"Expected: ({node.Start.Line}, {node.Start.Column}) Actual: ({otherNode.Start.Line}, {otherNode.Start.Column})");
+            }
+
+            if (!node.Equals(otherNode))
+            {
+                throw new TyeYamlException($"Scalar nodes have different values." + Environment.NewLine +
```
rynowak, line 128: <3

test/E2ETest/TestHelpers.cs
```diff
@@ -16,6 +16,7 @@
 using Xunit;
 using Xunit.Abstractions;
 using Microsoft.Tye;
+using YamlDotNet.RepresentationModel;
```
rynowak, line 19: pref: make a new class and call it `YamlAssert`. That follows the pattern of xUnit.

This is why I always hate on "Helpers" - they accumulate a bunch of clutter 😆

test/E2ETest/TestHelpers.cs
```diff
@@ -221,5 +222,37 @@ await using (cancellation.Token.Register(() => stoppedTask.TrySetCanceled()))
...
+        {
+            var yamlStream = new YamlStream();
+            using var reader = new StringReader(expected);
+            yamlStream.Load(reader);
+
+            var otherYamlStream = new YamlStream();
+            using var otherReader = new StringReader(expected);
+            otherYamlStream.Load(new StringReader(actual));
+
+            var yamlEqualityVisitor = new EqualityYamlNodeVisitor();
```
rynowak, line 236: Do we need to normalize those newlines?

rynowak: I guess not. It might matter if we used multi-line strings?

test/E2ETest/TyeGenerateTests.cs
```diff
@@ -50,7 +51,8 @@ public async Task SingleProjectGenerateTest()
                 var content = await File.ReadAllTextAsync(Path.Combine(projectDirectory.DirectoryPath, $"{projectName}-generate-{environment}.yaml"));
                 var expectedContent = await File.ReadAllTextAsync($"testassets/generate/{projectName}.yaml");
 
-                Assert.Equal(expectedContent.NormalizeNewLines(), content.NormalizeNewLines());
+                TestHelpers.AssertYamlContentEqual(expectedContent, content, output);
```
rynowak, line 54: using static ftw

test/E2ETest/TestHelpers.cs
rynowak: If we're not going to need the Nomalize newlines thing should we get rid of it?

test/E2ETest/Infrastructure/EqualityYamlNodeVisitor.cs
jkotalik: I don't see much reason to track the entire path, but the previous node makes sense.

jkotalik: Didn't add a context here. I realized the only case where this could be null is in VisitYamlMapping, which I now handle at the location.

test/E2ETest/Infrastructure/EqualityYamlNodeVisitor.cs
jkotalik: Yeah, I had it here because I didn't want to throw if otherNode had less children. I can break up the check into two places though.

jkotalik: Actually, nvrm I see what you are getting at.

jkotalik:
I assume this is good to go?

davidfowl:
Yah, super clean. Great job @jkotalik !

Issue #320, opened 2020-04-08
Add support for container tools package 


rynowak:
We should support using https://docs.microsoft.com/en-us/visualstudio/containers/container-msbuild-properties?view=vs-2019 to build containers for deployment and local run

davidfowl:
Related to #240 (we may be able to close that one)

staff0rd:
Not sure if related or not, but I'd like to see a `tye deploy --containers-only` or similar, where the project container images are built, tagged & pushed to the current container registry, but not deployed to k8s.

rynowak:
Not related but good idea 👍

rynowak:
we do have `tye build` which will do the building and tagging. I'd be open to adding `tye push`  or `tye build --push`

staff0rd:
oh, I did not see that, but also decided that I need the feature now, so I'll write it today :)

rynowak:
@davidfowl @jkotalik - quick discussion on naming for command? 

I kinda prefer `tye push` to `tye build --push`. I know we don't have to have a ridiculous amount of commands, but I also think we'd hit the natural limit for build/deploy with `generate, build, push, deploy`.

rynowak:
For anyone watching this, we've added `tye push` as a top level command. Keep this issue around because that wasn't the original purpose.

PR #365, opened 2020-04-14
Add push command

staff0rd:
This is what i'm using for #320


davidfowl:
I like push.

rynowak:
awesome! tried it out and it works!

PR #376, opened 2020-04-17
Support dapr log level and config for local run

### LINKED ISSUE #296: How can I inject dapr.io/config: "tracing" using tye?

rynowak:
Fixes #296

Adds conditional logic for passing these properties to `daprd`.

Corrects logic for dealing with config. Dapr's config annotation in k8s
refers to kubernetes resource by name, but that's not possible with
local run. In local run `dapr` will expect the `-config` parameter to be
a file on disk. So, this change adds logic to look for a file by naming
convention and pass that locally instead.

Additionally added the ability for an extension to output to the console
- this is needed because we want to tell the user if their config file
can't be found, and specifically where we looked. I made the decision to
log a message and skip if the config file isn't present, because we
don't have a way to make settings conditional based on environment yet -
it makes sense that someone might need to have a config in production
but not in local dev.

### DIFF
--- a/src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
+++ b/src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Globalization;
+using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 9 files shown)


src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
```diff
@@ -147,14 +169,14 @@ public override Task ProcessAsync(ExtensionContext context, ExtensionConfigurati
...
 
-                        if (config.Data.TryGetValue("config", out var daprConfig) && daprConfig is object)
+                        if (config.Data.TryGetValue("config", out var obj) && obj?.ToString() is string daprConfig)
                         {
-                            deployment.Annotations.TryAdd("dapr.io/config", daprConfig!.ToString() ?? string.Empty);
+                            deployment.Annotations.TryAdd("dapr.io/config", daprConfig);
                         }
 
-                        if (config.Data.TryGetValue("log-level", out var logLevel) && logLevel is object)
+                        if (config.Data.TryGetValue("log-level", out obj) && obj?.ToString() is string logLevel)
```
rynowak, line 177: simplified this and made more resiliant. If `obj` is null or tostrings to null we will just skip.

PR #383, opened 2020-04-19
Add  buildProperties to project service configuration

alexfdezsauco:
Add buildProperties to project service configuration.

The service configuration could look like this:

    ...
    services:
    - name: projectA
      project: ProjectA/ProjectA.csproj
      buildProperties: 
      - name: Configuration
        value: Debug

### DIFF
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -58,6 +58,7 @@ namespace Microsoft.Tye
                     service = project;
 
                     project.Build = configService.Build ?? true;
+                    project.BuildArgs = configService.BuildArgs;
                     project.Args = configService.Args;
                     project.Replicas = configService.Replicas ?? 1;
 
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -58,7 +58,7 @@ namespace Microsoft.Tye
                     service = project;
 
                     project.Build = configService.Build ?? true;
-                    project.BuildArgs = configService.BuildArgs;
+                    project.Properties = configService.Properties.ToDictionary(cs => cs.Name, cs => cs.Value);
                     project.Args = configService.Args;
                     project.Replicas = configService.Replicas ?? 1;
 
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -91,22 +91,16 @@ namespace Microsoft.Tye
                     // Special handling of .dlls as executables (it will be executed as dotnet {dll})
                     if (Path.GetExtension(expandedExecutable) == ".dll")
                     {
-                        expandedExecutable = Path.GetFullPath(
-                            Path.Combine(builder.Source.Directory.FullName, expandedExecutable));
+                        expandedExecutable = Path.GetFullPath(Path.Combine(builder.Source.Directory.FullName, expandedExecutable));
                         workingDirectory = Path.GetDirectoryName(expandedExecutable)!;
                     }
 
                     var executable = new ExecutableServiceBuilder(configService.Name!, expandedExecutable)
                     {
                         Args = configService.Args,
-                        WorkingDirectory =
-                                             configService.WorkingDirectory != null
-                                                 ? Path.GetFullPath(
-                                                     Path.Combine(
-                                                         builder.Source.Directory.FullName,
-                                                         Environment.ExpandEnvironmentVariables(
-                                                             configService.WorkingDirectory)))
-                                                 : workingDirectory,
+                        WorkingDirectory = configService.WorkingDirectory != null ?
+                        Path.GetFullPath(Path.Combine(builder.Source.Directory.FullName, Environment.ExpandEnvironmentVariables(configService.WorkingDirectory))) :
+                        workingDirectory,
                         Replicas = configService.Replicas ?? 1
                     };
                     service = executable;
@@ -124,7 +118,9 @@ namespace Microsoft.Tye
                 builder.Services.Add(service);
 
                 // If there are no bindings and we're in ASP.NET Core project then add an HTTP and HTTPS binding
-                if (configService.Bindings.Count == 0 && service is ProjectServiceBuilder project2 && project2.IsAspNet)
+                if (configService.Bindings.Count == 0 &&
+                    service is ProjectServiceBuilder project2 &&
+                    project2.IsAspNet)
                 {
                     // HTTP is the default binding
                     service.Bindings.Add(new BindingBuilder() { Protocol = "http" });
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -58,18 +58,18 @@ namespace Microsoft.Tye
                     service = project;
 
                     project.Build = configService.Build ?? true;
-                    project.Properties = configService.Properties.ToDictionary(cs => cs.Name, cs => cs.Value);
                     project.Args = configService.Args;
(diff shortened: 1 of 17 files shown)


davidfowl:
Instead of build args it should be a dictionary of properties.

alexfdezsauco:
Ok, I will implement this with a dictionary. But there are entries in the dictionary that need to be transtaled as build command arguments. In this case, I extracted the "dictionary" only for the supported build configuration from the build args.

A question: What to do with the all properties in the dictionary? I know what to do with the build Configuration. But the others?

davidfowl:
> Ok, I will implement this with a dictionary. But there are entries in the dictionary that need to be transtaled as build command arguments. In this case, I extracted the "dictionary" only for the supported build configuration from the build args.

They can all be translated via /p:{Key}={Value}, it's just msbuild properties. You can choose to recognize some as first class but I wouldn't at the moment.

alexfdezsauco:
I tried something different for the moment, added properties but only build Configuration as first class

alexfdezsauco:
Also translated extra properties as /p:{Key}={Value}

src/Microsoft.Tye.Core/ConfigModel/ConfigService.cs
```diff
@@ -24,5 +24,6 @@ public class ConfigService
         public List<ConfigVolume> Volumes { get; set; } = new List<ConfigVolume>();
         [YamlMember(Alias = "env")]
         public List<ConfigConfigurationSource> Configuration { get; set; } = new List<ConfigConfigurationSource>();
+        public List<ConfigConfigurationSource> Properties { get; set; } = new List<ConfigConfigurationSource>();
```
davidfowl, line 27: Did you reuse configuration source because it exists?

davidfowl:
This change needs a test, my guess is that this will fail with --docker as well.

src/Microsoft.Tye.Core/ConfigModel/ConfigService.cs
alexfdezsauco: Yes

davidfowl: I wouldn't. I would be specific about the intent of this field. It should be called buildProperties and should be a BuildProperty object

alexfdezsauco: Ok

alexfdezsauco: Fixed

alexfdezsauco:
> This change needs a test, my guess is that this will fail with --docker as well.

Will check this one.

alexfdezsauco:
> This change needs a test, my guess is that this will fail with --docker as well.

Added support for --docker option, working on the test

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -90,16 +91,22 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
...
                     }
 
                     var executable = new ExecutableServiceBuilder(configService.Name!, expandedExecutable)
                     {
                         Args = configService.Args,
-                        WorkingDirectory = configService.WorkingDirectory != null ?
-                        Path.GetFullPath(Path.Combine(builder.Source.Directory.FullName, Environment.ExpandEnvironmentVariables(configService.WorkingDirectory))) :
-                        workingDirectory,
+                        WorkingDirectory =
+                                             configService.WorkingDirectory != null
```
davidfowl, line 103: What's up with this formatting here.

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -117,21 +124,12 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
                 builder.Services.Add(service);
 
                 // If there are no bindings and we're in ASP.NET Core project then add an HTTP and HTTPS binding
-                if (configService.Bindings.Count == 0 &&
-                    service is ProjectServiceBuilder project2 &&
-                    project2.IsAspNet)
+                if (configService.Bindings.Count == 0 && service is ProjectServiceBuilder project2 && project2.IsAspNet)
```
davidfowl, line 127: Why did the formatting change here?

src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -11,6 +11,7 @@
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.Loader;
+using System.Text.RegularExpressions;
```
davidfowl, line 14: nit: Remove

src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -51,12 +53,32 @@ private async Task BuildAndRunProjects(Application application)
...
                 string args;
+                var buildArgs = string.Empty;
                 string workingDirectory;
                 if (serviceDescription.RunInfo is ProjectRunInfo project)
                 {
                     path = project.RunCommand;
                     workingDirectory = project.ProjectFile.Directory.FullName;
                     args = project.Args == null ? project.RunArguments : project.RunArguments + " " + project.Args;
+
+                    foreach (var supportedProperty in BuildPropertiesToOptionsMap)
```
davidfowl, line 64: Why do you need this?

test/E2ETest/Infrastructure/ConditionalTheoryAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Xunit;
+using Xunit.Sdk;
+
+namespace E2ETest
+{
+    using System;
```
davidfowl, line 11: Fix the usings to be out of the namespace..

src/Microsoft.Tye.Hosting/ProcessRunner.cs
alexfdezsauco: Just removed, I thought that some properties must be turned into command option. But is not required.

test/E2ETest/Infrastructure/ConditionalTheoryAttribute.cs
alexfdezsauco: Fixed

src/Microsoft.Tye.Core/ApplicationFactory.cs
alexfdezsauco: Not sure, I don't remember touching such part. But also fixed back.

src/Microsoft.Tye.Core/ApplicationFactory.cs
alexfdezsauco: Again, not sure. But also fixed back.

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -59,16 +59,17 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
...
 
                     await ProjectReader.ReadProjectDetailsAsync(output, project);
 
                     // We don't apply more container defaults here because we might need
                     // to prompt for the registry name.
-                    project.ContainerInfo = new ContainerInfo()
-                    {
-                        UseMultiphaseDockerfile = false,
-                    };
+                    project.ContainerInfo = new ContainerInfo() { UseMultiphaseDockerfile = false, };
```
davidfowl, line 72: The reformatting changes make it really hard to tell what actually changed. In the future when you send more PRs (and please send more PRs these are great) I'd refrain from doing these changes as they are a bit distracting in the diff.

davidfowl: No problem. Your contributing some great stuff! Keep it coming!

Issue #362, opened 2020-04-13
Fix nbgv and tye for docker image versions


rynowak:
This regressed in the move from rynowak/opulence.

Tye should be able to work with the version that's set by `NerdBank.GitVersioning`. Need to test that this works after a *clean*.

PR #384, opened 2020-04-20
Address feedback about how we do evaluation

rynowak:
Fixes: #362

This change splits up our evaluation code-path into a few steps:

- Load project
- Do a restore
- Mark dirty
- Run targets and read properties

The key is that we need to do a restore separately from other targets
because restore will bring in nuget packages that contain targets. You
need to let MSBuild load those before continuing. The code here is
pretty similar to what dotnet does.

Additionally blocking the NBGV scenario, there are some additional
assemblies that need to be provided by MSBuild. NBGV also loads
Microsoft.DotNet.Platform.Abstractions which is part of the SDK. Since
we had an allow-list of just nuget assemblies that wasn't resolved.

### DIFF
--- a/src/Microsoft.Tye.Core/ProjectReader.cs
+++ b/src/Microsoft.Tye.Core/ProjectReader.cs
@@ -145,16 +145,24 @@ namespace Microsoft.Tye
         {
             var sw = Stopwatch.StartNew();
 
+            // Currently we only log at debug level.
+            var logger = new ConsoleLogger(
+                verbosity: LoggerVerbosity.Normal,
+                write: message => output.WriteDebug(message),
+                colorSet: null,
+                colorReset: null);
+
             // We need to isolate projects from each other for testing. MSBuild does not support
             // loading the same project twice in the same collection.
             var projectCollection = new ProjectCollection();
 
             ProjectInstance projectInstance;
+            Microsoft.Build.Evaluation.Project msbuildProject;
 
             try
             {
                 output.WriteDebugLine($"Loading project '{project.ProjectFile.FullName}'.");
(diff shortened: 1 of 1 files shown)


davidfowl:
How much slower is it?

src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -166,18 +174,45 @@ private static void EvaluateProject(OutputContext output, ProjectServiceBuilder
...
+                    targetsToBuild: new[] { "Restore" },
+                    hostServices: null,
+                    flags: BuildRequestDataFlags.ClearCachesAfterBuild | BuildRequestDataFlags.SkipNonexistentTargets | BuildRequestDataFlags.IgnoreMissingEmptyAndInvalidImports);
+
+                var parameters = new BuildParameters(projectCollection)
+                {
+                    Loggers = new[] { logger, },
+                };
+
+                // We don't really look at the result, because it's not clear we should halt totally
```
davidfowl, line 198: This is worth a `REVIEW:` comment

rynowak: what is a review comment?

rynowak:
I think it's about 50-100ms slower per project, it's not that bad.

src/Microsoft.Tye.Core/ProjectReader.cs
jkotalik: A comment which collects dust 😆

rynowak: ok, then, I'm deleting the comment in advance :)

PR #385, opened 2020-04-20
Implement cleanup command

rynowak:
This is new command that can be used to delete everything running in a
cluster related to an application using the labels we use.

This currently uses a hardcoded set of resource types. It would be
possible to make this generic, which might become needed in the future.
See comments in the code.

---

The overall design for how to use the command:

```sh
tye cleanup
```

Will read a solution or tye.yaml to get the application name. It then
enumerates and deletes all resources in the cluser with
`app.kubernetes.io/part-of=application-name` (in the current namespace).

```sh
tye cleanup --what-if
```

Will do the same, but without actually doing the deletion.

```sh
tye cleanup -i
```

Will interactively give you a yes/no choice for whether to delete each
resource.


docs/reference/commandline/tye.md
```diff
@@ -29,6 +29,7 @@ For example, [`tye run`](tye-run.md) runs an application. Each command defines i
 
 | Command                                       | Function                                                            |
 | --------------------------------------------- | ------------------------------------------------------------------- |
-| [tye init](tye-init.md) | Scaffolds a tye.yaml file representing the application.                          |
-| [tye run](tye-run.md)               | Runs an application.                                     |
-| [tye deploy](tye-deploy.md)               | Deploys an application.                                             |
+| [tye init](tye-init.md)                       | Scaffolds a tye.yaml file representing the application.             |
+| [tye run](tye-run.md)                         | Runs an application.                                                |
+| [tye deploy](tye-deploy.md)                   | Deploys an application.                                             |
+| [tye cleanup](tye-cleanup.md)                 | Deletes a deployed application.                                     |
```
davidfowl, line 35: clean?

rynowak: I don't really love either of these tbh. I called it cleanup expecting a debate on the PR. I don't like `tye clean` because we have tye build, and it makes it seem like `tye build` and `tye clean` are related.

Some ideas:

- `tye cleanup` (current)
- `tye clean` (I this it will confuse people)
- `tye teardown` 
- `tye undeploy`
- `tye delete`

davidfowl: tye undeploy ftw

jkotalik: I think `tye clean` or `tye delete` makes the most sense, and it depends on what we want this command to do.

If we just want it to delete any resources that are in the cluster, I think we should go with just `delete` and maybe allow subcomponents (like accepting a name for a service).

If we want this command to reset or modify state of the cluster besides deletion, then `clean` may be more appropriate.

docs/reference/commandline/tye-cleanup.md
```diff
@@ -0,0 +1,70 @@
...
+tye cleanup [-?|-h|--help] [-i|--interactive] [-v|--verbosity <Debug|Info|Quiet>] [--what-if] [<PATH>]
+```
+
+## Description
+
+The `tye cleanup` command will delete a deployed application from Kubernetes. `tye cleanup` by default will:
+
+- List all resources that are part of an application
+- Print the list of resources (what-if)
+- Offer a choice delete each resource (interactive)
```
jkotalik, line 19:
```suggestion
- Offer a choice to delete each resource (interactive)
```

docs/reference/commandline/tye-cleanup.md
```diff
@@ -0,0 +1,70 @@
...
+
+`tye cleanup` operates in the current Kubernetes namespace. Use `kubectl config view --minify --output 'jsonpath={..namespace}'` to view the current namespace.
+
+Cleanup decides which resources to delete based on the `app.kubernetes.io/part-of=...` label. This label will be set to the application name for all resources created by Tye. `tye cleanup` does not rely on the list of services in `tye.yaml` or a solution file. 
+
+## Arguments
+
+`PATH`
+
+The path to either a file or directory to execute `tye cleanup` on. Can either be a yaml, sln, or project file, however it is recommend to have a tye.yaml file for `tye cleanup`.
```
jkotalik, line 30: Consider allowing service name, so someone can delete everything associated with a service

``` 
tye delete frontend
```

docs/reference/commandline/tye-cleanup.md
```diff
@@ -0,0 +1,70 @@
...
+
+    The verbosity of logs emitted by `tye cleanup`. Defaults to Info.
+
+- `--what-if`
+
+    Print each resource instead of deleting.
+
+## Examples
+
+- Deploy an application from the current directory:
```
jkotalik, line 54: Fix these examples.

src/tye/CleanupHost.cs
```diff
@@ -0,0 +1,220 @@
...
+
+            // We don't need to know anything about the services, just the application name.
+            var application = ConfigFactory.FromFile(path);
+
+            await ExecuteCleanupAsync(output, application, interactive, whatIf);
+        }
+
+        public static async Task ExecuteCleanupAsync(OutputContext output, ConfigApplication application, bool interactive, bool whatIf)
+        {
+            var config = KubernetesClientConfiguration.BuildDefaultConfig();
```
jkotalik, line 35: Would we want this cleanup task to be reusable in anyway (or pluggable into the deploy pipeline)? I think this is fine for now though to keep it simple.

src/tye/CleanupHost.cs
```diff
@@ -0,0 +1,220 @@
...
+            var kubernetes = new Kubernetes(config);
+
+            // Due to some limitations in the k8s SDK we currently have a hardcoded list of resource
+            // types that we handle deletes for. If we start adding extensibility for the *kinds* of
+            // k8s resources we create, or the ability to deploy additional files along with the
+            // resources we understand then we should revisit this.
+            //
+            // Basically the challenges are:
+            //
+            // - kubectl api-resources --all (and similar) are implemented client-side (n+1 problem)
```
jkotalik, line 51: What is the n+1 problem? I'm unfamiliar with it.

docs/reference/commandline/tye-cleanup.md
rynowak: I don't want to do service name because I think that raises a lot of questions - and we don't have per-service deployment yet.

For instance, does it update the config of every other service to remove its env-vars? (Probably not).

Per-application cleanup is easier to reason about because things like secrets/bindings are shared. Basically we *could* support per-service cleanup, but then what happens? does the usage of it by other services just dangle?

src/tye/CleanupHost.cs
rynowak: The [n+1 problem](https://stackoverflow.com/questions/97197/what-is-the-n1-selects-problem-in-orm-object-relational-mapping) is a problem that comes up a lot in data design - usually with databases.

Basically it's named for having to do a query to discover `N` and then doing `N` queries to resolve all of the results. Querying all resources in k8s requires that, you have to enumerate all of the *kinds of things*, and then enumerate all of the things *by kind*.

src/tye/Program.CleanupCommand.cs
```diff
@@ -0,0 +1,46 @@
...
+                CommonArguments.Path_Required,
+                StandardOptions.Interactive,
+                StandardOptions.Verbosity,
+
+                new Option(new[]{ "--what-if", }, "print what would be deleted without making changes")
+                {
+                    Argument = new Argument<bool>(),
+                },
+            };
+
```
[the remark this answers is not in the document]
rynowak: @JamesNK

docs/reference/commandline/tye-cleanup.md
jkotalik: That seems well justified.

src/tye/CleanupHost.cs
rynowak: I want to keep it simple. We can refactor when needed.

docs/reference/commandline/tye-cleanup.md
rynowak: What do you mean fix? You mean update them when we decide on a command name?

docs/reference/commandline/tye.md
rynowak: > If we just want it to delete any resources that are in the cluster, I think we should go with just

It definitely refers to the cluster. I'm kinda leaning towards `undeploy` because:

- it's clear that it undoes deploy
- it's a name you wouldn't guess, so you don't end up accidentally nuking the world

jkotalik: Agreed with `undeploy` offline

rynowak: Ok. going with undeploy.

Issue #100, opened 2020-03-12
Make multi-project sample more resiliant


rynowak:
These are the logs I get running `tye run` on the multi-project sample:

```txt
[10:39:56 INF] Executing application from /Users/ryan/github.com/dotnet/tye/samples/multi-project/tye.yaml
[10:39:56 INF] Default dashboard port 8000 has been reserved by the application, choosing random port.
[10:39:56 INF] Dashboard running on http://127.0.0.1:61320
[10:39:56 INF] Mapping external port 8000 to internal port(s) 61321, 61322 for frontend binding null
[10:39:57 INF] Running docker command run -d -e DOTNET_ENVIRONMENT=Development -e DOTNET_LOGGING__CONSOLE__DISABLECOLORS=true -e RABBITMQ_PORT=5672 -e SERVICE__BACKEND__PORT=7000 -e BACKEND_SERVICE_PORT=7000 -e SERVICE__BACKEND__HOST=host.docker.internal -e BACKEND_SERVICE_HOST=host.docker.internal -e SERVICE__FRONTEND__PORT=8000 -e FRONTEND_SERVICE_PORT=8000 -e SERVICE__FRONTEND__HOST=host.docker.internal -e FRONTEND_SERVICE_HOST=host.docker.internal -e SERVICE__RABBIT__PROTOCOL=rabbitmq -e RABBIT_SERVICE_PROTOCOL=rabbitmq -e SERVICE__RABBIT__PORT=5672 -e RABBIT_SERVICE_PORT=5672 -e SERVICE__RABBIT__HOST=host.docker.internal -e RABBIT_SERVICE_HOST=host.docker.internal -e APP_INSTANCE=rabbit_7d6a3050-1  -p 5672:5672 --name rabbit_7d6a3050-1 --restart=unless-stopped rabbitmq:3-management
[10:39:57 INF] Building project /Users/ryan/github.com/dotnet/tye/samples/multi-project/backend/backend.csproj
[10:39:57 INF] Building project /Users/ryan/github.com/dotnet/tye/samples/multi-project/frontend/frontend.csproj
[10:39:57 INF] Building project /Users/ryan/github.com/dotnet/tye/samples/multi-project/worker/worker.csproj
[10:39:58 INF] Launching service worker_1aa571d1-8: dotnet "/Users/ryan/github.com/dotnet/tye/samples/multi-project/worker/bin/Debug/netcoreapp3.1/worker.dll"
[...]

PR #391, opened 2020-04-20
Make multi-project sample more resiliant

rynowak:
Fixes: #100

Makes the sample/test more idiomatic and resiliant


rynowak:
windows build is still queued :(

PR #392, opened 2020-04-21
Allow tye yamls to be merged.

jkotalik:
Start of multirepo support (https://github.com/dotnet/tye/issues/239).

This adds the ability to have a service reference to a `tye.yaml` file inside of a `tye.yaml`. Currently, the config would look like:

```yaml
name: VotingSample
services:
- name: vote
  project: vote.csproj
- name: worker
  include: ../worker/tye.yaml
```

Working on a few tests as well. Drafting to figure out if people like the structure before writing a ton of tests.


rynowak:
Structure is what I expected to see 👍 

it would be cool if we could test this at the data layer - start writing assertions about `ApplicationBuilder`.

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -21,227 +21,257 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
...
 
-            var config = ConfigFactory.FromFile(source);
-            config.Validate();
-            var builder = new ApplicationBuilder(source, config.Name ?? source.Directory.Name.ToLowerInvariant());
-            if (!string.IsNullOrEmpty(config.Registry))
-            {
-                builder.Registry = new ContainerRegistry(config.Registry);
-            }
+            var queue = new Queue<ConfigApplication>();
+            var visited = new HashSet<string>();
```
rynowak, line 25: This needs a comparer - on windows at least some APIs normalize case and some preserve it. Depending on the API used you might see the same file twice with two casings.

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -21,227 +21,257 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
...
-            }
+            queue.Enqueue(rootConfig);
 
-            foreach (var configService in config.Services)
+            while (queue.Count > 0)
             {
-                ServiceBuilder service;
-                if (!string.IsNullOrEmpty(configService.Project))
+                var config = queue.Dequeue();
+                if (visited.Contains(config.Source.FullName))
```
rynowak, line 37: nit: you can use `if (!visited.Add(...))`

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -21,227 +21,257 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
...
                     }
-                    else if (service is ExecutableServiceBuilder executable)
+                    else if (!string.IsNullOrEmpty(configService.Import))
                     {
-                        executable.EnvironmentVariables.Add(envVar);
+                        var expandedYaml = Environment.ExpandEnvironmentVariables(configService.Import);
+                        var nestedConfig = ConfigFactory.FromFile(new FileInfo(Path.Combine(config.Source.DirectoryName, expandedYaml)));
+                        nestedConfig.Validate();
+                        queue.Enqueue(nestedConfig);
+                        // TODO do we need a service builder here?
```
rynowak, line 133: this should be represented by a service builder, but I would expect it to be one of the existing types.

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -21,227 +21,257 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
...
+                            Path = configRule.Path,
+                            Service = configRule.Service,
+                        };
+                        ingress.Rules.Add(rule);
+                    }
                 }
             }
 
-            return builder;
+            return root;
```
rynowak, line 274: One thing I expected to see here:

```yaml
# a.yaml
services:
- name: foo
  project: ....
- name: bar
   include: ../some-path/b.yaml
...
# b.yaml
services:
- name: bar
  project: ....
- name: redis
   image: redis
```

Currently right now you'll get an application with 3 services, `foo`, `bar`, and `redis`. Everything is visible to everything else, meaning that `redis` is a dependency of `foo` and `bar` - they both get access to `redis`'s environment variables.

I expected to see this dependency information represented somewhere, and the binding processing code updated to handle it.

src/shared/TempDirectory.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.Tye
 {
-    internal class TempDirectory : IDisposable
+    public class TempDirectory : IDisposable
```
rynowak, line 11: 😲

rynowak: this is already shared code? If we need it in other places I'm good with either making it public somewhere, or linking more places, but probably not both

test/UnitTests/ApplicationFactoryTests.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.Tye;
+using Microsoft.Tye.ConfigModel;
+using Tye;
+using Tye.Serialization;
+using Xunit;
+using Xunit.Abstractions;
+using YamlDotNet.RepresentationModel;
+using YamlDotNet.Serialization;
+using YamlDotNet.Serialization.NamingConventions;
+using static E2ETest.TestHelpers;
```
rynowak, line 14: this sounds like a good use case for an additional project or a shared folder. I get kinda worried when test projects reference test projects

test/UnitTests/ApplicationFactoryTests.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System.IO;
```
rynowak, line 1: Needs copyright

test/UnitTests/ApplicationFactoryTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        public async Task MultiRepo_Cycles()
+        {
+            // tye.yaml
+            //  tye.yaml loops to first tye.yaml
+
+            using var projectDirectory = CopyTestProjectDirectory(Path.Combine("single-project", "test-project"));
+
+            var projectFile = new FileInfo(Path.Combine(projectDirectory.DirectoryPath, "test-project.csproj"));
+
+            // Debug targets can be null if not specified, so make sure calling host.Start does not throw.
```
rynowak, line 37: bad copypast?

test/UnitTests/ApplicationFactoryTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        public ApplicationFactoryTests(ITestOutputHelper output)
+        {
+            _sink = new TestOutputLogEventSink(output);
+        }
+
+        [Fact]
+        public async Task MultiRepo_Cycles()
+        {
+            // tye.yaml
+            //  tye.yaml loops to first tye.yaml
```
rynowak, line 31: this comment doesn't make tons of sense

src/Microsoft.Tye.Core/ApplicationFactory.cs
rynowak: This seems pretty subtle, but it also kinda important. The reason why that it becomes possible to have a dependency in code on something you don't have in `tye.yaml`. So if we inject env-vars for a service you don't claim to reference, you've effectively written working code for something undocumented.

rynowak: This might be worth more debate or discussion, but I was assuming the services you *didn't* include to be private dependencies

test/UnitTests/ApplicationFactoryTests.cs
jkotalik: Yeah this was the main draft portion of this PR 😄

src/Microsoft.Tye.Core/ApplicationFactory.cs
jkotalik: If there is a reference to a specific service in a tye.yaml, for example:

```yaml
name: VotingSample
services:
- name: include
  include: ../worker/tye.yaml
- name: results
  project: results.csproj
```

```yaml
name: VotingSample
services:
- name: redis
  image: redis
  bindings:
    - port: 6379
      connectionString: ${host}:${port}
- name: worker
  project: worker.csproj
- name: postgres
  image:  postgres
  env:
  - name: POSTGRES_PASSWORD
    value: "pass@word1"
  bindings:
  - port: 5432
    connectionString: Server=${host};Port=${port};User Id=postgres;Password=${env:POSTGRES_PASSWORD};

```

We would have two service builders with the same name (worker) because we added a service builder for worker in both the results tye.yaml and the worker tye.yaml. Hence I didn't do it here yet, but wanted feedback/thoughts.

src/Microsoft.Tye.Core/ApplicationFactory.cs
jkotalik: Yeah that is an important distinction that we should fix. I'll look into how to make it so environment variables are only injected for direct dependents.

test/UnitTests/ApplicationFactoryTests.cs
jkotalik: I thought test projects didn't need copyright.

src/Microsoft.Tye.Core/ApplicationFactory.cs
jkotalik: This change is actually a bit more difficult than I thought. I'm going to make a follow up PR with this change after merging this.

PR #395, opened 2020-04-22
Enable running restore.sh

rynowak:


rynowak:
Bypassing CI to avoid using cores 👍 This script is just for local use.

PR #396, opened 2020-04-23
Improve docs somewhat.

brendandburns:
cc @davidfowl @jkotalik 

Partially addresses #367

### DIFF
--- a/docs/tutorials/hello-tye/00_run_locally.md
+++ b/docs/tutorials/hello-tye/00_run_locally.md
@@ -129,18 +129,21 @@ Now that we have two applications running, let's make them communicate. By defau
     dotnet add frontend/frontend.csproj package Microsoft.Tye.Extensions.Configuration  --version "0.1.0-*"
     ```
 
-5. Now register this client in `Startup.cs` class in `ConfigureServices` of the `frontend` project:
+5. Now register this client in `frontend` by adding the following to the existing `ConfigureServices` method to the existing `Startup.cs` file:
 
    ```C#
+   ...
    public void ConfigureServices(IServiceCollection services)
    {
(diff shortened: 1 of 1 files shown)


docs/tutorials/hello-tye/00_run_locally.md
```diff
@@ -129,35 +129,42 @@ Now that we have two applications running, let's make them communicate. By defau
...
        services.AddRazorPages();
-
+        /** Add the following to wire the client to the backend **/
        services.AddHttpClient<WeatherClient>(client =>
        {
             client.BaseAddress = Configuration.GetServiceUri("backend");
        });
+       /** End added code **/
    }
+   ...
```
loic-sharma, line 146: What're your thoughts on using diff syntax?

```diff
  ...
  public void ConfigureServices(IServiceCollection services)
  {
      services.AddRazorPages();

+      services.AddHttpClient<WeatherClient>(client =>
+      {
+          client.baseAddress = Configuration.GetServiceUri("backend");
+      });
  }
...
```

rynowak: What I don't love about diff syntax is that it's not easy copy-pastable because of the `+` at the beginning of each line. The fencing with comments, I think makes it pretty clear.

davidfowl:
Apparently this is much simpler. Getting clarification for an update

rynowak:
> Apparently this is much simpler. Getting clarification for an update

Oops. I was testing this out while you commented and didn't see that. We can simplify this anyway now that it's merged.

rynowak:
I'm going to add doing the tutorial(s) with linux to our release checklist since no one on the team devs with linux.

PR #400, opened 2020-04-23
Add the ability to specify a namespace for tye deploy.

brendandburns:
cc @davidfowl 

I monkey-patched the path from the CLI down into the generator. Let me know if there is a preferred path to pass this through.

### DIFF
--- a/src/Microsoft.Tye.Core/GenerateKubernetesManifestStep.cs
+++ b/src/Microsoft.Tye.Core/GenerateKubernetesManifestStep.cs
@@ -12,6 +12,8 @@ namespace Microsoft.Tye
 
         public string Environment { get; set; } = "production";
 
+        public string? Namespace { get; set; } = null;
+
 
         public override Task ExecuteAsync(OutputContext output, ApplicationBuilder application, ServiceBuilder service)
         {
--- a/src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
+++ b/src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
@@ -26,6 +26,9 @@ namespace Microsoft.Tye
             var metadata = new YamlMappingNode();
             root.Add("metadata", metadata);
             metadata.Add("name", project.Name);
+            if (!String.IsNullOrEmpty(application.Namespace)) {
+                metadata.Add("namespace", application.Namespace);
+            }
 
             if (service.Annotations.Count > 0)
             {
@@ -104,6 +107,9 @@ namespace Microsoft.Tye
             var metadata = new YamlMappingNode();
             root.Add("metadata", metadata);
             metadata.Add("name", project.Name);
+            if (!String.IsNullOrEmpty(application.Namespace)) {
+                metadata.Add("namespace", application.Namespace);
+            }
 
             if (deployment.Annotations.Count > 0)
             {
--- a/src/tye/Program.DeployCommand.cs
+++ b/src/tye/Program.DeployCommand.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.CommandLine;
 using System.CommandLine.Invocation;
@@ -21,6 +22,7 @@ namespace Microsoft.Tye
                 CommonArguments.Path_Required,
                 StandardOptions.Interactive,
                 StandardOptions.Verbosity,
+                StandardOptions.Namespace,
             };
 
             command.AddOption(new Option(new[] { "-f", "--force" })
@@ -29,7 +31,7 @@ namespace Microsoft.Tye
                 Required = false
             });
 
-            command.Handler = CommandHandler.Create<IConsole, FileInfo, Verbosity, bool, bool>(async (console, path, verbosity, interactive, force) =>
+            command.Handler = CommandHandler.Create<IConsole, FileInfo, Verbosity, bool, bool, string>(async (console, path, verbosity, interactive, force, @namespace) =>
             {
                 // Workaround for https://github.com/dotnet/command-line-api/issues/723#issuecomment-593062654
                 if (path is null)
@@ -45,7 +47,10 @@ namespace Microsoft.Tye
                 {
                     throw new CommandException($"No services found in \"{application.Source.Name}\"");
                 }
-
+                if (!String.IsNullOrEmpty(@namespace))
+                {
+                    application.Namespace = @namespace;
+                }
                 await ExecuteDeployAsync(new OutputContext(console, verbosity), application, environment: "production", interactive, force);
             });
 
@@ -75,7 +80,7 @@ namespace Microsoft.Tye
                 new ValidateSecretStep() { Environment = environment, Interactive = interactive, Force = force, },
             };
 
-            steps.Add(new GenerateKubernetesManifestStep() { Environment = environment, });
+            steps.Add(new GenerateKubernetesManifestStep() { Environment = environment, Namespace = application.Namespace });
             steps.Add(new DeployServiceYamlStep() { Environment = environment, });
 
             ApplyRegistryAndDefaults(output, application, interactive, requireRegistry: true);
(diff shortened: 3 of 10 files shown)


src/Microsoft.Tye.Core/GenerateKubernetesManifestStep.cs
```diff
@@ -12,6 +12,8 @@ public sealed class GenerateKubernetesManifestStep : ServiceExecutor.Step
 
         public string Environment { get; set; } = "production";
 
+        public string Namespace { get; set; } = "default";
```
jkotalik, line 15: I think we need to be able to specify the namespace in the `tye.yaml` file as well.

```yaml
namespace: mynamespace
name: myapp
...
```

Potentially, the namespace information could be used by `tye run` as well for isolation.

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
```diff
@@ -26,6 +26,9 @@ internal static class KubernetesManifestGenerator
             var metadata = new YamlMappingNode();
             root.Add("metadata", metadata);
             metadata.Add("name", project.Name);
+            if (!String.IsNullOrEmpty(project.Namespace)) {
```
jkotalik, line 29:
```suggestion
            if (!string.IsNullOrEmpty(project.Namespace))
            {
```

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
```diff
@@ -104,6 +107,9 @@ internal static class KubernetesManifestGenerator
...
```
jkotalik, line 110:
```suggestion
            if (!string.IsNullOrEmpty(project.Namespace))
            {
```

src/tye/Program.DeployCommand.cs
```diff
@@ -29,7 +30,15 @@ public static Command CreateDeployCommand()
...
-            command.Handler = CommandHandler.Create<IConsole, FileInfo, Verbosity, bool, bool>(async (console, path, verbosity, interactive, force) =>
+            command.AddOption(new Option(new [] { "-n", "--namespace"})
+            {
+                Description = "Specify the namespace for the deployment",
+                Required = false,
+                Argument = new Argument<string>(),
+                Name = "ns"
+            });
+
+            command.Handler = CommandHandler.Create<IConsole, FileInfo, Verbosity, bool, bool, string>(async (console, path, verbosity, interactive, force, @namespace) =>
```
jkotalik, line 41: Does specifying `ns` instead of `@namespace` work here?

src/tye/Program.DeployCommand.cs
```diff
@@ -104,7 +113,11 @@ private static async Task DeployApplicationManifestAsync(OutputContext output, A
             }
 
             output.WriteDebugLine("Running 'kubectl apply'.");
-            output.WriteCommandLine("kubectl", $"apply -f \"{tempFile.FilePath}\"");
+            var cmd = $"apply -f \"{tempFile.FilePath}\"";
+            if (!String.IsNullOrWhiteSpace(ns)) {
```
jkotalik, line 117:
```suggestion
            if (!string.IsNullOrEmpty(ns))
            {
```

src/tye/Program.DeployCommand.cs
```diff
@@ -104,7 +113,11 @@ private static async Task DeployApplicationManifestAsync(OutputContext output, A
...
+                cmd = cmd + $" --namespace {ns}";
```
jkotalik, line 118: I don't think you should include the namespace in the output here, as the namespace information is in the tempFile here. Maybe changing the above log for `Running 'kubectl apply'.` to `Running 'kubectl apply' on namespace {ns}.

jkotalik:
There are some tests you need to update that check the k8s content with the new namespace mapping.

src/Microsoft.Tye.Core/GenerateKubernetesManifestStep.cs
rynowak, line 15: I think a better default for this property would be `null`. Currently we use the namespace of the current context (different from being hardcoded to `default`).

rynowak: Yeah, it would be nice to have if this could be read from the config as well. I can see the reason why you'd want both - we don't currently expose the concept of environments to the user, so it makes sense to give manual control via CLI - and to make it possible to check in.

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
rynowak: @brendandburns - if you run `dotnet format` at the command line it should autoformat stuff for you. You may need to run `dotnet tool restore` to bring down the tool. I'm not sure if the build script installs it or not.

src/tye/Program.DeployCommand.cs
rynowak: `@namespace` is better 😆

rynowak: However you may need to have https://github.com/dotnet/tye/pull/400/files#diff-c1c4c07b0626c051d8c7848f451427c8R38 match the parameter name of the lambda here. This command line parser does something kinda wierd where it requires the parameter names of the lambda to match how the parameters are defined. 

I say if it's working, leave it as-is :)

src/tye/Program.DeployCommand.cs
rynowak: yeah, it's little surprising to me that this ends up in the manifest and the command line. Is there a reason for that?

We actually have a few code paths that interact with manifest generation - have a currently undocumented command that generates the manifest to a file on disk https://github.com/dotnet/tye/blob/master/src/tye/Program.GenerateCommand.cs

so it seems better to thread this through `GenerateKubernetesManifestStep` than the command line

src/tye/Program.DeployCommand.cs
brendandburns: this does work. `ns` doesn't because of the naming.

src/Microsoft.Tye.Core/GenerateKubernetesManifestStep.cs
brendandburns: done.

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
brendandburns: done.

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
brendandburns: done.

src/tye/Program.DeployCommand.cs
brendandburns: done.

src/tye/Program.DeployCommand.cs
brendandburns: done.

brendandburns:
Ok, I think I addressed all of the changes, this is much cleaner, thanks!

Please take another look.

brendandburns:
fwiw, the format is flagging on line-endings, but dotnet format on my linux machine shows no problems and makes no changes. I assume that this is a Linux LF vs Windows CRLF thing.

brendandburns:
I would suggest that this project converge on LF instead of CRLF.

rynowak:
> fwiw, the format is flagging on line-endings, but dotnet format on my linux machine shows no problems and makes no changes. I assume that this is a Linux LF vs Windows CRLF thing.

Thanks for the feedback, I'm going to follow up on this and some of the other formatting related things. I'm not having the same problem on mac, but it be due to how I've configured something.

rynowak:
Thanks!

brendandburns:
You can definitely adjust git settings to do the right thing wrt line endings, we just need to decide on what the canonical line ending we want for this repo (I'd suggest LF) and then everyone can configure git to do the conversion.

--brendan

________________________________

PR #401, opened 2020-04-23
Remove PackageVersion tag with specific version number from tye.csproj

alexfdezsauco:
Remove PackageVersion tag with specific version number from tye.csproj

### DIFF
--- a/src/tye/tye.csproj
+++ b/src/tye/tye.csproj
@@ -7,7 +7,6 @@
     <AssemblyName>tye</AssemblyName>
     <PackageId>Microsoft.Tye</PackageId>
     <ToolCommandName>tye</ToolCommandName>
-    <PackageVersion>0.2.0-dev-1</PackageVersion>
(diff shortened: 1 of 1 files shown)


jkotalik:
For my knowledge, why did you add this in the first place for your original PR? For local testing?

alexfdezsauco:
Yes, for a local testing. Actually, for local development, while I waiting for the acceptance and the official prerelease version. But I forgot to ignore it in the PR. Sorry.

jkotalik:
I see for testing out of repo.

alexfdezsauco:
Can you also remove this microsoft.tye.0.2.0-dev-1 from the nuget feed? `d` is greater than `a` so, the next alpha updates will be not installed using `dotnet tool install -g Microsoft.Tye --version "0.2.0*"`.

jkotalik:
@rynowak do you know how to do this for the azdo feed?

rynowak:
this should be fixed now

PR #404, opened 2020-04-24
Fix line endings and formatting

rynowak:
We were missing a `.gitattributes` so it means that files ended up with mixed line-endings depending how the user's git/editor was configured. A lot of the sample code that was generated by templates in particular had all CRLF. 

Thanks @brendandburns for pointing this out!

After nuking my working copy checking out files again only the solutions are left with CRLF:

image


.gitattributes
```diff
@@ -0,0 +1,4 @@
+* text=auto
+*.cs text=auto diff=csharp 
+*.sh text eol=lf
+*.sln text eol=crlf
```
rynowak, line 4: The actual diff in this PR is pretty useless because it's almost all line-endings other than this.

rynowak: Explicitly setting `eol=crlf` here means it will be CRLF on all platforms. Setting `auto` means that files will be checked in as LF but may be checked out as whatever the user/os prefers

PR #408, opened 2020-04-25
Fix docker image building

davidfowl:
- Use the assembly name as the entrypoint instead of the project file name.

### DIFF
--- a/src/Microsoft.Tye.Core/DockerfileGenerator.cs
+++ b/src/Microsoft.Tye.Core/DockerfileGenerator.cs
@@ -42,7 +42,7 @@ namespace Microsoft.Tye
             await using var stream = File.OpenWrite(filePath);
             await using var writer = new StreamWriter(stream, encoding: Encoding.UTF8, bufferSize: -1, leaveOpen: true);
 
-            var entryPoint = Path.GetFileNameWithoutExtension(project.ProjectFile.Name);
+            var entryPoint = project.AssemblyName;
             output.WriteDebugLine($"Writing Dockerfile to '{filePath}'.");
             if (container.UseMultiphaseDockerfile ?? true)
             {


src/Microsoft.Tye.Core/DockerfileGenerator.cs
```diff
@@ -42,7 +42,7 @@ public static async Task WriteDockerfileAsync(OutputContext output, ApplicationB
             await using var stream = File.OpenWrite(filePath);
             await using var writer = new StreamWriter(stream, encoding: Encoding.UTF8, bufferSize: -1, leaveOpen: true);
 
-            var entryPoint = Path.GetFileNameWithoutExtension(project.ProjectFile.Name);
+            var entryPoint = project.AssemblyName;
```
rynowak, line 45: <3

rynowak: This is still kind a heuristic, but it's definitely a better heuristic than the project name. I think if someone where in the specialized case of using a different filename than the assembly name, we would probably say "don't do that" :)

It's a little tricky to get the actual filename that MSBuild is going to output IIRC, because it's an itemgroup and it's only created after running a bunch of targets.

davidfowl: Right, it's the target path right?

rynowak:
I'm the worlds worst code-reviewer. We have the same bug in the publishing code path. I'll fix it.

rynowak:
Or I'm not. I guess on saturday when you compare someone changes in PR to the files you have locally, they are sometimes not the same.

rynowak:
I was thinking of `TransformProjectsIntoContainers` - but it already does the right thing. Let the record of my confusion exist throughout the ages as a monument to hubris.

src/Microsoft.Tye.Core/DockerfileGenerator.cs
rynowak: I think `TargetName` is actually what you want: https://github.com/dotnet/aspnetcore-tooling/blob/master/src/Razor/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets#L204

That's what we use to generate the name of the views assembly in razor.

rynowak: but I think it earlier SDKs it was set dynamically by CoreCompile - I'm not sure when that changed.

PR #409, opened 2020-04-25
Fix restore.sh

rynowak:
This script had some wrong arguments (bad copy-paste from a more
complex repo). The restore.cmd script already had this fixed.


rynowak:
Bypassing CI because this isn't used in the CI 😆

PR #414, opened 2020-04-27
Add concept of dependencies per service

jkotalik:
Removes environment variable injection for services that don't explicitly depend on a service.


jkotalik:
Merging cause I got a spicy follow up PR on top of this <3

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -155,6 +160,11 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
                         throw new CommandException("Unable to determine service type.");
                     }
 
+                    service.Dependencies.AddRange(parentDependencies);
+                    parentDependencies.Add(service.Name);
```
rynowak, line 164: Seeing these two things right next to each other (inside a loop over services) makes me assume that this has an order dependency.  Maybe some examples in the comments would help? What does parent dependencies represent?

src/Microsoft.Tye.Hosting/Model/Application.cs
```diff
@@ -132,8 +132,10 @@ private List<EffectiveBinding> ComputeBindings(Service service, string defaultHo
         {
             var bindings = new List<EffectiveBinding>();
 
-            foreach (var s in Services.Values)
+            foreach (var serv in service.Description.Dependencies)
```
rynowak, line 135: 😦

test/E2ETest/ApplicationTests.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System.Collections.Generic;
```
rynowak, line 1: missing copyright.

test/E2ETest/ApplicationTests.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.Tye;
+using Test.Infrastructure;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace E2ETest
+{
+    public class ApplicationTests
```
rynowak, line 11: It would be really nice to see some unit tests of the dependency matching logic (which services get which dependnecies). I'm not 100% convinced that it's correct.

PR #419, opened 2020-04-29
Add host argument

alexfdezsauco:
Add host argument to tye to allow optionally expose the tye host and services


davidfowl:
I wonder if it makes more sense to specify the whole url instead of the pieces?

alexfdezsauco:
Not sure. This `tye run --port  PORT -host IP_ADDRESS` make sense for me.

alexfdezsauco:
Looks like something breaks the network tests.

src/tye/ApplicationBuilderExtensions.cs
```diff
@@ -125,6 +125,7 @@ public static Application ToHostingApplication(this ApplicationBuilder applicati
                 services.Add(service.Name, new Service(description));
             }
 
+
```
jkotalik, line 128: nit: remove newline.

alexfdezsauco:
Merged with master but some tests still fail on Linux

jkotalik:
I reran the test and am still seeing them fail. Do you think there could be something that is causing regressions from this PR?

Other PRs seem to be passing.

alexfdezsauco:
Let me check, I think this PR doesn't break the build. This looks that only fail on linux. But I will check again.

alexfdezsauco:
Looks like an issue in the front end project. 

image

The backend pass but the frontend raise an internal error.

alexfdezsauco:
Not sure about the last modification. Can you review the implications?

src/tye/Program.RunCommand.cs
```diff
@@ -113,8 +119,8 @@ private static Command CreateRunCommand()
                 output.WriteInfoLine("Launching Tye Host...");
                 output.WriteInfoLine(string.Empty);
 
-                await using var host = new TyeHost(application.ToHostingApplication(), options);
-                await host.RunAsync();
+                await using var tyeHost = new TyeHost(application.ToHostingApplication(args.Host), options);
```
jkotalik, line 122: I'm confused why we need to pass in the host here, shouldn't it be passed to the ApplicationFactory and set there?

src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -182,6 +183,17 @@ private void ConfigureApplication(WebApplication app)
             app.MapFallbackToPage("/_Host");
         }
 
+        private string ComputeHost(WebApplication app)
+        {
+            var host = app.Configuration["host"];
+            if (string.IsNullOrEmpty(host))
+            {
+                host = "127.0.0.1";
```
jkotalik, line 191:
```suggestion
                return "127.0.0.1";
```

src/Microsoft.Tye.Hosting/Model/Application.cs
```diff
@@ -139,7 +139,8 @@ private List<EffectiveBinding> ComputeBindings(Service service, string defaultHo
                 foreach (var b in s.Description.Bindings)
                 {
                     var protocol = b.Protocol;
-                    var host = b.Host ?? (service.Description.RunInfo is DockerRunInfo ? s.Description.Name : defaultHost);
+                    // Just ignored the binding host.
+                    var host = service.Description.RunInfo is DockerRunInfo ? s.Description.Name : defaultHost;
```
jkotalik, line 143: Yeah this seems odd, when running locally, b.Host contains the parameter that is passed in for --host. Shouldn't that be used here?

alexfdezsauco: Yes, but doesn't work with that value.

jkotalik: So is the host provided here supposed to just be for the dashboard or all backend services? I think that's where I'm personally confused.

alexfdezsauco: The value assigned to the `host` is used for "something" with the all backed services. The usage of value in `b.Host` creates a wrong configuration and can reach only one. That was the reason for tests failures.  

Ignoring the `b.Host` for the service conguration setup all "correctly" in local mode.

src/tye/Program.RunCommand.cs
alexfdezsauco: Fixed

src/Microsoft.Tye.Hosting/TyeHost.cs
alexfdezsauco: Fixed

alexfdezsauco:
Merged with master again and fix.

alexfdezsauco:
@davidfowl @jkotalik please take a look into this. I have to ignore the binding host to make this pass the tests in all platforms.

alexfdezsauco:
Hi @davidfowl, @rynowak, @jkotalik. Any chance to merge this into? 

If I run tye in this way

 `tye run -host 192.168.43.152`

allows me access to the services using the specified ip address.

image

I need this.  

This is very useful for demos or production (why not?).

alexfdezsauco:
Merged from master

alexfdezsauco:
Think this is solve with this https://github.com/dotnet/tye/blob/master/docs/recipes/ingress.md

alexfdezsauco:
This is not exactly solved by ingress. So, I create a new pull request https://github.com/dotnet/tye/pull/708

PR #420, opened 2020-04-29
Add status report generator

rynowak:
Thanks @anurse!

### DIFF
--- a/eng/Generate-StatusReport.ps1
+++ b/eng/Generate-StatusReport.ps1
@@ -1,0 +1,173 @@
+<#
+.SYNOPSIS
+    Generates a template status report for the repository
+.PARAMETER GitHubToken
+    A GitHub Bearer Token. The easiest way to get one of these is to go to https://github.com/settings/tokens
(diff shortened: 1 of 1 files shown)


rynowak:
Bypassing CI - no product or test change

eng/Generate-StatusReport.ps1
```diff
@@ -0,0 +1,173 @@
...
+    "Content-Type" = "application/json";
+}
+
+$Response = Invoke-RestMethod -Method "POST" -Headers $Headers -Body $RequestBody $GraphQlEndpoint
+
+$Cursor = $null;
+$ContributorPRs = @();
+do {
+    if ($Cursor) {
+        throw "Multi-page results not yet supported!"
```
anurse, line 130: FYI, I just got lazy here and figured I wouldn't be running the script over a period that contained 100 PRs. It's possible it could happen and we'd need to do paging. It's [not terribly hard though](https://graphql.github.io/learn/pagination/)

PR #424, opened 2020-04-29
Allows cloning dependencies

jkotalik:
Follow up to #414. Now adds ability to have a URL to clone from. Also addresses feedback after PR.

### DIFF
--- a/src/Microsoft.Tye.Core/DirectoryExtensions.cs
+++ b/src/Microsoft.Tye.Core/DirectoryExtensions.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Tye
     {
         // Calling Directory.Delete causes an exception for .git folders:
         //     System.UnauthorizedAccessException : Access to the path '17a475ecca365c678e907bd4c73e4c65b341c6' is denied.
-        public static void DeleteDirectory( string d)
+        public static void DeleteDirectory(string d)
         {
             foreach (var sub in Directory.EnumerateDirectories(d))
             {
--- a/src/Microsoft.Tye.Core/DirectoryExtensions.cs
+++ b/src/Microsoft.Tye.Core/DirectoryExtensions.cs
@@ -1,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+
+namespace Microsoft.Tye
+{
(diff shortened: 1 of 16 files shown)


davidfowl:
Few questions:
- How does it fail if you don't have git installed?
- How does it fail if you use private repo that requires a password.
- If there's no tye.yaml in the root, does it fail after clone?

jkotalik:
> How does it fail if you don't have git installed?

I'll add a check for that. It would fail calling ProcessUtils.RunAsync.

> How does it fail if you use private repo that requires a password.

Do you think capturing the output from git clone would be good enough here for this scenario? I feel like it would be hard to know exactly what

> If there's no tye.yaml in the root, does it fail after clone?

No, it follows the same conventions for slns, csprojs, etc if a tye.yaml isn't present in the root of the project.

jkotalik:
I'm also thinking we may need to specify a path to the tye.yaml inside of the repository (ex if the tye.yaml isn't present in the root).

davidfowl:
> I'll add a check for that. It would fail calling ProcessUtils.RunAsync.

We do this today for Docker, we check if it's installed and running. I'd suggest following the same pattern.

> Do you think capturing the output from git clone would be good enough here for this scenario? I feel like it would be hard to know exactly what

We should at least understand what it does. Maybe it'll prompt and because we aren't redirecting stdin it'll fail.

jkotalik:
Errors are outputted correctly, example of repository not found.
```
Failed to clone repository https://github.com/jkotalik/TyeMultiRepoVotingz with exit code 128.
    Cloning into 'C:\Users\jukotali\AppData\Local\Temp\1nwk0keb.zrk\.tye\deps\vote'...
    remote: Repository not found.
    fatal: repository 'https://github.com/jkotalik/TyeMultiRepoVotingz/' not found
```

src/Microsoft.Tye.Core/DirectoryExtensions.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+
+namespace Microsoft.Tye
+{
+    public static class DirectoryExtensions
```
rynowak, line 9: :(

jkotalik: What name/stuff would you prefer?

Issue #351, opened 2020-04-10
tye deploy does not work if KUBECONFIG has more than one entry


giggio:
### Describe the bug

When the env var `KUBECONFIG` is set to more than one file, `tye deploy` fails.

On Windows pwsh, simply set it like this:

````pwsh
$env:KUBECONFIG="$env:USERPROFILE\.kube\config;$env:USERPROFILE\.kube\other-config"
````

This is the log after running `tye deploy --interactive`, using the second [step of tutorial](https://github.com/dotnet/tye/blob/master/docs/tutorials/hello-tye/02_add_redis.md#deploying-redis).

````
Loading Application Details...
Processing Service 'frontend'...
    Compiling Services...
    Publishing Project...
    Building Docker Image...
        Created Docker Image: 'giggio/frontend:1.0.0'
    Pushing Docker Image...
        Pushed docker image: 'giggio/frontend:1.0.0'
    Validating Secrets...
An unhandled exception has occurred, how unseemly:
k8s.Exceptions.KubeConfigException: kubeconfig file not found at C:\Users\giovanni\.kube\config;C:\Users\giovanni\.kube\other-config;
   at k8s.KubernetesClientConfiguration.LoadKubeConfigAsync(FileInfo kubeconfig, Boolean useRelativePaths)
   at k8s.KubernetesClientConfiguration.LoadKubeConfig(FileInfo kubeconfig, Boolean useRelativePaths)
   at k8s.KubernetesClientConfiguration.BuildConfigFromConfigFile(FileInfo kubeconfig, String currentContext, String masterUrl, Boolean useRelativePaths)
   at k8s.KubernetesClientConfiguration.BuildConfigFromConfigFile(String kubeconfigPath, String currentContext, String masterUrl, Boolean useRelativePaths)
   at k8s.KubernetesClientConfiguration.BuildDefaultConfig()
   at Microsoft.Tye.ValidateSecretStep.ExecuteAsync(OutputContext output, ApplicationBuilder application, ServiceBuilder service) in /_/src/Microsoft.Tye.Core/ValidateSecr
etStep.cs:line 54
   at Microsoft.Tye.ServiceExecutor.ExecuteAsync(ServiceBuilder service) in /_/src/Microsoft.Tye.Core/ServiceExecutor.cs:line 50
   at Microsoft.Tye.Program.ExecuteDeployAsync(OutputContext output, ApplicationBuilder application, String environment, Boolean interactive, Boolean force) in /_/src/tye/
[...]

PR #429, opened 2020-04-30
Update version of Kubernetes Client

jkotalik:
Fixes https://github.com/dotnet/tye/issues/351.

I verified locally that having two kubeconfig files listed in the environment variable does not throw and returns an expected result.

### DIFF
--- a/src/Microsoft.Tye.Core/ValidateIngressStep.cs
+++ b/src/Microsoft.Tye.Core/ValidateIngressStep.cs
@@ -56,14 +56,6 @@ namespace Microsoft.Tye
             output.WriteDebugLine($"Validating ingress class '{ingressClass}'.");
             var config = KubernetesClientConfiguration.BuildDefaultConfig();
 
-            // Workaround for https://github.com/kubernetes-client/csharp/issues/372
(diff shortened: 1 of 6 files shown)


jkotalik:
@mkArtakMSFT I think we need to regenerate third party notices as the latest kubeclient package took a dependency on another 3rd party package.

rynowak:
There's another workaround we can undo here as well. Look at all the places we create a Kubernetes client

src/Microsoft.Tye.Core/ValidateIngressStep.cs
```diff
@@ -56,14 +56,6 @@ public override async Task ExecuteAsync(OutputContext output, ApplicationBuilder
             output.WriteDebugLine($"Validating ingress class '{ingressClass}'.");
             var config = KubernetesClientConfiguration.BuildDefaultConfig();
 
-            // Workaround for https://github.com/kubernetes-client/csharp/issues/372
-            var store = await KubernetesClientConfiguration.LoadKubeConfigAsync();
```
jkotalik, line 60: @rynowak I noticed when verifying this locally that config.Namespace is null when using the default namespace. Do you want to still have config.Namespace be "default"?

Issue #19, opened 2020-03-01
Add heathcheck support


davidfowl:
We should add health checking support to the host to know if a service is running properly.

areller:
@davidfowl I've experimented with it a bit and I have something in this branch https://github.com/areller/tye/tree/replica-health

Basically, you can define a `health` field in the service description that can be configured with these arguments

* endpoint (The http endpoint that will be checked every interval for HTTP 2xx)
* testInterval (The interval of the testing in seconds, default: 1)
* gracePeriod (After a replica became sick, how many seconds to wait until we kill the replica, default: 5)
* bootPeriod (After a replica has started, how many seconds to wait for it to become healthy before killing it, default: 5)

Here is an example yaml

```
# tye application configuration file
# read all about it at https://github.com/dotnet/tye
name: single-project-health
services:
- name: test-project
  project: test-project/test-project.csproj
  bindings:
    - port: 8081
  health:
    endpoint: /health
```

(Replicas of services without the `health` field will automatically become healthy upon creation)

The heart of this feature is in this file (and line)
https://github.com/areller/tye/blob/94e123bf9b0f3c01269cf5881aacf539b2517a85/src/Microsoft.Tye.Hosting/Hospital.cs#L135

This can be improved by

* Adding another state, readiness (a replica can be healthy but not ready, in that case, we don't kill it)
* Displaying health status/statistics in the dashboard

Once you have healthcheck (and maybe readiness) you can start talking about more advanced concepts like dependencies.

davidfowl:
@areller do you still want to work on this?

areller:
Sure! What do you think about the initial version (with the `Hospital.cs` file)? Is it a good direction?

jkotalik:
This is done!

PR #432, opened 2020-04-30
adding health checks 

areller:
adding health checks according to this https://github.com/dotnet/tye/issues/19

## Changes
1. Replicas can be cancelled independent of the service
2. There are 2 new configurations for a service: `liveness`, `readiness`. each can hold these settings

```
http:
  path: /health
  port: 80 # default: pick the first binding of the service
  protocol: http # default: none
  headers:
    - name: A
       value: B
initialDelay: 5 # default: 0
period: 1 # default: 1
timeout: 1 # default: 1
successThreshold: 1 # default: 1
failureThreshold: 3 # default: 3
```
3. There are 2 new replica states: `Healthy`, `Ready` (`Ready` = passed both `liveness` and `readiness` probing. `Healthy` = passed `liveness` but not `readiness`).  
If neither `liveness` nor `readiness` were provided, replica will be upgraded to `Ready` upon creation. If only `liveness` if provided, replica will be upgraded to `Ready` upon passing the `liveness` probe. if only `readiness` is provided, replica will be upgraded to `Healthy` upon creation.
4. There is a `ReplicaMonitor` that reads the `liveness` and `readiness` probe configuration, and kills/changes the state of the replica according to probe results
5. Proxy and Ingress only forward to `Ready` replicas
6. The generated Kubernetes manifest contains the `liveness` and `readiness` configuration.


src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -219,210 +219,215 @@ private async Task StartContainerAsync(Application application, Service service,
 
             async Task RunDockerContainer(IEnumerable<(int ExternalPort, int Port, int? ContainerPort, string? Protocol)> ports)
             {
-                var hasPorts = ports.Any();
+                while (!dockerInfo.StoppingTokenSource.IsCancellationRequested)
```
davidfowl, line 222: Why is there a loop here?

areller: When I implement the monitor, it might decide to kill a container because it's unhealthy. In that case I'd want the `DockerRunner` to recreate the replica.
Another way to do it, is to kill the container by calling `docker restart`, and then there is no need for that loop. what do you think?

areller: The argument behind the former, is that it's consistent with how I'll do it in `ProcessRunner`

davidfowl: The docker runner already restarts the replica albiet with the same name.

areller: @davidfowl I think that the problem with doing `docker restart` is that the health monitor won't be able to tell that the container was restarted (because there won't be `Stopped` and `Started` events)

areller:
@davidfowl there are two major things left

1. generating k8s yamls with liveness and readiness
2. I don't forward traffic to non ready replicas, unless there is a single replica. proxying with a single replica causes some tests that use external docker images (e.g. redis) to fail, and I need to check why.

Maybe it's best to start reviewing this PR and I can submit the rest later on this PR or another PR?

test/E2ETest/testassets/projects/health-checks/tye-all.yaml
```diff
@@ -0,0 +1,36 @@
+# tye application configuration file
+# read all about it at https://github.com/dotnet/tye
+name: health-checks
+services:
+  - name: all
+    project: api/api.csproj
+    replicas: 3
+    bindings:
+      - port: 8004
+    liveness:
```
davidfowl, line 10: These 2 probes are bigger than the entire yaml, that makes me 😢

areller: @davidfowl yeah... I thought it's best to go with something flexible like in k8s, but you can get a way with just providing a `path`, the rest have default values

areller:
I'll probably add the k8s manifest generation tomorrow.

jkotalik:
@areller this change is amazing. Got my coffee ☕  and will be reviewing now.

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
```diff
@@ -452,6 +462,64 @@ internal static class KubernetesManifestGenerator
...
+            {
+                var builderHttp = builder.Http!;
+                var httpGet = new YamlMappingNode();
+
+                probe.Add("httpGet", httpGet);
+                httpGet.Add("path", builderHttp.Path);
+
+                if (builderHttp.Protocol != null)
+                {
+                    httpGet.Add("scheme", builderHttp.Protocol!.ToUpper());
```
jkotalik, line 480: I think you don't need `!` here, other similar cases in this file as well.

```suggestion
                    httpGet.Add("scheme", builderHttp.Protocol.ToUpper());
```

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
```diff
@@ -452,6 +462,64 @@ internal static class KubernetesManifestGenerator
...
+                }
+
+                if (builderHttp.Port != null)
+                {
+                    httpGet.Add("port", builderHttp.Port!.ToString()!);
+                }
+                else
+                {
+                    // If port is not given, we pull default port
+                    var binding = service.Bindings.First(b => builderHttp.Protocol is null || b.Protocol == builderHttp.Protocol);
```
jkotalik, line 490: It does feel a bit awkward to have health checks per service rather than per endpoint, but it seems to be what k8s does. Or maybe we require a port?

src/Microsoft.Tye.Core/ProjectServiceBuilder.cs
```diff
@@ -54,5 +54,8 @@ public ProjectServiceBuilder(string name, FileInfo projectFile)
         public Dictionary<string, string> BuildProperties { get; } = new Dictionary<string, string>();
 
         public List<SidecarBuilder> Sidecars { get; } = new List<SidecarBuilder>();
+
+        public ProbeBuilder? Liveness { get; set; }
+        public ProbeBuilder? Readiness { get; set; }
```
jkotalik, line 59:
```suggestion

        public ProbeBuilder? Readiness { get; set; }
```

src/Microsoft.Tye.Hosting/DockerRunner.cs
jkotalik: From what I can tell with this loop, you would call `docker rm` and other commands I don't think you want to run: https://github.com/dotnet/tye/blob/65c7ae0d80d0b3749127839732cf7d1ac9926b79/src/Microsoft.Tye.Hosting/DockerRunner.cs#L429

The restart loop for rerunning docker is here: https://github.com/dotnet/tye/blob/65c7ae0d80d0b3749127839732cf7d1ac9926b79/src/Microsoft.Tye.Hosting/DockerRunner.cs#L382. Can you modify this loop instead?

src/Microsoft.Tye.Hosting/HttpProxyService.cs
```diff
@@ -117,23 +122,30 @@ public async Task StartAsync(Application application)
...
 
                         _logger.LogInformation("Service {ServiceName} is using {Urls}", targetServiceDescription.Name, string.Join(",", uris.Select(u => u.ToString())));
 
                         // The only load balancing strategy here is round robin
                         long count = 0;
-                        RequestDelegate del = context =>
+                        RequestDelegate del = async context =>
                         {
                             var next = (int)(Interlocked.Increment(ref count) % uris.Count);
+                            for (int i = 0; !_readyPorts.ContainsKey(uris[next].Port) && i < uris.Count; i++, next = (int)(Interlocked.Increment(ref count) % uris.Count)) ;
```
jkotalik, line 135: Lot of stuff going on in this for loop, can you break it up a bit and add a comment about what it is doing?

My best guess is loop through all ready ports until we find one that is ready.

src/Microsoft.Tye.Hosting/Model/V1/V1ReplicaStatus.cs
```diff
@@ -18,5 +18,7 @@ public class V1ReplicaStatus
         public int? ExitCode { get; set; }
         public int? Pid { get; set; }
         public IDictionary<string, string>? Environment { get; set; }
+
```
jkotalik, line 21:
```suggestion
```

src/Microsoft.Tye.Hosting/PortAssigner.cs
```diff
@@ -50,7 +50,7 @@ static int GetNextPort()
                         binding.Port = GetNextPort();
                     }
 
-                    if (service.Description.Replicas == 1)
+                    if ((service.Description.Readiness is null || service.Description.RunInfo is IngressRunInfo) && service.Description.Replicas == 1)
```
jkotalik, line 53: Why add the check for `service.Description.RunInfo is IngressRunInfo`

src/Microsoft.Tye.Hosting/ReplicaMonitor.cs
```diff
@@ -0,0 +1,448 @@
...
+                {
+                    StartLivenessProbe(serviceDesc.Liveness);
+                    StartReadinessProbe(serviceDesc.Readiness);
+                }
+            }
+
+            private void StartLivenessProbe(Probe probe, ReplicaState moveToOnSuccess = ReplicaState.Healthy)
+            {
+                // currently only HTTP is available
+                if (probe.Http is null)
```
jkotalik, line 129:
```suggestion
                if (probe.Http == null)
```

src/Microsoft.Tye.Hosting/ReplicaMonitor.cs
```diff
@@ -0,0 +1,448 @@
...
+                {
+                    _logger.LogWarning("cannot start probing replica {name} because probe configuration is not set", _replica.Name);
```
jkotalik, line 131:
```suggestion
                    _logger.LogWarning("Cannot start probing replica {name} because probe configuration is not set", _replica.Name);
```

src/Microsoft.Tye.Hosting/ReplicaMonitor.cs
jkotalik: And for the rest of the file

src/Microsoft.Tye.Hosting/ReplicaMonitor.cs
jkotalik: And for the rest of the file.

src/Microsoft.Tye.Hosting/ReplicaMonitor.cs
```diff
@@ -0,0 +1,448 @@
...
+                            break;
+                    }
+                });
+
+                _readinessProber.Start();
+            }
+
+            private void MoveToHealthy(ReplicaState from)
+            {
+                _logger.LogInformation("replica {name} is moving to an healthy state", _replica.Name);
```
jkotalik, line 233: These logs should probably be at debug level.

src/Microsoft.Tye.Hosting/ReplicaMonitor.cs
```diff
@@ -0,0 +1,448 @@
...
+                }
+                else
+                {
+                    _logger.LogWarning(message);
+                }
+            }
+
+            public override void Start()
+            {
+                Func<ReplicaBinding, bool> bindingClosure = (_httpProberSettings.Port.HasValue, _httpProberSettings.Protocol != null) switch
```
jkotalik, line 415: This probably deserves a comment as well.

src/Microsoft.Tye.Hosting/PortAssigner.cs
areller: because I forgot to remove when added `service.Description.Readiness is null` :)

src/Microsoft.Tye.Hosting/HttpProxyService.cs
areller: that's right. I'm looping through all the ports (starting from the current selected port, `# of ports` times) to find one that is ready. I'll try to simplify and add comment

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
areller: requiring a port will indeed make it more compatible with k8s, but it will make the `tye.yaml` more verbose and less maintainable in my opinion. one thing that k8s does to help with maintainability is to allow to specify a variable as a port. e.g.

```
port: liveness-port
```

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
areller: yep. I removed it and it didn't throw any warnings. probably because it's inside a block that runs if `builderHttp.Protocol != null`.
It's weird because at times the compiler will throw a warning when not using the `!` operator, even though the expression is in a context where it's guaranteed to not be null...

src/Microsoft.Tye.Hosting/DockerRunner.cs
areller: @jkotalik yeah, makes sense. I'll try doing it that way.

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
areller: found an example: if you go to `ReplicaMonitor.cs`:113 and remove the `!` after `serviceDesc.Readiness`, the compiler will throw a warning, even though there is no way for that variable to be null at that point.

src/Microsoft.Tye.Hosting/DockerRunner.cs
areller: @jkotalik changed the `DockerRunner` to do `docker restart` when `StoppingTokenSource` is canceled.

src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
```diff
@@ -187,6 +201,165 @@ private static void HandleServiceVolumes(YamlSequenceNode yamlSequenceNode, List
             }
         }
 
+        private static void HandleServiceProbe(YamlMappingNode yamlMappingNode, ConfigProbe probe)
+        {
+            foreach (var child in yamlMappingNode.Children)
```
jkotalik, line 206: Can we add some tests for these parsing wise? Much faster for checking these kinds of conditions like making sure initialDelay can't be negative.

src/Microsoft.Tye.Hosting/ReplicaMonitor.cs
```diff
@@ -0,0 +1,455 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Tye.Hosting.Model;
+
+namespace Microsoft.Tye.Hosting
+{
+    public class ReplicaMonitor : IApplicationProcessor
+    {
+        private ILogger _logger;
```
jkotalik, line 19: @davidfowl to review this on top of me.

src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
```diff
@@ -187,6 +201,165 @@ private static void HandleServiceVolumes(YamlSequenceNode yamlSequenceNode, List
             }
         }
 
+        private static void HandleServiceProbe(YamlMappingNode yamlMappingNode, ConfigProbe probe)
```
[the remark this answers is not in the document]
areller: Yeah. That would probably make more sense but it will break compatibility with the rest of the methods. What do you think?

src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
areller: I'll try to come up with something. You want it as part of this PR? or break it up in a different one?

jkotalik: I think following the structure here: https://github.com/dotnet/tye/blob/master/test/UnitTests/TyeDeserializationTests.cs#L25.

Mostly just parsing/ validation of tye.yaml to make sure we cover these edge cases correctly.

No preference whether we add them in this PR now or follow up, I'd just like them in general 😄

jkotalik:
I'm going to go ahead and merge this, let's do some follow up on top of it 😄 .

Again, thanks @areller for all the work you have done here

areller:
@jkotalik thank you! Will create a PR in the following days

PR #435, opened 2020-05-02
Add distribute tracing recipe and extension

rynowak:
Adding a zipkin extension for local run (deployment is next).

This will automatically inject zipkin as a service to the application for local run. This also takes control of the `--dtrace` argument if not specified at the command line. 

Refactored how options get passed to the host, so that extensions can do stuff like this 👍 

```yaml
name: frontend-backend

# THIS IS NEW
extensions:
- name: zipkin
# endregion cool and good

services:
- name: backend
  project: backend/backend.csproj
- name: frontend
  project: frontend/frontend.csproj
```

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -15,6 +15,7 @@
 | Topic | Description|
 |-------|------------|
 |**[Using Ingress](recipes/ingress.md)** | Using `tye` with an ingress for serving public traffic.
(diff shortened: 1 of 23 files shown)


rynowak:
> Like, are you going to do prometheus next?

(deployment is next)

PR #437, opened 2020-05-03
Add diagnostics sidecar and support for zipkin in deployed apps

rynowak:
Adds a diagnostics sidecar and the ability to push distributed traces to zipkin for deployed applications.

The sidecar uses the same diagnostics sink code we use for local run, so we can easily add support for providers to each one. The one delta is that the diagnostics code can use service discovery, so in local run we have to provide diagnostics with a URI/key/etc, and in deployed applications that can come from service discovery.

From this point it will be really easy for us to add anything push-based for diagnostics. The main blockers have been worked out 👍

### DIFF
--- a/docs/recipes/zipkin.yaml
+++ b/docs/recipes/zipkin.yaml
@@ -1,0 +1,40 @@
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: zipkin
+  labels:
+    app.kubernetes.io/name: 'zipkin'
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: zipkin
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: 'zipkin'
+    spec:
+      containers:
+      - name: zipkin
+        image: openzipkin/zipkin
+        imagePullPolicy: Always
+        ports:
+        - containerPort: 9411
+...
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: zipkin
+  labels:
+    app.kubernetes.io/name: 'zipkin'
+spec:
+  selector:
+    app.kubernetes.io/name: zipkin
+  type: ClusterIP
+  ports:
+  - name: http
+    protocol: TCP
+    port: 9411
+    targetPort: 9411
+...
--- a/src/Microsoft.Tye.Core/CombineStep.cs
+++ b/src/Microsoft.Tye.Core/CombineStep.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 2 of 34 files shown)


docs/recipes/zipkin.yaml
```diff
@@ -0,0 +1,40 @@
...
+    app.kubernetes.io/name: 'zipkin'
+spec:
+  selector:
+    app.kubernetes.io/name: zipkin
+  type: ClusterIP
+  ports:
+  - name: http
+    protocol: TCP
+    port: 9411
+    targetPort: 9411
```
rynowak, line 39: Note, I didn't take the step of actually deploying zipkin for you, because it requires databases.

It really seems like there isn't a *standard* or *official* route for zipkin. We might want to pick Jaeger as the recommended dtrace package. They seem to have lot more options and documentation for deployment.

src/Microsoft.Tye.Core/CombineStep.cs
```diff
@@ -49,12 +50,22 @@ public override Task ExecuteAsync(OutputContext output, ApplicationBuilder appli
...
+            foreach (var sidecar in project.Sidecars)
+            {
+                sidecar.Outputs.Add(ComputeBindings(application, sidecar.Dependencies));
+            }
+
+            return Task.CompletedTask;
+        }
 
-            foreach (var o in service.Dependencies)
+        private ComputedBindings ComputeBindings(ApplicationBuilder application, IEnumerable<string> dependencies)
```
rynowak, line 63: Refactored to a method so I could call it from 2 places.

src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
```diff
@@ -274,6 +274,36 @@ internal static class KubernetesManifestGenerator
...
+
+                foreach (var sidecar in project.Sidecars)
+                {
+                    sidecar.EnvironmentVariables.Add(new EnvironmentVariableBuilder("TMPDIR")
+                    {
+                        Value = "/var/tye/diagnostics",
+                    });
+                }
+            }
+
```
rynowak, line 306: All of these pieces took a while to figure out. Basically there's a few things in the way if you want to use eventpipe in a sidecar:

- You need eventpipe to find the other process
- You need a way to find out which process you want (alternatively handle all of them)
  - event pipe just gives you PIDs and that's all
  - I took the approach of using `ps -p <pid> -o command=` to output the args of each process
- If you share process space (needed to enumerate processes in another container) then the target process won't be pid=1.

So basically it gets much more complicated once the number of processes is > 2 (you + one other .net process). I wanted to be resilient for the case where someone did a `dotnet run` or something like that.

Open to ideas about how to make this better.

src/Microsoft.Tye.Extensions/Zipkin/ZipkinExtension.cs
```diff
@@ -14,48 +14,73 @@ internal sealed class ZipkinExtension : Extension
...
-                {
-                    context.Output.WriteDebugLine("zipkin service already configured. Skipping...");
-                }
-                else
+                context.Output.WriteDebugLine("zipkin service already configured. Skipping...");
+            }
+            else
+            {
+                context.Output.WriteDebugLine("Injecting zipkin service...");
+                var service = new ContainerServiceBuilder("zipkin", "openzipkin/zipkin")
```
rynowak, line 24: Injecting zipkin as a service is useful for deployment, because then it can show up as a binding, and tye already knows how to prompt for it.

Basically we can model diagnostics sinks as *external* services and a lot of things *just work*.

src/Microsoft.Tye.Extensions/Zipkin/ZipkinExtension.cs
```diff
@@ -14,48 +14,73 @@ internal sealed class ZipkinExtension : Extension
...
 
-                        // make zipkin available as a dependency of everything.
-                        if (!s.Dependencies.Contains(service.Name))
-                        {
-                            s.Dependencies.Add(service.Name);
-                        }
+                    // make zipkin available as a dependency of everything.
+                    if (!s.Dependencies.Contains(service.Name))
+                    {
+                        s.Dependencies.Add(service.Name);
```
rynowak, line 48: @jkotalik - this worked out nicely 👌 - this way the sidecar only has a binding for zipkin.

src/Microsoft.Tye.Extensions/Zipkin/ZipkinExtension.cs
```diff
@@ -14,48 +14,73 @@ internal sealed class ZipkinExtension : Extension
...
+                }
+            }
+            else if (context.Operation == ExtensionContext.OperationKind.Deploy)
+            {
+                foreach (var project in context.Application.Services.OfType<ProjectServiceBuilder>())
+                {
+                    // Bring your rain boots.
+                    project.RelocateDiagnosticsDomainSockets = true;
+
+                    var sidecar = new SidecarBuilder("tye-diag-agent", "rynowak/tye-diag-agent", "0.1")
```
rynowak, line 68: LOL don't @ me

davidfowl:
We really shouldn’t point to your random docker hub image

rynowak:
Yeah, I'll log an issue :)

PR #439, opened 2020-05-03
Add recipe for Elastic stack (ELK)

rynowak:


src/Microsoft.Tye.Hosting.Diagnostics/LoggingSink.cs
```diff
@@ -26,7 +26,7 @@ public LoggingSink(Microsoft.Extensions.Logging.ILogger logger, DiagnosticsProvi
 
         public IDisposable Attach(EventPipeEventSource source, ReplicaInfo replicaInfo)
         {
-            using var loggerFactory = LoggerFactory.Create(builder => ConfigureLogging(replicaInfo.Service, replicaInfo.Replica, builder));
+            var loggerFactory = LoggerFactory.Create(builder => ConfigureLogging(replicaInfo.Service, replicaInfo.Replica, builder));
```
rynowak, line 29: This was a bug. We'd dispose the logger factory when this method exits, but we want to dispose it when the event pipe is closed.

src/Microsoft.Tye.Extensions/Elastic/ElasticStackExtension.cs
```diff
@@ -0,0 +1,87 @@
...
+                            ContainerPort = 9200,
+                            Protocol = "http",
+                        },
+                    },
+                };
+                context.Application.Services.Add(elastic);
+
+                if (config.Data.TryGetValue("logPath", out var obj) && 
+                    obj is string logPath &&
+                    !string.IsNullOrEmpty(logPath))
```
jkotalik, line 50: Should we create the logPath folder if it doesn't exist, or will something downstream create the folder?

rynowak: docker will create it, I tried

PR #442, opened 2020-05-05
Enable log aggregation in deployed services

rynowak:


samples/.gitignore
```diff
@@ -0,0 +1 @@
+.logs/
```
rynowak, line 1: Adding this for sanity of everyone when testing the logging functionality. You **don't** want to check in the files elasticsearch creates 👍

src/tye-diag-agent/Program.cs
```diff
@@ -30,17 +30,17 @@ public static void Main(string[] args)
...
                         {
-                            if (!DiagnosticsProvider.TryParse(child.Value, DiagnosticsProvider.ProviderKind.Unknown, out var provider))
-                            {
-                                throw new InvalidOperationException("Could not parse diagnostics provider: " + child.Value);
-                            }
-
                             // The value used to connect to the service should be accessible with Tye service discovery.
-                            if (provider.Value is string && provider.Value.StartsWith("service:"))
+                            DiagnosticsProvider provider;
+                            if (child.Value is string && child.Value.StartsWith("service:"))
```
rynowak, line 35: change here is that we were modeling providers in command line args as an array:

```sh
--provider:0=zipkin=...
--provider:1=elastic=...
```

Now they are an object:

```sh
--provider:zipkin=...
--provider:elastic=...
```

Much easier to generate and less arcane.

src/Microsoft.Tye.Extensions/Elastic/ElasticStackExtension.cs
```diff
@@ -29,14 +29,13 @@ public override Task ProcessAsync(ExtensionContext context, ExtensionConfigurati
                     {
                         new BindingBuilder()
                         {
-                            Name = "kibbana",
+                            Name = "kibana",
```
jkotalik, line 32: nice

Issue #240, opened 2020-03-27
Docker Debian Based Images assumed with TFM Pattern


spboyer:
https://github.com/dotnet/tye/blob/2a0ffaf88cc135a3fcdd41f7ae83eb3c1dd69af9/src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs#L88

his pattern would assume that all development, if using Linux containers, would be done on Debian based containers given the TFM base image is always such. How would we specify an alpine or other flavor? Should we consider looking leveraging [Container MS Build Properties](https://docs.microsoft.com/en-us/visualstudio/containers/container-msbuild-properties?view=vs-2019) or expanding upon them?

i.e.

```xml
<PropertyGroup>
   <DockerbaseImage>{TFM}-alpine</DockerbaseImage>
</PropertyGroup>
```

rynowak:
Triage: the container tools package requires a dockerfile, which we want to avoid in a lot of cases. It seems like a good idea that we're support some customization via MSBuild or another path that doesn't require a dockerfile 👍

rynowak:
We're open to a PR for this - using the project file to feed the base image to the dockerfile generator for both run and deploy.

We're open to a pr that looks like:

```C#
<PropertyGroup>
   <ContainerBaseImage>...</ContainerBaseImage>
   <ContainerBaseTag>...</ContainerBaseTag>
</PropertyGroup>
```

`ContainerBaseImage` specifies the repository.

`ContainerBaseTag` specifies the tag.

Most users will want to end up changing `ContainerBaseTag` if they are using a supported .NET runtime image, the choice of OS is represented in our images by different tags.

PR #444, opened 2020-05-06
Fixes for docker scenarios

### LINKED ISSUE #267: Support HTTPS in tye run --docker
### Describe the bug
When running projects in a container HTTPS doesn't specify the dev cert and this results in warnings in the browser. We need to export the cert with a password, set the password in user secrets for the application, and mount the cert into the container pointing kestrel at it.

davidfowl:
- Support https properly inside of containers by injecting the dev cert into the right place.
- Only inject .NET specific environment for projects being transformed into containers.
- Use the same base image selection as deployment
- Added backoff to restarts

Fixes #267

### DIFF
--- a/src/Microsoft.Tye.Hosting/Model/DockerRunInfo.cs
+++ b/src/Microsoft.Tye.Hosting/Model/DockerRunInfo.cs
@@ -13,7 +13,6 @@ namespace Microsoft.Tye.Hosting.Model
         {
             Image = image;
             Args = args;
-            IsAspNet = Image.StartsWith("mcr.microsoft.com/dotnet/core/aspnet", StringComparison.OrdinalIgnoreCase);
         }
 
         public bool Private { get; set; }
--- a/src/Microsoft.Tye.Hosting/Model/DockerRunInfo.cs
+++ b/src/Microsoft.Tye.Hosting/Model/DockerRunInfo.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 
 namespace Microsoft.Tye.Hosting.Model
@@ -15,6 +16,7 @@ namespace Microsoft.Tye.Hosting.Model
         }
 
         public bool Private { get; set; }
+        public bool IsAspNet => Image.StartsWith("mcr.microsoft.com/dotnet/core/aspnet", StringComparison.OrdinalIgnoreCase);
 
         public string? NetworkAlias { get; set; }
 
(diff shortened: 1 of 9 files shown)


davidfowl:
Is docker just broken?

src/Microsoft.Tye.Hosting/Model/DockerRunInfo.cs
```diff
@@ -15,6 +16,7 @@ public DockerRunInfo(string image, string? args)
         }
 
         public bool Private { get; set; }
+        public bool IsAspNet => Image.StartsWith("mcr.microsoft.com/dotnet/core/aspnet", StringComparison.OrdinalIgnoreCase);
```
spboyer, line 19: Is this the only way to determine ASPNET? If #240 is supported then this fails as a check allowing devs to create their own base images.

rynowak: > Is this the only way to determine ASPNET?

yeah this seems backwards. We *already know* that this came from an ASP.NET Core project. We throw away that knowledge and use a heuristic that's easily broken - improving that heuristic would take a lot of work - we should just try to propagate the knowledge that it was an ASP.NET Core project originally.

davidfowl: I cheated I can change it though. I was also looking at our backyard tests 😬

src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs
```diff
@@ -75,7 +75,8 @@ private async Task TransformProjectToContainer(Service service, ProjectRunInfo p
             var outputFileName = project.AssemblyName + ".dll";
             var dockerRunInfo = new DockerRunInfo(containerImage, $"dotnet {outputFileName} {project.Args}")
             {
-                WorkingDirectory = "/app"
+                WorkingDirectory = "/app",
+                IsAspNet = project.IsAspNet
```
rynowak, line 79: <3

src/Microsoft.Tye.Hosting/Model/DockerRunInfo.cs
```diff
@@ -13,10 +13,11 @@ public DockerRunInfo(string image, string? args)
         {
             Image = image;
             Args = args;
+            IsAspNet = Image.StartsWith("mcr.microsoft.com/dotnet/core/aspnet", StringComparison.OrdinalIgnoreCase);
```
rynowak, line 16: Is there a point to still having this initializer?  This is the name of the base-image. Someone using a pre-built image as a dependency would have this base-image as a layer (possibly), not as the name of their image. I'm struggling to think of a case where this would be true (other than when we transform a project).

If we want to go deep on lighting up asp.net core functionality for arbitrary images, then a label is probably the right way to do is.

davidfowl: I don't want to add IsAspNet to ContainerServiceBuilder. Here's where I'm using it here https://github.com/dotnet/tye/blob/122d73c3e55135ffdf85f1e52e222b17e733aab6/test/E2ETest/TyeRunTests.cs#L212.

Do you want to support container labels as part of this PR?

Issue #146, opened 2020-03-19
Support for Dockerfile


adiazcan:
### What should we add or change to make your life better?
Thinking on our projects, my team develop backend using dotnet core and the frontend using reactjs. 

If it possible to add a Dockerfile like we can add a dockerimage? 

In this case, Tye could build the Dockerfile, generate an appropiate dockerimage and run the image

### Why is this important to you?
With this option, we can use Tye for any type of projects that has a Dockerfile

rynowak:
Thanks for the this feedback. I think we can already do this for a .NET project - if there's a dockerfile in the project directory we will use this rather than generating one. This doesn't really do everything that you're asking for though.

rynowak:
Here's something I can imagine us doing. Supporting configuring services that build docker images and push, but aren't .NET based.

Currently we have some special behaviors for .NET projects:

- We build and run them
- We generate env-vars for service discovery in a format that .NET `Microsoft.Extensions.Configuration` understands
- We use ASP.NET Core's configuration setting to set listening URLs
- We put secrets on disk in k8s that match .NET's configuration format

Micronetes has a (currently undocumented) other schema for service-discovery + secrets that it can use when the service isn't a .NET project. We'd need to provide configuration support for describing how to build an image, as well plumb the service discovery stuff through all of these layers. We also need a non-ASP.NET Core centric way of telling an app about the listening port/address.

adiazcan:
Ok, I know the limitations from the service discovery perspective. 

I don't know if it's possible or if an out of scope request for this project but why not get the service information from Dockerfile? Having service information on Dockerfile could be a requirement for non-dotnet projects.

davidfowl:
> We build and run them

We only support .NET projects but you can launch an executable or run a docker image. Building a docker file seems logical as well.

> We generate env-vars for service discovery in a format that .NET Microsoft.Extensions.Configuration understands

We generate both a format that .NET expects and another format:

```
{SERVICENAME}_SERVICE_HOST - Host name of the service
{SERVICENAME}_SERVICE_PORT - Port of the service
{SERVICENAME}_SERVICE_PROTOCOL - Protocol of the service
```

> We use ASP.NET Core's configuration setting to set listening URLs

This [PR](https://github.com/dotnet/tye/pull/165/files) tries to improve the situation by providing port information via the PORT env variable.

This would be the same for deployment.

> We put secrets on disk in k8s that match .NET's configuration format

[...]

PR #448, opened 2020-05-06
Allow specifying docker file for tye run

jkotalik:
Fixes https://github.com/dotnet/tye/issues/146

I don't know if well get this into the next release (I'm reluctant to just because I haven't finished the tye deploy story). There are also some paper cuts that are going to make it hard for adoption, including not automatically allocating ports for docker projects (looking into this).

### DIFF
--- a/src/Microsoft.Tye.Hosting/DockerRunner.cs
+++ b/src/Microsoft.Tye.Hosting/DockerRunner.cs
@@ -206,13 +206,15 @@ namespace Microsoft.Tye.Hosting
 
             if (docker.DockerFile != null)
             {
-                var exitCode = await ProcessUtil.RunAsync(
+                var dockerBuildResult = await ProcessUtil.RunAsync(
                     $"docker",
-                    $"build \"{docker.DockerFile.DirectoryName}\" -t {dockerImage} -f \"{docker.DockerFile}\"",
-                    docker.WorkingDirectory);
+                    $"build \"{docker.DockerFileContext.DirectoryName ?? docker.DockerFile.DirectoryName}\" -t {dockerImage} -f \"{docker.DockerFile}\"",
+                    docker.WorkingDirectory,
+                    throwOnError: false);
 
-                if (exitCode.ExitCode != 0)
+                if (dockerBuildResult.ExitCode != 0)
                 {
+                    _logger.LogInformation("Running docker command with exception info {ExceptionStdOut} {ExceptionStdErr}", dockerBuildResult.StandardOutput, dockerBuildResult.StandardError);
                     throw new CommandException("'docker build' failed.");
                 }
             }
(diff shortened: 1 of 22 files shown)


jkotalik:
Something is going on with the skip conditions for Docker on Windows. Last four builds have failed due to tests not being skipped. I'll look into it.

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -195,6 +195,20 @@ private async Task StartContainerAsync(Application application, Service service,
             var volumes = "";
             var workingDirectory = docker.WorkingDirectory != null ? $"-w {docker.WorkingDirectory}" : "";
             var hostname = "host.docker.internal";
+            var dockerImage = docker.Image ?? service.Description.Name;
+
+            if (docker.DockerFile != null)
+            {
+                var exitCode = await ProcessUtil.RunAsync(
+                    $"docker",
+                    $"build \"{docker.DockerFile.DirectoryName}\" -t {dockerImage} -f \"{docker.DockerFile}\"",
```
davidfowl, line 204: We need to support a different build context as well.

jkotalik: How would you recommend exposing that? In the tye.yaml as another part of a service?

test/E2ETest/testassets/projects/dockerfile/tye.yaml
```diff
@@ -0,0 +1,11 @@
+# tye application configuration file
+# read all about it at https://github.com/dotnet/tye
+name: frontend-backend
+services:
+- name: backend
+  dockerFile: backend/Dockerfile
+  bindings:
+    - port: 8002
```
davidfowl, line 8: Random port

davidfowl: You should be able to set containerPort to 80 (but I think it'll default for http) and set nothing else

jkotalik: That needs to be fixed. Right now, a random port isn't specified (and I don't think one is specified for any non-project, but need to confirm that).

davidfowl: Not sure what you mean. If you set containerPort a random port will get generated

src/Microsoft.Tye.Hosting/DockerRunner.cs
davidfowl: Yep

rynowak:
> I don't know if well get this into the next release (I'm reluctant to just because I haven't finished the tye deploy story). There are also some paper cuts that are going to make it hard for adoption, including not automatically allocating ports for docker projects (looking into this).

The port issue needs to be fixed 👍 

I don't really mind if this goes in without deploy support at first. As long as we can document what's working, and it's useful to somebody there's no reason to delay it.

jkotalik:
> The port issue needs to be fixed 👍

So apparently, I didn't know this, but you need to specify the containerPort. Random ports do seem to work if you specify that.

davidfowl:
@amarie401 time to try that first scenario again

PR #456, opened 2020-05-07
Log from stdout when dotnet publish fails

### LINKED ISSUE #339: Improve logs for publish failing for tye deploy
Currently, if dotnet publish fails, the logs look like:
```
Processing Service 'vote'...
    Compiling Services...
    Publishing Project...
Drats! 'deploy' failed:
        'dotnet publish' failed.
```

This should output the reason why dotnet publish failed.

jkotalik:
Fixes #339 

`dotnet publish` logs to stdout instead of stderr with the error. Use ProcessUtils.RunAsync and log to the OutputContext instead.

We can also port other calls to do this as well.

### DIFF
--- a/src/Microsoft.Tye.Core/PublishProjectStep.cs
+++ b/src/Microsoft.Tye.Core/PublishProjectStep.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.CommandLine.Invocation;
 using System.IO;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Tye.Core/PublishProjectStep.cs
```diff
@@ -36,18 +37,29 @@ public override async Task ExecuteAsync(OutputContext output, ApplicationBuilder
...
+
+                foreach (var line in publishResult.StandardOutput.Split(Environment.NewLine))
+                {
+                    output.WriteInfoLine(line);
+                }
+
+                foreach (var line in publishResult.StandardError.Split(Environment.NewLine))
+                {
+                    output.WriteInfoLine(line);
+                }
```
rynowak, line 61: This will be fine, but it wasn't really what I had in mind. I'll send a follow up :)

jkotalik: Yeah I had a feeling it wasn't.

Issue #345, opened 2020-04-10
Can't run Tye for a solution


isaacabraham:
Followed the steps in the [Front End / Back End tutorial](https://github.com/dotnet/tye/blob/master/docs/tutorials/hello-tye/00_run_locally.md) I've been able to successfully create two F# ASP .NET projects which Tye can launch individually. However, when I try to launch them both via a solution file, Tye says: `"Drats! 'run' fail: No services found in "microservice.sln"`.

```

Microsoft Visual Studio Solution File, Format Version 12.00
# Visual Studio 15
VisualStudioVersion = 15.0.26124.0
MinimumVisualStudioVersion = 15.0.26124.0
Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "frontend", "frontend\frontend.fsproj", "{CB8B3182-36E7-49C0-A7DC-BB39542B15B7}"
EndProject
Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "backend", "backend\backend.fsproj", "{C2559B7D-07D4-4096-9929-1AFB9A324022}"
EndProject
Global
	GlobalSection(SolutionConfigurationPlatforms) = preSolution
		Debug|Any CPU = Debug|Any CPU
		Debug|x64 = Debug|x64
		Debug|x86 = Debug|x86
		Release|Any CPU = Release|Any CPU
		Release|x64 = Release|x64
		Release|x86 = Release|x86
	EndGlobalSection
	GlobalSection(SolutionProperties) = preSolution
		HideSolutionNode = FALSE
	EndGlobalSection
	GlobalSection(ProjectConfigurationPlatforms) = postSolution
		{CB8B3182-36E7-49C0-A7DC-BB39542B15B7}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
		{CB8B3182-36E7-49C0-A7DC-BB39542B15B7}.Debug|Any CPU.Build.0 = Debug|Any CPU
		{CB8B3182-36E7-49C0-A7DC-BB39542B15B7}.Debug|x64.ActiveCfg = Debug|Any CPU
		{CB8B3182-36E7-49C0-A7DC-BB39542B15B7}.Debug|x64.Build.0 = Debug|Any CPU
		{CB8B3182-36E7-49C0-A7DC-BB39542B15B7}.Debug|x86.ActiveCfg = Debug|Any CPU
[...]

PR #459, opened 2020-05-08
Check for output type exe

jkotalik:
Fixes https://github.com/dotnet/tye/issues/345.

This just adds another check for <OutputType>exe</OutputType>. It's a pretty limited check.


davidfowl:
Part of that work item was better diagnostics around when projects were skipped or included

jkotalik:
https://github.com/dotnet/tye/issues/461 as that issue had like 3 different parts to it.

rynowak:
> Part of that work item was better diagnostics around when projects were skipped or included

What would this look like? A line of console output for every class library and unit test project?

PR #463, opened 2020-05-08
Ingress changes

davidfowl:
## Easy changes

- Support a single ingress, not many
- Default http and https bindings if none specified
- Treat null path the same as /
- Add the ability to preserve the path (need to do deployment)

## Hard changes (might get cut)

- Make all services private when an ingress is present
- Run ingress in a container when --docker is run
- Hide bindings in the dasboard for private services


rynowak:
> Support a single ingress, not many

One thought that I had - I kinda wonder if we've been approaching this the wrong way. What we've done with ingress reflects the implementation details in a way that makes me think we're not done here - it could be simpler. We're modeling ingress as something you *add on the side of a service*, rather than an *aspect of a service*. 

Can we think about what it would look like to model it in the opposite way? *Services have ingress rules instead of ingress rules pointing at services*

---

I agree btw about removing multiple ingresses. It seems like a detail that doesn't matter, because it wouldn't result in multiple ingress controller IPs in the cluster.

davidfowl:
Why do you think that’s better to attach Ingress rules to services?

rynowak:
> Why do you think that’s better to attach Ingress rules to services?

I haven't decided that it's better 😁. I'm currently working through a bunch of things (not Tye) where decisions were made that were coupled to how Kubernetes is factored. This is the one case that I thought of inside Tye where we made decisions that seem really influenced by *exactly* what k8s does as far as how the concepts are layered.

PR #469, opened 2020-05-12
Tye should validate that service names are valid DNS names

kishanAnem:
#411 
Tye should validate that service names are valid DNS names


davidfowl:
You need to update these names:
- https://github.com/dotnet/tye/blob/f6c1e65e8b04c8006f12ebd4f04712b6d9b00903/test/E2ETest/testassets/projects/nginx-ingress/nginx.conf#L5
- https://github.com/dotnet/tye/blob/f6c1e65e8b04c8006f12ebd4f04712b6d9b00903/test/E2ETest/testassets/projects/nginx-ingress/nginx.conf#L16

src/Microsoft.Tye.Core/ConfigModel/ConfigService.cs
```diff
@@ -4,13 +4,19 @@
...
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
+using Tye;
 using YamlDotNet.Serialization;
 
 namespace Microsoft.Tye.ConfigModel
 {
     public class ConfigService
     {
+        const string errorMessage = "a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')";
```
rynowak, line 14:
```suggestion
        const string errorMessage = "A service name must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')";
```

rynowak: Let's make it clearer that it's about service names. I don't expect most users to know what a DNS label is :)

src/Microsoft.Tye.Core/ConfigModel/ConfigFactory.cs
```diff
@@ -97,5 +98,8 @@ private static ConfigApplication FromYaml(FileInfo file)
             using var parser = new YamlParser(file);
             return parser.ParseConfigApplication();
         }
+
+        private static string Normalization(string name)
```
rynowak, line 102: Can we replace with `-` instead of just removing the character? The common case is for names to look like: `MyCompany.MyDepartment.MyApp` so that would become `mycompanymydepartmentmyapp` instead of the more readable `mycompany-mydepartment-myapp`.

src/Microsoft.Tye.Core/ConfigModel/ConfigFactory.cs
```diff
@@ -73,7 +74,7 @@ private static ConfigApplication FromSolution(FileInfo file)
                 {
                     var service = new ConfigService()
                     {
-                        Name = Path.GetFileNameWithoutExtension(projectFile.Name).ToLowerInvariant(),
+                        Name = Normalization(Path.GetFileNameWithoutExtension(projectFile.Name).ToLowerInvariant()),
```
rynowak, line 77: nitpick: move `ToLowerInvariant()` into the normalization method so it's not copied all over 👍

src/Microsoft.Tye.Core/ConfigModel/ConfigFactory.cs
kishanAnem: Sure will do

src/Microsoft.Tye.Core/ConfigModel/ConfigService.cs
rynowak: This change appears to have been reverted. It's back to saying `a DNS-1035 label...`

samples/apps-with-ingress/tye.yaml
```diff
@@ -17,8 +17,8 @@ ingress:
       - host: a.example.com
         service: app-a
       - host: b.example.com
-        service: app-b 
-
+        service: app-b
+
```
jkotalik, line 21:
```suggestion

```

src/Microsoft.Tye.Core/ConfigModel/ConfigFactory.cs
```diff
@@ -97,5 +98,8 @@ private static ConfigApplication FromYaml(FileInfo file)
             using var parser = new YamlParser(file);
             return parser.ParseConfigApplication();
         }
+
+        private static string Normalization(string name)
+            => Regex.Replace(name.ToLowerInvariant(), "[^0-9A-Za-z-]+", "-");
```
kishanAnem, line 103: ex : "@some.proj!.csproj". it wont be ideal case. with this regex results :

>serivce name =  -some-proj-

isn't ok..

rynowak: There will always be cases that break, it's a heuristic. Have you seen a project called `@some.proj!.csproj` before? 😆

kishanAnem: Before also replacing “.” Is missing. Anyway doubt has clarified Thanks 😁..

kishanAnem: Everything fine??🙂

rynowak:
Looks like the nginx ingress test is failing: https://dev.azure.com/dnceng/public/_build/results?buildId=643307&view=ms.vss-test-web.build-test-results-tab&runId=19977296&resultId=100035&paneView=asg-cloudtest.asg-cloudtest-attachments.asg-cloudtest.asg-cloudtest-attachments

Are the tests passing for you locally?

kishanAnem:
Yeah I did. Will fix.

kishanAnem:
In Anixity forgot to change tests. sorry my bad wasting your time. Thanks @rynowak. :)

kishanAnem:
I think it’s completed. @rynowak isn’t it?

src/Microsoft.Tye.Core/ConfigModel/ConfigService.cs
```diff
@@ -4,13 +4,19 @@
...
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
+using Tye;
 using YamlDotNet.Serialization;
 
 namespace Microsoft.Tye.ConfigModel
 {
     public class ConfigService
     {
+        const string errorMessage = "A service name must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')";
```
davidfowl, line 14: nit: Style These should be ErrorMessage and MaxLengtErrorMessage (pascalcase)

test/E2ETest/TyeInitTests.cs
```diff
@@ -74,5 +74,20 @@ public void Init_ProjectKinds()
 
             YamlAssert.Equals(expectedContent, content);
         }
+
+
+        [Fact]
+        public void console_normalization_svc_name()
```
davidfowl, line 80: The casing should match the other tests (not all lowercase

kishanAnem:
Hi @davidfowl , done

src/Microsoft.Tye.Core/ConfigModel/ConfigFactory.cs
jkotalik, line 102:
```suggestion
        private static string NormalizeServiceName(string name)
```

And add a comment about why we are replacing all characters with -. Also a link to the DNS name spec would be useful.

jkotalik:
I still don't know how I feel about blocking all capital letters for service name. Other characters I understand, but maybe we should just call ToLower on the service name? @davidfowl thoughts?

src/Microsoft.Tye.Core/ConfigModel/ConfigFactory.cs
kishanAnem: 2cb2a43, 64a74df these 2 commits aren't mine. 
so just have reverted and pushed. isn't fine? please help me to reslove PR. @davidfowl @jkotalik .

jkotalik: I think those commits are fine, but I'd recommend doing a `git rebase -i origin/master`, removing the unnecessary commits, and then doing a `git push -f`. Be careful though!

kishanAnem: I’ve not done before will do. Thanks  @jkotalik

davidfowl:
> Other characters I understand, but maybe we should just call ToLower on the service name? @davidfowl thoughts?

Yep

kishanAnem:
I'm closing this PR. will open new one #478

PR #474, opened 2020-05-14
Trim trailing slash

### LINKED ISSUE #471: Setting DockerFileContext doesn't seem to work
### Describe the bug
When manually specifying the DockerFileContext, it gives you the error message from Docker but not the command where it failed.

This doesn't work
```
- name: todo-client
  dockerFile: "./client/Dockerfile"
  dockerFileContext: ./client
  bindings:
    - containerPort: 80
      protocol: http
```

This works

```
- name: todo-client
  dockerFile: "./client/Dockerfile"
  bindings:
    - containerPort: 80
      protocol: http
```

jkotalik:
Fixes #471 

From what I can tell, the issue is that docker build doesn't like providing an absolute path with a trailing slash.

### DIFF
--- a/src/Microsoft.Tye.Hosting/DockerRunner.cs
+++ b/src/Microsoft.Tye.Hosting/DockerRunner.cs
@@ -208,7 +208,8 @@ namespace Microsoft.Tye.Hosting
             {
                 var dockerBuildResult = await ProcessUtil.RunAsync(
                     $"docker",
-                    $"build \"{docker.DockerFileContext?.DirectoryName ?? docker.DockerFile.DirectoryName}\" -t {dockerImage} -f \"{docker.DockerFile}\"",
+                    // Supplying an absolute path with trailing slashes fails here, so trim trailing slash.
+                    $"build \"{docker.DockerFileContext?.FullName.Trim(Path.DirectorySeparatorChar) ?? docker.DockerFile.DirectoryName}\" -t {dockerImage} -f \"{docker.DockerFile}\"",
                     docker.WorkingDirectory,
                     throwOnError: false);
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -208,7 +208,8 @@ private async Task StartContainerAsync(Application application, Service service,
             {
                 var dockerBuildResult = await ProcessUtil.RunAsync(
                     $"docker",
-                    $"build \"{docker.DockerFileContext?.DirectoryName ?? docker.DockerFile.DirectoryName}\" -t {dockerImage} -f \"{docker.DockerFile}\"",
+                    // Supplying an absolute path with trailing slashes fails here, so trim trailing slash.
+                    $"build \"{docker.DockerFileContext?.FullName.Trim(Path.DirectorySeparatorChar) ?? docker.DockerFile.DirectoryName}\" -t {dockerImage} -f \"{docker.DockerFile}\"",
```
davidfowl, line 212: I think we should normalize earlier (near where he file is read)

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -208,7 +208,7 @@ private async Task StartContainerAsync(Application application, Service service,
             {
                 var dockerBuildResult = await ProcessUtil.RunAsync(
                     $"docker",
-                    $"build \"{docker.DockerFileContext?.DirectoryName ?? docker.DockerFile.DirectoryName}\" -t {dockerImage} -f \"{docker.DockerFile}\"",
+                    $"build \"{docker.DockerFileContext?.FullName ?? docker.DockerFile.DirectoryName}\" -t {dockerImage} -f \"{docker.DockerFile}\"",
```
davidfowl, line 211: Move the fallback as well to so we don't need to write this logic here?

PR #477, opened 2020-05-15
angular dotnet core sample for project tye

rahulsahay19:
@davidfowl 

Hi,

As discussed I have created a PR for your review. I have also added a Readme file explaining setup part.

Thanks,
Rahul Sahay


samples/apps-with-core-angular/tye.yaml
```diff
@@ -0,0 +1,19 @@
...
+name: project-tye
+services:
+- name: moviesapi
+  project: MoviesAPI/MoviesAPI.csproj
+  bindings:
+  - protocol: https
+    port: 5001
+  
+- name: moviesapp
+  image: moviesapp:dev
```
jkotalik, line 16: For `tye run`, instead of specifying the image, could you specify the `dockerFile` as a relative path to the docker file? Curious to see if that works. This doesn't work for `tye deploy` yet, but it will soon.

rahulsahay19: @jkotalik 
I have tested and added the change as default option tye file. However, in readme, I have explained both the scenarios with Dockerfile and with image. Thanks.

samples/apps-with-core-angular/tye.yaml
```diff
@@ -0,0 +1,19 @@
...
+# read all about it at https://github.com/dotnet/tye
+#
+# when you've given us a try, we'd love to know what you think:
+#    https://aka.ms/AA7q20u
+#
+name: project-tye
+services:
+- name: moviesapi
+  project: MoviesAPI/MoviesAPI.csproj
+  bindings:
```
davidfowl, line 11: Remove the port?

samples/apps-with-core-angular/tye.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - protocol: https
+    port: 5001
+  
+- name: moviesapp
+  dockerFile: MoviesApp/Dockerfile
+  bindings:
+  - protocol: http
+    port: 4400
```
davidfowl, line 19: Remove the port?

samples/apps-with-core-angular/tye.yaml
rahulsahay19: Hi, Do you want me to remove the port? I kept it because angular service will be calling that API on that particular port? angular app port can be removed as its not getting called from any app

rahulsahay19: @davidfowl  Hi, 

You were saying sometime back that you wanted to simplify this further your way. Then, maybe you can merge this and do the required changes on top of it. Otherwise, you can pass on the instructions what you wanted to do. I will accommodate that. Because I don't understand "Remove the port?" thing. If I remove it from .net app, then angular app will fail as it won't be able to discover the port what api is exposing.

Thanks,
Rahul

PR #489, opened 2020-05-21
Add docker file support for deploy 

jkotalik:
Fixes https://github.com/dotnet/tye/issues/146

Adds a new type that represents a DockerFile (so it gets ports and stuff).

### DIFF
--- a/src/Microsoft.Tye.Core/DockerFileServiceBuilder.cs
+++ b/src/Microsoft.Tye.Core/DockerFileServiceBuilder.cs
@@ -1,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Tye
+{
+    public class DockerFileServiceBuilder : ProjectServiceBuilder
+    {
+        public DockerFileServiceBuilder(string name, string image)
+            : base(name)
+        {
+            Image = image;
+        }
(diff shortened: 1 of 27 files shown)


rynowak:
Is the idea that we'll generate a k8s deployment/service for a dockerfile pretty much the same we we would for a project? SPICY.

jkotalik:
Yep, only for files that have a docker file though, not just random images.

davidfowl:
> Is the idea that we'll generate a k8s deployment/service for a dockerfile pretty much the same we we would for a project? SPICY.

I'm worried about the versioning/tagging.

src/Microsoft.Tye.Core/DockerFileServiceBuilder.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Tye
+{
+    public class DockerFileServiceBuilder : ProjectServiceBuilder
```
rynowak, line 7: What does this represent? Is this a .NET project with a Dockerfile?

Or is this meant to capture a non-project that has a Dockerfile?

Bascially the use of inheritance seems like a convenient hack.

jkotalik: > Or is this meant to capture a non-project that has a Dockerfile?

This one.

It was a very convenient "refactoring".

PR #496, opened 2020-05-22
Fix dotnet feed publishing

sebastienros:
Removing the steps that create AssetManifest files from Linux to prevent the Publish task from trying to upload the same assets multiple times in parallel and fail on file locks.
Removing empty parameter on macOS step.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -118,8 +118,6 @@ stages:
         - script: eng/common/cibuild.sh
             --configuration $(_BuildConfig)
             --prepareMachine
-            $(_OfficialBuildArgs)
-            /p:DotNetPublishUsingPipelines=$(_PublishUsingPipelines)
           displayName: Build
         - task: PublishBuildArtifacts@1
           displayName: Upload TestResults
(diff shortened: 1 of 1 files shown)


sebastienros:
I'd like to know where this pipeline file was taken from to see if the source was fixed.
Also please check that there are no custom artifacts that are created specifically in the Linux step.

sebastienros:
I verified that both Windows and Linux steps are generating the same package assets:

```
  <Package Id="Microsoft.Tye" Version="0.3.0-alpha.20271.2" />
  <Package Id="Microsoft.Tye.Core" Version="0.3.0-alpha.20271.2" />
  <Package Id="Microsoft.Tye.Extensions" Version="0.3.0-alpha.20271.2" />
  <Package Id="Microsoft.Tye.Extensions.Configuration" Version="0.3.0-alpha.20271.2" />
  <Package Id="Microsoft.Tye.Hosting" Version="0.3.0-alpha.20271.2" />
  <Package Id="Microsoft.Tye.Hosting.Diagnostics" Version="0.3.0-alpha.20271.2" />
  <Package Id="Microsoft.Tye.Hosting.Runtime" Version="0.3.0-alpha.20271.2" />
  <Package Id="Microsoft.Tye.Proxy" Version="0.3.0-alpha.20271.2" />
  <Package Id="tye-diag-agent" Version="0.3.0-alpha.20271.2" />
  ```

So unless they contain custom native binaries then it should be fine. Otherwise their package names should have included the os.

jkotalik:
Thanks.

PR #502, opened 2020-05-24
Liveness and Readiness Documentation, Samples and Small Changes

areller:
## Changes

* Expanded the `schema.md` document to contain `liveness` and `readiness` configuration
* Added a sample project and a recipe 
* Added basic deserialization and validation tests
* Changed some logs in `ReplicaMonitor` to debug, and some logs to information. 

About the last point, If service has liveness/readiness with `initialDelay` of zero, when running `tye run` you will see some warnings, because it takes time for the service to start accepting requests.  

This is okay since the fact that the default for `failureThreshold` is 3 (and the minimum is 1) accounts for it.  

Still, no reason to show warning logs so I've changed these logs to Debug, and other logs that the user should see, to Information. 

Problem is that `tye run` shows debug logs, and there is no way to control it. 
Should I create a PR that makes the verbosity of the logs configurable?

### DIFF
--- a/docs/recipes/probes.md
+++ b/docs/recipes/probes.md
@@ -11,13 +11,19 @@ Probes serve two main functions
 
 Before we get to the definition of a probe and how it works, it's important to understand the life cycle of a replica.  
 
-Each replica in Tye can be in one of these three states  
+When using the local orchestrator (`tye run`), each replica in Tye can be in one of these three states  
 
 * `Started` - A replica is in this state when it has just been started, and hasn't been probed yet.  
 * `Healthy` - A replica is in this state when it passes the `liveness` probe, but not the `readiness` probe. 
 * `Ready` - A replica is in this state when it passes both the `liveness` and `readiness` probes.  
 
 (*Internally, there are more states that a replica can be in, but those are not relevant to this discussion*)
+
+The orchestrator is responsible for switching between the different states based on the feedback from the probes, as described in later sections.  
+
+When deploying the application to Kubernetes (via `tye deploy`), the life cycle is represented differently and is managed by Kubernetes. Click [here](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) to read more about the life cycle of a Pod in Kubernetes.  
+
+Throughout the rest of the document, we'll be referring to the life cycle of the replica as it is represented in the local orchestrator (i.e. `Started`, `Healthy`, `Ready`).
 
 ## Types of Probes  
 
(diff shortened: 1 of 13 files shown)


jkotalik:
> Problem is that tye run shows debug logs, and there is no way to control it.
Should I create a PR that makes the verbosity of the logs configurable?

Feel free to change the default to informational and add a way to have different verbosities for logs for run.

docs/recipes/probes.md
```diff
@@ -0,0 +1,172 @@
...
+Probes serve two main functions  
+
+* Keeping traffic away from replicas which are not ready to receive it.
+* Restarting replicas which are in a Bad/Unhealthy state.
+
+## Life cycle of a Replica
+
+Before we get to the definition of a probe and how it works, it's important to understand the life cycle of a replica.  
+
+Each replica in Tye can be in one of these three states
```
jkotalik, line 14: I'd try to emphasize that for tye deploy, Kubernetes is managing these states for started, healthy, and ready.

docs/recipes/probes.md
```diff
@@ -0,0 +1,172 @@
...
+* Demote a `Ready` replica to `Healthy` when the probe fails.  
+
+Tye only routes traffic (either via a service binding, or an ingress) to `Ready` replicas.  
+
+## Life cycle of a Replica in the Absence of Probes
+
+By now, the life cycle of a replica when both `liveness` and `readiness` probes are configured should be clear, but how does the life cycle of a replica look like when both probes or one of the probes is absent?  
+
+* When both `liveness` and `readiness` probes are absent, a replica gets promoted from `Started` to `Ready` automatically, upon creation.
+* When only the `readiness` probe is absent, a replica gets promoted from `Started` to `Healthy` only after it passes the `liveness` probe, but upon being promoted to `Healthy`, it's automatically promoted again, to `Ready`.
```
jkotalik, line 48: I'd invert these sentences to "When only the `readiness` probe is present.

docs/recipes/probes.md
```diff
@@ -0,0 +1,172 @@
...
+    replicas: 2
+    liveness:
+      http:
+        path: /healthy
+    readiness:
+      http:
+        path: /ready
+```
+
+The service is configured to respond successfully to both the `liveness` and `readiness` probes, so after you run Tye, you should see these log lines
```
jkotalik, line 76:
```suggestion
The service is configured to respond successfully to both the `liveness` and `readiness` probes, so after executing `tye run`, you should see these log lines  
```

docs/recipes/probes.md
```diff
@@ -0,0 +1,172 @@
...
+        path: /healthy
+    readiness:
+      http:
+        path: /ready
+```
+
+The service is configured to respond successfully to both the `liveness` and `readiness` probes, so after you run Tye, you should see these log lines  
+
+```
+[18:14:05 INF] Replica simple-webapi_a2d67bd9-4 is moving to an healthy state
```
jkotalik, line 79: Are these logs now debug level? If we change the tye run command to not log debug by default, make sure to specify the log level to debug.

areller: These logs were `debug` before this PR, but were changed to `info`

docs/recipes/probes.md
```diff
@@ -0,0 +1,172 @@
...
+
+As you can see, both replicas pass the `liveness` probe and get prompted to an `Healthy` state, and shortly after, both replica pass the `readiness` probe and get promoted to a `Ready` state.  
+
+The sample application exposes an endpoint that allows you modify the responses that the `/healthy` and `/ready` endpoints, in order to see the probes in action.  
+
+For example, if you send an *HTTP POST* request to this endpoint `http://localhost:8080/set` with this body  
+
+```
+{
+	"Ready": false,
```
jkotalik, line 93: Can you use query parameters instead of a post message to populate these values? It's a lot easier for people to create a get request with a url (ex just copying the url into a browser) versus a POST request (using postman or curl).

docs/recipes/probes.md
jkotalik: Ah gotcha.

jkotalik:
> Problem is that tye run shows debug logs, and there is no way to control it.
Should I create a PR that makes the verbosity of the logs configurable?

Yeah, feel free to create a follow up PR. Tye deploy currently has -v/ --verbosity option to specify the log level.

areller:
@jkotalik any news on this?

davidfowl:
This is really good @areller I think we're good to merge this now but I have some general feedback:
- We should show using this with the health checks library that ASP.NET Core ships.
- Do we avoid the proxy when there are no probes? (We should)

areller:
@davidfowl 1. wasn't familiar with the library. going to check it out and make an health check with it.
2. you mean, if a service has a single replica and no probes, do we not create a proxy? we don't https://github.com/dotnet/tye/blob/fc4dc9281665f3274cc8cec3ae9bc572d163e08c/src/Microsoft.Tye.Hosting/ProxyService.cs#L60

davidfowl:
> 2. you mean, if a service has a single replica and no probes, do we not create a proxy? we don't

Sweet!

davidfowl:
@areller I'm going to merge this and then you can send an update based on the ASP.NET Core library,

PR #506, opened 2020-05-26
log verbosity in tye run

areller:
We allow the user to specify the level of output verbosity when running `tye run`
The levels are 

* **Debug** - display all logs
* **Info** (Default) - display only info (or warn/error) logs
* **Quiet** - display only warn/error logs

### DIFF
--- a/src/Microsoft.Tye.Hosting/TyeHost.cs
+++ b/src/Microsoft.Tye.Hosting/TyeHost.cs
@@ -12,7 +12,6 @@ using System.Net.Sockets;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Hosting;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -112,8 +111,16 @@ private static WebApplication BuildWebApplication(Application application, HostO
             // Logging for this application
             builder.Host.UseSerilog((context, configuration) =>
             {
+                var logLevel = options.LogVerbosity switch
+                {
+                    Verbosity.Quiet => LogEventLevel.Warning,
+                    Verbosity.Info => LogEventLevel.Information,
+                    Verbosity.Debug => LogEventLevel.Verbose,
+                    _ => default
```
davidfowl, line 119: What's the default?

areller: You mean `_ => default`?
This is to appease the C# compiler :)
It throws a warning otherwise

davidfowl: I’m asking why use default and not an end value like Verbose or Info for the default case

areller: That `_ => default` should never happen. It's just an artifact of the compiler treating the enum as an integer. There is already a place where we set the default value for the verbosity, so I didn't want to repeat that in multiple places.

davidfowl: Ah, I got it.

davidfowl:
Thanks!

Issue #482, opened 2020-05-16
Tye 0.2.* and 0.3.* with Dapr extension isn't working


thangchung:
### Describe the bug

`tye` `0.3.0-alpha.20265.1+ef441e87047930e04d24646e0d89183b07552747` with `dapr extensions` isn't working, and the the same issue with `tye` `0.2.0-alpha.20258.3`

### To Reproduce

I upgraded my practical-dapr project to use `tye` 0.3 following the guidance document at https://github.com/dotnet/tye/blob/master/docs/getting_started.md#working-with-ci-builds, as below

```
$ dotnet tool install -g Microsoft.Tye --version "0.3.0-*" --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json
```

Everything was working after I adapted my code with the `tye` tool at version 0.3.
But when I added `dapr extensions` config as below

```yaml
extensions:
- name: dapr
  log-level: debug
```

Then all worked projects above couldn't run anymore

![image](https://user-images.githubusercontent.com/422341/82119997-1f63ec00-97ad-11ea-82fd-6dee0bf29f6b.png)

After I cancelled `tye CLI` it threw the exception

```
# tye run
Loading Application Details...
Launching Tye Host...

[19:48:59 INF] Executing application from D:\github\practical-dapr\tye.yaml
[19:49:00 INF] Dashboard running on http://127.0.0.1:8000
[19:49:00 INF] Docker image mcr.microsoft.com/mssql/server:2017-latest already installed
[19:49:00 INF] Docker image redis already installed
[19:49:00 INF] Creating docker network tye_network_8c0faa56-4
[19:49:00 INF] Running docker command network create --driver bridge tye_network_8c0faa56-4
[19:49:01 INF] Running image mcr.microsoft.com/mssql/server:2017-latest for sqlserver_98aff541-2
[19:49:01 INF] Running image redis for redis_3a1d29cb-6
[19:49:01 INF] Running image mcr.microsoft.com/dotnet/core/sdk:3.1 for graphapi-proxy_412bbd63-a
[19:49:01 INF] Running image mcr.microsoft.com/dotnet/core/sdk:3.1 for identityserver-proxy_e3751a17-f
[19:49:01 INF] Running image mcr.microsoft.com/dotnet/core/sdk:3.1 for productcatalogapi-proxy_11e14a58-6
[...]

PR #510, opened 2020-05-27
Fix dapr and environment variable injection

jkotalik:
This has two fixes:
1. Fixes dapr due to improperly populating dependencies (fixes https://github.com/dotnet/tye/issues/482).
2. Fixes regression where services couldn't inject environment variables set on themselves due to a service not having a direct dependency on itself. For example, after adding the fix to the DaprExtension, dapr was still failing to start because it couldn't populate the environment variable for DAPR_GRPC_PORT.

### DIFF
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -401,11 +401,7 @@ namespace Microsoft.Tye
             // Iterate through all services and add the current services as a dependency (except ourselves)
             foreach (var s in root.Services)
             {
-                if (dependencies.Contains(s.Name, StringComparer.OrdinalIgnoreCase)
-                    && !s.Name.Equals(serviceName, StringComparison.OrdinalIgnoreCase))
-                {
-                    s.Dependencies.Add(serviceName);
-                }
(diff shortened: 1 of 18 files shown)


src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -398,14 +398,10 @@ private static void AddToRootServices(ApplicationBuilder root, HashSet<string> d
...
-            // Iterate through all services and add the current services as a dependency (except ourselves)
+            // Iterate through all services and add the current services as a dependency
             foreach (var s in root.Services)
             {
-                if (dependencies.Contains(s.Name, StringComparer.OrdinalIgnoreCase)
-                    && !s.Name.Equals(serviceName, StringComparison.OrdinalIgnoreCase))
-                {
-                    s.Dependencies.Add(serviceName);
-                }
+                s.Dependencies.Add(serviceName);
```
rynowak, line 404: does this mean that every service has a dependency on itself? 

If so, why isn't the right fix elsewhere (use env-vars from service's dependencies + the service)?

jkotalik: Yeah, you are right that fix is more sensible.

Issue #511, opened 2020-05-29
Pass build arguments into docker build step when using a yaml file


ossentoo:
Hi,

My application pulls private nuget packages from an Azure DevOps feed in the docker image build.  We normally accomplish this in our Azure DevOps pipeline by passing a build argument to docker build.

I've setup a yaml file to run locally at the moment. This is failing during the dotnet restore step in the Dockerfile as expected, because it needs the credential to the private nuget feed.

Is there a way of passing this in? I'm running locally so, I don't mine passing the PAT in somehow.

I've tried this:

name: application
services:
- name: web
  dockerFile: ../application-web/src/Dockerfile --build-arg PAT=***

I get this output when running tye run with a tye.yaml file in the same directory:

[06:44:57 INF] Building docker image web from docker file
[06:44:57 INF] [web]:unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /Users/blah/application-web/src/Dockerfile --build-arg PAT=***: no such file or directory

jkotalik:
Yeah, today passing in `dockerFile` assumes that it is only a path to the docker file, and doesn't include args.

@davidfowl for `docker run`, do we allow passing in args as well? Is that through the `args` section in the yaml?

ossentoo:
Ok.

Is there any work around for now?  Can I call docker build for example for now?

davidfowl:
I thought we had support for arguments being passed to build but there isn't we'll need to add it.

ossentoo:
I'm going to work on a PR for this issue

PR #521, opened 2020-05-31
Added support for docker build arguments

ossentoo:
This resolves the #511  and provides support for docker build arguments.

### DIFF
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -134,7 +134,7 @@ namespace Microsoft.Tye
                             DockerFile = Path.Combine(source.DirectoryName, configService.DockerFile),
                             // Supplying an absolute path with trailing slashes fails for DockerFileContext when calling docker build, so trim trailing slash.
                             DockerFileContext = GetDockerFileContext(source, configService),
-                            BuildArgs = configService.DockerFileArgs
+                            DockerBuildArgs = configService.DockerFileArgs
                         };
                         service = dockerFile;
 
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -133,7 +133,8 @@ namespace Microsoft.Tye
                             Replicas = configService.Replicas ?? 1,
                             DockerFile = Path.Combine(source.DirectoryName, configService.DockerFile),
                             // Supplying an absolute path with trailing slashes fails for DockerFileContext when calling docker build, so trim trailing slash.
-                            DockerFileContext = GetDockerFileContext(source, configService)
+                            DockerFileContext = GetDockerFileContext(source, configService),
+                            BuildArgs = configService.DockerFileArgs
                         };
                         service = dockerFile;
 
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -134,7 +134,7 @@ namespace Microsoft.Tye
                             DockerFile = Path.Combine(source.DirectoryName, configService.DockerFile),
                             // Supplying an absolute path with trailing slashes fails for DockerFileContext when calling docker build, so trim trailing slash.
                             DockerFileContext = GetDockerFileContext(source, configService),
-                            DockerBuildArgs = configService.DockerFileArgs
+                            BuildArgs = configService.DockerFileArgs
                         };
                         service = dockerFile;
 
--- a/src/Microsoft.Tye.Core/DockerFileServiceBuilder.cs
+++ b/src/Microsoft.Tye.Core/DockerFileServiceBuilder.cs
@@ -15,8 +15,8 @@ namespace Microsoft.Tye
         }
         public string Image { get; set; }
 
-        public Dictionary<string, string> BuildArgs { get; set; } = new Dictionary<string, string>();
         public string? DockerFile { get; set; }
+        public Dictionary<string, string> BuildArgs { get; set; } = new Dictionary<string, string>();
 
         public string? DockerFileContext { get; set; }
     }
--- a/src/Microsoft.Tye.Core/DockerFileServiceBuilder.cs
+++ b/src/Microsoft.Tye.Core/DockerFileServiceBuilder.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.Generic;
+
 namespace Microsoft.Tye
 {
     public class DockerFileServiceBuilder : ProjectServiceBuilder
@@ -13,6 +15,7 @@ namespace Microsoft.Tye
         }
         public string Image { get; set; }
 
+        public Dictionary<string, string> BuildArgs { get; set; } = new Dictionary<string, string>();
         public string? DockerFile { get; set; }
 
         public string? DockerFileContext { get; set; }
--- a/src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
+++ b/src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
@@ -478,17 +478,18 @@ namespace Tye.Serialization
         }
         private static void HandleServiceDockerArgsNameMapping(YamlMappingNode yamlMappingNode, IDictionary<string, string> dockerArguments)
         {
-            // only expecting on child at this level
-            var child = yamlMappingNode!.Children.FirstOrDefault();
-            var key = YamlParser.GetScalarValue(child.Key);
-            var value = YamlParser.GetScalarValue(key, child.Value);
-
-            if (string.IsNullOrEmpty(key) || string.IsNullOrEmpty(value) || yamlMappingNode!.Children.Count > 1)
+            foreach (var child in yamlMappingNode!.Children)
             {
-                throw new TyeYamlException(child.Key.Start, CoreStrings.FormatUnrecognizedKey(key));
-            }
+                var key = YamlParser.GetScalarValue(child.Key);
+                var value = YamlParser.GetScalarValue(key, child.Value);
 
-            dockerArguments.Add(key, value);
+                if (string.IsNullOrEmpty(key) || string.IsNullOrEmpty(value))
+                {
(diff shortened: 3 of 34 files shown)


ossentoo:
Weird, not sure why the formatting build for the PR has failed.  Code looks fine on the line mentioned.  Any comments as to why?

davidfowl:
Maybe reuse BuildProperties?

jkotalik:
Aren't BuildProperties specifically properties that are used for arguments to msbuild? Seems like a bad idea to reuse.

src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
```diff
@@ -48,6 +47,14 @@ private static void HandleServiceNameMapping(YamlMappingNode yamlMappingNode, Co
                     case "dockerFile":
                         service.DockerFile = YamlParser.GetScalarValue(key, child.Value);
                         break;
+                    case "dockerFileArgs":
```
jkotalik, line 50: @davidfowl thoughts on making these nested instead of flat? For example
```yaml
services
- name: foo
  docker:
    file: <path to file>
    context: <path to context directory>
    args: <args>

File: src/Microsoft.Tye.Core/ApplicationFactory.cs [hunk 2]
```diff
@@ -130,7 +130,8 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
                             Replicas = configService.Replicas ?? 1,
                             DockerFile = Path.Combine(source.DirectoryName, configService.DockerFile),
                             // Supplying an absolute path with trailing slashes fails for DockerFileContext when calling docker build, so trim trailing slash.
-                            DockerFileContext = GetDockerFileContext(source, configService)
+                            DockerFileContext = GetDockerFileContext(source, configService),
+                            BuildArgs = configService.DockerFileArgs
```
jkotalik, line 134:
```suggestion
                            DockerBuildArgs = configService.DockerFileArgs
```

src/Microsoft.Tye.Core/DockerFileServiceBuilder.cs
```diff
@@ -13,6 +15,7 @@ public DockerFileServiceBuilder(string name, string image)
         }
         public string Image { get; set; }
 
+        public Dictionary<string, string> BuildArgs { get; set; } = new Dictionary<string, string>();
```
jkotalik, line 18: Move below DockerFile to keep ordering consistent.

src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
```diff
@@ -444,5 +460,20 @@ private static void HandleServiceBuildPropertyNameMapping(YamlMappingNode yamlMa
...
+        {
+            // only expecting on child at this level
+            var child = yamlMappingNode!.Children.FirstOrDefault();
+            var key = YamlParser.GetScalarValue(child.Key);
+            var value = YamlParser.GetScalarValue(key, child.Value);
+
+            if (string.IsNullOrEmpty(key) || string.IsNullOrEmpty(value) || yamlMappingNode!.Children.Count > 1)
+            {
+                throw new TyeYamlException(child.Key.Start, CoreStrings.FormatUnrecognizedKey(key));
+
```
jkotalik, line 473:
```suggestion
```

src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
```diff
@@ -393,6 +400,15 @@ private static void HandleBuildProperties(YamlSequenceNode yamlSequenceNode, Lis
                 buildProperties.Add(buildProperty);
             }
         }
+        private static void HandleDockerFileArgs(YamlSequenceNode yamlSequenceNode, Dictionary<string, string> dockerArguments)
+        {
+            foreach (var child in yamlSequenceNode.Children)
+            {
+                YamlParser.ThrowIfNotYamlMapping(child);
+                var keyValue = HandleServiceDockerArgsNameMapping((YamlMappingNode)child);
```
jkotalik, line 408: nit: seems odd to return a key value pair here when you could just add the pair in HandleServiceDockerArgsNameMapping.

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -468,9 +468,15 @@ void Log(string data)
                         service.Logs.OnNext(data);
                     }
 
+                    var arguments = $"build \"{docker.DockerFileContext?.FullName}\" -t {dockerImage} -f \"{docker.DockerFile}\"";
+                    foreach (var buildArg in docker.BuildArgs)
+                    {
+                        arguments += $" --build-arg {buildArg.Key}={buildArg.Value}";
```
jkotalik, line 474: nit: StringBuilder is probably a good idea here as this will new up a string for each build arg.

davidfowl:
No they are name value pairs. The entire configuration system is a bunch of discriminated unions based on a few key properties. The rest are readily being reused for things (like args being used for exes and docker image running). I don’t have a problem making new properties but we should’ve consistent about it

test/E2ETest/Microsoft.Tye.E2ETests.csproj
```diff
@@ -33,6 +33,11 @@
 
   <ItemGroup>
     <None Remove="testassets\generate\dockerfile.yaml" />
+    <None Remove="testassets\projects\single-phase-dockerfile-args\appsettings.Development.json" />
```
jkotalik, line 36: nit: revert this.

src/tye/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
```
jkotalik, line 1: nit: revert this file change.

test/E2ETest/TyeBuildTests.Dockerfile.cs
```diff
@@ -86,5 +87,44 @@ public async Task TyeBuild_SinglePhase_ExistingDockerfile()
...
+        {
+            var projectName = "single-phase-dockerfile-args";
+            var environment = "production";
+            var imageName = "test/web";
+
+            await DockerAssert.DeleteDockerImagesAsync(output, imageName);
+
+            using var projectDirectory = CopyTestProjectDirectory(projectName);
+            Assert.True(File.Exists(Path.Combine(projectDirectory.DirectoryPath, "Dockerfile")), "Dockerfile should exist.");
+
```
jkotalik, line 103:
```suggestion
```

test/E2ETest/TyeBuildTests.Dockerfile.cs
```diff
@@ -86,5 +87,44 @@ public async Task TyeBuild_SinglePhase_ExistingDockerfile()
...
+            application.Registry = new ContainerRegistry("test");
+
+            try
+            {
+                await BuildHost.ExecuteBuildAsync(outputContext, application, environment, interactive: false);
+
+                Assert.Single(application.Services.Single().Outputs.OfType<DockerImageOutput>());
+                var builder = (DockerFileServiceBuilder)application.Services.First();
+                var valuePair = builder.BuildArgs.First();
+                Assert.Equal("pat", valuePair.Key);
```
jkotalik, line 119: Some other things you could test here:
- What happens when duplicated build args are passed in?
- Passing in multiple args

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -468,9 +468,15 @@ void Log(string data)
                         service.Logs.OnNext(data);
                     }
 
+                    var arguments = $"build \"{docker.DockerFileContext?.FullName}\" -t {dockerImage} -f \"{docker.DockerFile}\"";
```
jkotalik, line 471: I'm still worried this won't give people the flexibility to add any arbitrary command. For example, there are a ton of [docker build options that can be passed in](https://docs.docker.com/engine/reference/commandline/build/#options) that can be passed in. Potentially supporting those may important as well outside of build-arg

jkotalik:
This is heading in the right direction, thanks for contributing.

src/Microsoft.Tye.Core/DockerFileServiceBuilder.cs
ossentoo: done

src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
ossentoo: refactored

src/tye/Properties/launchSettings.json
ossentoo: Do you mean the file: launchSettings.json should be deleted?

I made a copy of the single-phase-dockerfile test folder, which also contains a file like this and the contents are the same.  I can change the ports or other details in the file if you think the two tests could end up clashing

src/Microsoft.Tye.Hosting/DockerRunner.cs
ossentoo: I would prefer to limit the scope of this change to build-arg only at this stage.  

I think if the other docker build options are name value pairs, this change should still work.  However, i take the point that i haven't tested anything other than build-args

src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
davidfowl: Not as part of this change. That's a bigger one.

src/Microsoft.Tye.Hosting/DockerRunner.cs
ossentoo: Done

test/E2ETest/Microsoft.Tye.E2ETests.csproj
ossentoo: Removed

src/tye/Properties/launchSettings.json
ossentoo: Removed

test/E2ETest/TyeBuildTests.Dockerfile.cs
ossentoo: what's the behaviour that you think is valid for duplicate args?  take the first of the duplicates, or fail gracefully?

ossentoo: Added multiple parameter functionality and test

PR #523, opened 2020-05-31
using the ASP.NET Core HealthChecks library for the liveness/readiness sample

areller:
As discussed in #502

### DIFF
--- a/docs/recipes/probes.md
+++ b/docs/recipes/probes.md
@@ -66,20 +66,35 @@ tye run
 
 The sample has a single service with a `liveness` and a `readiness` probe, as shown in this snippet
 
-```
+```yaml
 services:
   - name: simple-webapi
     project: webapi/webapi.csproj
     replicas: 2
     liveness:
       http:
-        path: /healthy
+        path: /lively
     readiness:
       http:
         path: /ready
 ```
 
-The service is configured to respond successfully to both the `liveness` and `readiness` probes, so after executing `tye run`, you should see these log lines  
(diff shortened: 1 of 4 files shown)


docs/recipes/probes.md
```diff
@@ -66,20 +66,35 @@ tye run
...
+    // for example: you may consider failing this check if your service has encountered a fatal exception, or if you've detected a memory leak or a substantially long average response time
+    .AddCheck("someLivenessCheck", new MyGenericCheck(_statusDictionary, "someLivenessCheck"), failureStatus: HealthStatus.Unhealthy, tags: new[] { "liveness" })
+    // this registers a "readiness" check. A service that fails a readiness check is considered to be unable to serve traffic temporarily. The orchestrator doesn't restart a service that fails this check, but stops sending traffic to it until it responds to this check positively again.
+    // for example: you may consider failing this check if your service is currently unable to connect to some external service such as your database, cache service, etc...
+    .AddCheck("someReadinessCheck", new MyGenericCheck(_statusDictionary, "someReadinessCheck"), failureStatus: HealthStatus.Unhealthy, tags: new[] { "readiness" });
+```
+
+(Starting .NET 5, the HealthChecks library ships together with ASP.NET Core)
```
jkotalik, line 95: I believe this package has been shipping since 2.2. I'm not sure if we are shipping the HealthChecks library with ASP.NET Core though.

areller: you're right. I saw that the HealthChecks library moved to the `aspnetcore` repository after .NET 5, but it has been and still part of the extensions library.

davidfowl:
Thanks for the doc! I think we should leave the readiness check and liveness check as fake checks that don't change as a result of an http request. It's fine in a sample but I would leave it out of the doc since it's not exactly what you would do in practice.

areller:
@davidfowl  I removed the whole talk about the `/set` endpoint away from the doc, but I added it to the `/` endpoint of the sample service. how's that?

jkotalik:
LGTM, thanks!

Issue #499, opened 2020-05-23
watch project changes


alefcarlos:
### What should we add or change to make your life better?

It would be cool if we had hot reload, like `dotnet watch run` when dependency type is `project`

### Why is this important to you?

It is because it is easier to make changes on the fly. I always use `donet watch` when  running applications.

cvalerio:
I agree, since tye is a development tool it should allow to reload a specific project at runtime (via dashboard) or leverage `dotnet watch run`, maybe with a switch option at `tye run` command.

jkotalik:
I opened a PR that allows passing `--watch` to tye. This will call `dotnet watch run` on all projects. I'd appreciate any feedback on it.

jkotalik:
This is now merged. Would appreciate people trying the nightlies to double check as we'd like to make a release of tye this week 😄

alefcarlos:
I'm gonna try it!

myty:
@jkotalik Adding `--watch` to tye run is definitely picking up file changes and rebuilding the project, so 👍 for that, however it doesn't seem to pick up the change until I actually shutdown tye and start it backup again.  Do I need to do something different with the yaml file maybe? https://github.com/myty/openbrewerydb-graphql-dotnet/blob/feature/dockerize/tye.yaml

jkotalik:
> however it doesn't seem to pick up the change until I actually shutdown tye and start it backup again. Do I need to do something different with the yaml file maybe?

Can you clarify what this means? If you need to shutdown tye and start it again, it makes me think it's not working.

myty:
Yeah, It's definitely rebuilding when the files changes, but I have a feeling it's not updating the container maybe? Not sure exactly 😄

jkotalik:
Ah, that may be a scenario that needs to be updated. I'll look into that.

myty:
Awesome. Thank you!

By the way, keep up the great work. Really liking the vision and direction of this project.

cveld:
@jkotalik With `tye run --watch`, whenever I change a file, the full dotnet project is rebuilt which disconnects the Blazor Server websocket connection. Any clue what I can do to circumvent this? 

```
tye --version
0.11.0-alpha.22111.1+3edef5428949c518c078844d6438e5ba86fce600
```

```
dotnet --version
6.0.202
```

PR #526, opened 2020-06-03
Adds --watch as an option for tye run

jkotalik:
Fixes https://github.com/dotnet/tye/issues/499.

Currently I went with the simple approach of calling `dotnet watch` on each project.

One thing I was trying to figure out was if we could make it so dotnet watch didn't need to build on first iteration, as we build all projects right beforehand. Passing in --no-build doesn't work though.

I also considered directly implementing dotnet watch in tye itself, however that seems bug prone and I didn't see a huge benefit.


src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -168,6 +168,14 @@ private void LaunchService(Application application, Service service)
             // Set by BuildAndRunService
             var args = service.Status.Args!;
             var path = service.Status.ExecutablePath!;
+
+            if (_options.Watch && service.ServiceType == ServiceType.Project)
+            {
+                path = "dotnet";
+                // Don't use launch settings as we already configured ports earlier.
+                args = $"watch run {service.Status.ProjectFilePath} --no-launch-profile " + args;
```
davidfowl, line 176: This isn't great because watch launches another process no? We want direct tracking of the process being launched. Test what happens when you kill the process.

jkotalik: Yeah you are right, this is tracking the watch process.

jkotalik: Looks like I couldn't get away with the easy path 😄

areller:
Actually thought about this feature yesterday!!

src/Microsoft.Tye.Hosting/ProcessRunner.cs
davidfowl: That's why I didn't do it yet 😄

jkotalik:
> Actually thought about this feature yesterday!!

Beat ya to it 😄 

I don't think it will be horrible to implement the right version of this in Tye though. I'll take a crack at it now.

jkotalik:
Still have a bit of cleanup to do here.

Issue #532, opened 2020-06-08
tye undeploy fails without a tye.yaml


rynowak:
Using `tye undeploy` will fail to find any resources if you don't have a `tye.yaml`

### Repro

Use the `frontend-backend` sample

Delete `tye.yaml`

`tye deploy -i`

`tye undeploy` 

The `tye undeploy` step will fail to delete the k8s resources. 

Now run `tye init` and `tye undeploy` and it will do the thing :)

Issue #518, opened 2020-05-30
Tye undeplo is not working


NileshGule:
### Describe the bug
I have deployed an application using Tye deploy successfully. When I run the Tye undeploy command it fails to undeploy the resources which were deployed to an AKS cluster. I can list the pods which are running using kubectl get pods command

### To Reproduce
Deploy an application services using Tye deploy command

Run Tye Undeploy command. **Tye shows Found 0 resource(s)**

### Further technical details

tye version : 0.3.0-alpha.20273.1+65c09d737e36edd3757b248967db252a7fff9e75
- If possible rerun the command with `-v debug`
<img width="1119" alt="tye pod part of" src="https://user-images.githubusercontent.com/1298374/83327512-11cf5b80-a2af-11ea-8ed6-76eded50a5e9.PNG">
<img width="719" alt="tye undeploy no resources found" src="https://user-images.githubusercontent.com/1298374/83327513-13991f00-a2af-11ea-8452-50d2ae885959.PNG">
<img width="816" alt="tye deploy running pods" src="https://user-images.githubusercontent.com/1298374/83327516-1562e280-a2af-11ea-8ca5-95c8bdc84994.PNG">

 and include the output
- The platform Windows 10

davidfowl:
@jkotalik Did you try undeploy with dockerfile built containers?

NileshGule:
@jkotalik you can find the repo on which I am trying Tye currently here
https://github.com/nileshgule/pd-tech-fest-2019

jkotalik:
Yeah sounds like that is most likely the issue.

jkotalik:
@NileshGule I think the bug is a bit on both sides here. The issue is that there is no application name defined in the tye.yaml. I see a few actions here:
1. Default to the application name, if not defined, to the current directory name.
2. Required that a name is defined in the tye.yaml.

@davidfowl thoughts?

NileshGule:
hi @jkotalik,
It would be better to keep the deploy & undeploy commands in sync. If the tye deploy is able to deploy the application without specifying the application name, then undeploy can use the same convention used while deploying the application to reverse the action.

Other option could be to make the application name mandatory.

jkotalik:
@NileshGule I fixed it and did the first option you suggested.

jkotalik:
Feel free to try the latest nightly build and see if that works.

PR #537, opened 2020-06-09
Fix undeploy

jkotalik:
Fixes https://github.com/dotnet/tye/issues/518
Fixes https://github.com/dotnet/tye/issues/532

Application name didn't have a default in these scenarios. I went ahead and added inferring the application name to many places.

Also verified that undeploy with a dockerfile works.


src/Microsoft.Tye.Core/ConfigModel/NameInferer.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+
+namespace Microsoft.Tye.ConfigModel
+{
+    public static class NameInferer
```
davidfowl, line 9: internal?

PR #551, opened 2020-06-19
Actually call dotnet run with watch

jkotalik:
Prior, we would detect that the project needed to be rebuild, however we didn't actually rebuild the project as we just keep on running the executable. This was due to a bad refactor at the end of the feature.

### DIFF
--- a/src/Microsoft.Tye.Hosting/ProcessRunner.cs
+++ b/src/Microsoft.Tye.Hosting/ProcessRunner.cs
@@ -304,6 +304,8 @@ namespace Microsoft.Tye.Hosting
                         if (_options.Watch && service.Description.RunInfo is ProjectRunInfo runInfo)
                         {
                             var projectFile = runInfo.ProjectFile.FullName;
+                            processInfo.Executable = "dotnet";
+                            processInfo.Arguments = $"run {service.Status.ProjectFilePath} --no-launch-profile {processInfo.Arguments}";
 
                             var fileSetFactory = new MsBuildFileSetFactory(_logger,
                                 projectFile,
--- a/src/Microsoft.Tye.Hosting/ProcessRunner.cs
+++ b/src/Microsoft.Tye.Hosting/ProcessRunner.cs
@@ -368,7 +368,7 @@ namespace Microsoft.Tye.Hosting
                         }
                     }
 
-                  
+
                 }
             }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -298,13 +298,16 @@ async Task RunApplicationAsync(IEnumerable<(int ExternalPort, int Port, string?
 
                                 WriteReplicaToStore(pid.ToString());
                                 service.ReplicaEvents.OnNext(new ReplicaEvent(ReplicaState.Started, status));
+                            },
+                            Build = async () =>
+                            {
+                                var buildResult = await ProcessUtil.RunAsync("dotnet", $"build \"{service.Status.ProjectFilePath}\" /nologo", throwOnError: false, workingDirectory: application.ContextDirectory);
```
davidfowl, line 304: What happens on failure? Watch has this state machine to handle that case. Is that handled here?

test/E2ETest/TyeRunTests.cs
```diff
@@ -291,7 +291,7 @@ public async Task FrontendBackendWatchRunTest()
                         return;
                     }
 
-                    await Task.Delay(500);
+                    await Task.Delay(5000);
```
davidfowl, line 294: Why?

jkotalik: test was failing on CI, pretty much need to wait for dotnet build which apparently takes longer than 5 seconds

davidfowl: Yep!! Or we could wait for the replica to be up again after going down.

jkotalik: Yeah, potentially if this is flaky.

src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -317,15 +362,7 @@ async Task RunApplicationAsync(IEnumerable<(int ExternalPort, int Port, string?
                         else
                         {
                             var result = await ProcessUtil.RunAsync(processInfo, status.StoppingTokenSource.Token, throwOnError: false);
```
davidfowl, line 364: The result here is no longer used.

Issue #564, opened 2020-06-26
[SILENT ERROR] Tye run does not supports for CsProj containing "TargetFrameworkS"


tebeco:
### Describe the bug
If you are trying to do `tye run` and one of the `xxproj`  is using `TargetFrameworks` it will silently crash on restore/build
And `tye run` output an error about a "Missing file"
Few issue here:
* We might want to know the name of the file missing
* We might want to know that the `dotnet xxx /t:....` failed ==> [Cannot find Target ResolveReferences](https://github.com/dotnet/tye/blob/e368d04e914a3880a1a10e21cf5c271966241dbd/src/Microsoft.Tye.Core/ProjectReader.cs#L208-L218)
* We might want to be able to do `tye run --framework netcoreapp3.1` (see investigation)
* We might want to be able to do in `tye.yaml`
```yaml
name: THE_NAME
services:
- name: THE_SERVICE
  project: src/THE_SERVICE/THE_SERVICE.csproj
  framework: netcoreapp3.1
```

### To Reproduce
DISCLAMER:
I did all the investigation on a CORP project, so i'm not entierlly sure this repro is enought
Naive repro:
* `dotnet new webapi`
* replace `TargetFramework` by `TargetFrameworks`
* `tye init`
* `tye run`

### Further technical details
- Include the output of `tye --version`
```
[12:24]❯ tye --version
0.3.0-alpha.20318.3+224101f108311b4d38b226eed75d0ec03fa14db1
```
- The platform: `Windows`

If the repro is not enought, I believe that this one might be very similar to CORP repo:
```
mkdir repro
cd repro
git clone https://github.com/dotnet-experimentations/CentrallyManagingNuGetPackageVersions/
cd CentrallyManagingNuGetPackageVersions
cd WithMsBuild   # <==== THIS IS IMPORTANT :D
tye init
tye run
```

```
[12:23:23 INF] Executing application from C:\dev\xxxxxxx\tye.yaml
[12:23:23 INF] Dashboard running on http://127.0.0.1:8000
[12:23:23 INF] Building projects
[12:23:42 INF] Launching service XXXXXXXXXXXXXXXXXX-aspnetcore_f992c4c5-d:
[12:23:42 INF] Launching service YYYYYYYYYYYYYYYYY_5a5c648c-6:
[12:23:42 ERR] Failed to launch process for service XXXXXXXXXXXXXXXXXX-aspnetcore_f992c4c5-d
[...]

PR #567, opened 2020-06-27
Handling Multiple TargetFrameworks through BuildProperties

tebeco:
fixes https://github.com/dotnet/tye/issues/564

## Breaking
**/!\ `-f` has been REMOVED from `--force`**
 
## Reason
Try to address https://github.com/dotnet/tye/issues/564

# Implementation
As I did not wanted to re-invent the wheel ... and MsBuild, I tried to re-use `MsBuild Props` for that
 - [x] Add a `--framework | -f` to `RunCommandArgs`
 - [x] Add a `--framework | -f` to `BuildCommandArgs`
 - [x] Add a `--framework | -f` to `DeployCommandArgs`
 - [x] Add a `--framework | -f` to `UnDeployCommandArgs`
 - [x] Add a `--framework | -f` to `PushCommandArgs`
 - [x] Add a `--framework | -f` to `GenerateCommandArgs`
 - [x] Use the existing `buildProperties` yaml list for `tye.yaml`
 - [x] Map it to `BuildProperties`
 - [x] Don't throw anymore if a project is using plural form `TargetFrameworks` AND one is explicitly specified as a `BuildProperty`
 - [x] Make `--framework | -f ` a `StandardOptions`
 - [x] Change `-f` from `--force` to `--framework`

# Testing
## ApplicationHost
- [x] [Should pass] `tye.yaml` without explicit `buildProperties` set but one for CLI
- [x] [Should pass] `tye.yaml` with explicit `buildProperties` beeing overriden but one for CLI
- [x] [Should Fail] `tye.yaml` without `buildProperties`
- [x] [Should Fail] `tye.yaml` with a `TargetFramework` that is not in the existing list of `TargetFrameworks`

## tye run
- [x] [Should pass] `tye.yaml` with explicit `buildProperties` set
- [x] [Should pass] `tye.yaml` without `buildProperties` but passed to `ApplicationFactory` (the CLI is supposed to pass it down to this API)
- [x] [Should pass] `csproj` contains both `TargetFramework` and `TargetFrameworks` without `buildProperty`

## tye build
- [x] [Should pass] `tye.yaml` with explicit `buildProperties` set
- [x] [Should pass] passing a `framework` as CLI would do

## Testing questions
* I'm not sure I see test for `tye push`
* I'm not sure I can test properly for `tye generate` (hidden command)
* I'm not sure I can test properly for `tye deploy | undeploy` (needs a Regsitry ?)

### How do I test cli argument ?
I'm not supposed to test `System.CommandLine` as it's not this "product" though i would have been able to test CLI overring YAML
* [Should pass] `tye.yaml` without `buildProperties` but `-f`
* [Should pass] `tye.yaml` without `buildProperties` but `-framework`
* [Should pass] `-f` should override `buildProperties` from `tye.yaml`


jkotalik:
@tebeco how is this going?

Overall, we will accept this change if it is ready for review.

tebeco:
Did a break on it any suggestion for all these null ?
also is the underlying idea good ?

tebeco:
@jkotalik 
Trying to get back on it, did I missed something in the contribution docs about E2E, I got ~30+ test failling about
`An attempt was made to access a socket in a way forbidden by its access permissions.`

Hope I don't have to run VS as an `Administrator` :s 
(I see it's passing here)

tebeco:
@jkotalik 
I'm having weird issue ... after sometime "something" change my computer state, and i'm globally receiving `premission denied` on the `Socket.Bind` ... for ... unknow reason, I'm going crazy since my only known fix is a reboot.

I hope it's not an AntiCheat from a game doing that, or the current `Dev Insider` win10 build being unstable

samples/app-with-targetframeworks/MultipleTargetFrameworks/MultipleTargetFrameworks.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.1;netcoreapp2.2</TargetFrameworks>
```
jkotalik, line 4: I'd recommend using 2.1 instead of 2.2 here as 2.2 is out of support.

src/tye/ApplicationBuilderExtensions.cs
```diff
@@ -122,7 +123,17 @@ public static Application ToHostingApplication(this ApplicationBuilder applicati
...
+                        if (project.BuildProperties.TryGetValue("TargetFramework", out var targetFramework))
+                        {
+                            if (!project.TargetFrameworks.Contains(targetFramework))
+                            {
+                                throw new InvalidOperationException($"Unable to run {project.Name}. The specified TargetFramework is not one of the existing TargetFrameworks in the project.");
+                            }
+                        }
+                        else
+                        {
+                            throw new InvalidOperationException($"Unable to run {project.Name}. Multi-targeted projects are not supported.");
```
jkotalik, line 135: Change this exception message to say "specify a target framework in your tye.yaml or with --framework"

src/tye/Program.UndeployCommand.cs
```diff
@@ -24,24 +25,58 @@ public static Command CreateUndeployCommand()
                 StandardOptions.Verbosity,
                 StandardOptions.Tags,
 
+                new Option(new string[]{"-f", "--framework" })
```
jkotalik, line 28: You can put this in the StandardOptions instead of duplicating it for all commands.

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Tye
 {
     public static class ApplicationFactory
     {
-        public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, FileInfo source, ApplicationFactoryFilter? filter = null)
+        public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, FileInfo source, string? framework, ApplicationFactoryFilter? filter = null)
```
jkotalik, line 17:
```suggestion
        public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, FileInfo source, string? framework = null, ApplicationFactoryFilter? filter = null)
```

jkotalik:
I see a sample, but are there any functional tests?

samples/app-with-targetframeworks/MultipleTargetFrameworks/MultipleTargetFrameworks.csproj
tebeco: oO my bad, forgot to change that before pushing ^^

src/tye/Program.UndeployCommand.cs
tebeco: I though about that but there's a small issue 
there's 2 Command that uses `-f` for `--force` 
do I simply remove `-f` for `--framework` ?

PS:
one of the 2 command has it declared / passed down but never use it
should i also remove it  ?

src/Microsoft.Tye.Core/ApplicationFactory.cs
tebeco: do i report this for all commands ?

tebeco:
> I see a sample, but are there any functional tests?

Joker here, i changed to "Ready for review" ASAP to get first feedback on the PR
Tomorrow is National Day here so I'll get some time to investigate that
What kind on "Assert" are you expected ?
pass and throw on "overall setup" ?

* 2 TFs and -f / --framework => ok
* 2 TFs => fail
* 2 TFs + TF => ok
* 2 TFs + `buildProperties` => ok
* 2 TFs + `-f other` <== ???

src/Microsoft.Tye.Core/ApplicationFactory.cs
tebeco: (un resolving so that i don't forget)

samples/app-with-targetframeworks/MultipleTargetFrameworks/MultipleTargetFrameworks.csproj
tebeco: (done)

src/tye/ApplicationBuilderExtensions.cs
tebeco: I'll repro the same error with `dotnet run` and copy paste the error msg
is that fine for you ?

jkotalik: Yep.

src/tye/Program.UndeployCommand.cs
jkotalik: Bleh, that unfortunate. @davidfowl @JunTaoLuo should we change the -f universally to be --framework and change --force to something else?

JunTaoLuo: I think it makes sense to use -f for --framework just to keep in line with the rest of dotnet commands (like dotnet run). How about leaving --force without a short option?

tebeco: would prefer `--force` rather than `-force` for consistency over all CLI in the `dotnet` echo system TBH

src/tye/ApplicationBuilderExtensions.cs
tebeco: done

src/Microsoft.Tye.Core/ApplicationFactory.cs
tebeco: done

src/tye/Program.UndeployCommand.cs
tebeco: done but it would require a review:
* I moved all `ForceOptions` to StandardOptions.CreateForce("description here")`
* There's no `-f` anymore on that `CreateForce), so it kinda break few commands which might not be ideal

The idea would to be able to check if there's already `Standard.Framework` and depending on it use `-f` for `--force`
Doing this would have a bad side effect if in some "month/year", for ... reason `--framework` appear, then the `-f` would change from `force` to `framework`

tebeco:
(starting to work on tests)

tebeco:
@JunTaoLuo @jkotalik 
Just to be sure, can I have some feedback on the current scope of this PR ?
I'm slowly adding E2E testing, but I'm not sure how far you think test should/could go.

Also now that I see 33 commits inside this PR I would also warn before it goes too wild

tebeco:
@jkotalik @JunTaoLuo 
Will stop until feedback not sure what's missing / what to do / PR might go wild

* deleted all my previous comment
* added more test 
* rebased on `0.4.0`

as some part of this feature may be tricky to test (or are not tested yet)

tebeco:
hello @jkotalik 

Do you want me to rebase the branch on `main` ? I see the associated issue had been moved to the `backlog`, not sure what else I can do to help

tebeco:
Just rebased on default branch to solve conflicts

jkotalik:
We approve of the yaml change to move -f from force to framework.

src/tye/Program.DeployCommand.cs
```diff
@@ -132,5 +130,24 @@ internal static void ApplyRegistry(OutputContext output, ApplicationBuilder appl
                 // No registry specified, and that's OK!
             }
         }
+
+        public class DeployCommandArguments
```
JunTaoLuo, line 134: Make this a private class.

src/tye/Program.InitCommand.cs
```diff
@@ -20,29 +20,34 @@ private static Command CreateInitCommand()
...
-
-                console.Out.WriteLine($"Time Elapsed: {elapsedTime.Hours:00}:{elapsedTime.Minutes:00}:{elapsedTime.Seconds:00}:{elapsedTime.Milliseconds / 10:00}");
+                var elapsedTime = watch.Elapsed;
+                output.WriteInfoLine($"Time Elapsed: {elapsedTime.Hours:00}:{elapsedTime.Minutes:00}:{elapsedTime.Seconds:00}:{elapsedTime.Milliseconds / 10:00}");
             });
 
             return command;
         }
+
+        public class InitCommandArguments
```
JunTaoLuo, line 42: Private

src/tye/Program.UndeployCommand.cs
```diff
@@ -23,25 +24,50 @@ public static Command CreateUndeployCommand()
...
+
+                var filter = ApplicationFactoryFilter.GetApplicationFactoryFilter(args.Tags);
+
+                return UndeployHost.UndeployAsync(output, args.Path, args.Namespace, args.Interactive, args.WhatIf, args.Framework, filter);
             });
 
             return command;
         }
+
+        public class UndeployCommandArguments
```
JunTaoLuo, line 54: Private

test/E2ETest/testassets/projects/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
JunTaoLuo, line 4: projects shouldn't specify both `TargetFramework` and `TargetFrameworks` at the same time. I don't think testing such a scenario is needed.

test/E2ETest/ApplicationFactoryTests.cs
```diff
@@ -27,11 +29,12 @@ public async Task CyclesStopImmediately()
 - name: test-project
   include: tye.yaml";
             var yamlFile = Path.Combine(projectDirectory.DirectoryPath, "tye.yaml");
-
             await File.WriteAllTextAsync(yamlFile, content);
+
             // Debug targets can be null if not specified, so make sure calling host.Start does not throw.
             var outputContext = new OutputContext(_sink, Verbosity.Debug);
-            var application = await ApplicationFactory.CreateAsync(outputContext, new FileInfo(yamlFile));
+            var projectFile = new FileInfo(yamlFile);
```
JunTaoLuo, line 36: What's the point of this change?

test/E2ETest/TyeRunTests.cs
```diff
@@ -871,10 +873,13 @@ public async Task DockerFileChangeContextTest()
 - name: frontend
   project: frontend/frontend.csproj";
 
-            var projectFile = Path.Combine(projectDirectory.DirectoryPath, "tye.yaml");
-            await File.WriteAllTextAsync(projectFile, content);
+
```
JunTaoLuo, line 876: nit: extra line

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -98,6 +98,13 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
                         {
                             project.BuildProperties.Add(buildProperty.Name, buildProperty.Value);
                         }
+
+                        if (framework != null)
+                        {
+                            // Only use the TargetFramework for the "--framework" if it's not defined already from the YAML
```
JunTaoLuo, line 104: Is this comment accurate? I believe the cli options wins over yaml.

src/tye/Program.DeployCommand.cs
tebeco: done, thx

src/tye/Program.UndeployCommand.cs
tebeco: done, thx

src/tye/Program.InitCommand.cs
tebeco: done, thx

test/E2ETest/testassets/projects/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too.csproj
tebeco: As I understand it's weird to read, this is how `MsBuild` and `dotnet` works today
So I tried to cover existing "possible scenario"
This is also how `dotnet/tye` works in `tye.yaml` when using `buildProperties` for `TargetFramework` if the csproj uses `TargetFrameworks`
What should I do ? Do you want me to remove this sample anyway ?

eg:
image

test/E2ETest/TyeRunTests.cs
tebeco: oupsie, fixed

test/E2ETest/ApplicationFactoryTests.cs
tebeco: Do you want me to rollback that change ?

src/Microsoft.Tye.Core/ApplicationFactory.cs
tebeco: oOo I forgot that one, I think it's a "not funny" one ╚(•⌂•)╝
It does seems obvious that a "CLI args" would "win" overall 

if `tye.yaml` have 2 projects
  * 1st in csproj use single `TargetFramework>net5.0`
  * 2nd in csproj use multi `TargetFrameworks>net5.0;netcoreapp3.1`
  * what should do `tye run --framework netcoreapp3.1` ? this would throw, and I'm fine with that, just want to be sure


I'll have to debug that again (was a long time ago), I think the was the `string? framework` is computed there was no way to differentiate if the value was passed down from the `yaml` or from the `CLI` especially when the scenario was the previous one

This lead me to also write the `e2e` you asked about :
https://github.com/dotnet/tye/pull/567#discussion_r481923444

I took the approach of the `null check` since that was the only info I had at this very place in code.
I'll try to re-debug it again to be sure of all of this, If you have any change suggestion on that one it will be more than welcome

test/E2ETest/testassets/projects/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too.csproj
JunTaoLuo: So yes and no, for example take a look at the docs https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-target-frameworks. The idea is that you can target multiple frameworks in your project by using `TargetFrameworks`. This is especially important for libraries which might need to maintain compatibility with older frameworks but also take advantage of new APIs but can also be done for apps like you've demonstrated. However, if you do multi-target apps you are essentially building for both TFMs and you will need to specify a specific target framework to run. The mechanism to do this is to specify a single `TargetFramework`, for example via CLI. However, it doesn't make sense to do so in a csproj file since if you are specifying a single `TargetFramework` at build time, why even bother defining multiple frameworks in the first place?

test/E2ETest/ApplicationFactoryTests.cs
JunTaoLuo: Refactoring is fine, and I do see where this change is necessary for tests to work. But in this particular test and several like it, there's no reason to split the instantiation of the project file into a separate line. Usually we try to avoid unnecessary churn in the code, but given it doesn't affect and functionality, I'm okay with keeping this change. Though I would say in general, if there's no functional benefits, try to avoid churn.

test/E2ETest/testassets/projects/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too.csproj
tebeco: sorry, as I tried to review your comments here, I know realize that my answer is split in two with 
https://github.com/dotnet/tye/pull/567#discussion_r482027153

Also this is also practical aside library, especially when you are migrating from `netcoreapp3.1` to `net5.0` for example it allows you to have both code build / test / E2E and decide which artifact you will ship to production (or even both for canary or blue/green)

Of course the `--framework` is, defacto, mandatory for such use, but it's not limited to library. I did not tried to overthink it to be fair, I just tried to make sure the behavior of `tye` was somehow the same as `dotnet`.
I'm fine with keeping or removing it, but I guess this will depends on the other comment bellow

test/E2ETest/ApplicationFactoryTests.cs
tebeco: noted, so far I'll keep it like this just tell me, anytime in this PR lifetime if you want it inlined

src/Microsoft.Tye.Core/ApplicationFactory.cs
JunTaoLuo: Hmmm this brings up a good question, if someone specifies --framework, should it apply to all projects or just projects that are multi-targeted? I find it kind of strange that if you run `tye run --framework netcoreapp3.1` you might still end up with services running on `net5.0` but I guess that is probably the most convenient. The alternative is to force the user to specify the TFM in tye.yaml instead, which is okay as well but makes it really not much different than just picking a particular TFM in the csproj directly (i.e. why bother with TargetFrameworks in the first place).

What do you think @jkotalik @glennc ?

tebeco: there's also this :
```xml
<TargetFrameworks>netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
```

`tye run --framework net5.0`

if you try this with `dotnet run` IIRC, this will fail

FUNNY NOTE:
```xml
<TargetFramework>net5.0</TargetFramework>
<TargetFrameworks>netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
```
This is why this `if` is here

how should these behave ?
`try run`
`tye run --framework net5.0`
`tye run --framework netcoreapp3.1`
`tye run --framework net6.0`

tebeco: for info:
not declared look like a failure today
image

tebeco: but this looks "okish"
image

tebeco: That being said and to be consistent I could try to change and force the framework everywhere and make it fails it there's no match
that's totally fine to either "force all" from CLI or specifics goes in YAML
that will avoid "magic/question"

will start to work on that just in case

src/tye/UndeployHost.cs
```diff
@@ -17,16 +17,12 @@ namespace Microsoft.Tye
 {
     public static class UndeployHost
     {
-        public static async Task UndeployAsync(IConsole console, FileInfo path, Verbosity verbosity, string @namespace, bool interactive, bool whatIf, string[] tags)
+        public static async Task UndeployAsync(OutputContext output, FileInfo path, string @namespace, bool interactive, bool whatIf, string? framework = null, ApplicationFactoryFilter? filter = null)
         {
             var watch = System.Diagnostics.Stopwatch.StartNew();
 
-            var output = new OutputContext(console, verbosity);
+            var application = await ApplicationFactory.CreateAsync(output, path, framework, filter);
```
JunTaoLuo, line 24: On a second look, it looks like we don't need to add the framework option to Undeploy at all. The other commands run a build/publish step which will need a framework to be specified. However for undeploy, only the applicationname is resolved from the application. We can just pass null for the framework here.

test/E2ETest/testassets/projects/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too.csproj
JunTaoLuo: As I said previously,
> but can also be done for apps like you've demonstrated
I'm agreeing with you that there is valid use of multi-targeted apps. The comment about libraries is more of a side note. My main point is that it doesn't make sense to specify both TargetFramework and TargetFrameworks in an app. You either specify one or the other.

test/E2ETest/ApplicationFactoryTests.cs
JunTaoLuo: No need, I'm mainly adding the note for when you make contributions in the future! 😄

src/Microsoft.Tye.Core/ApplicationFactory.cs
JunTaoLuo: A few observations here:

1. Again, the scenario where TargetFramework and TargetFrameworks are both specified is not really valid. Don't dwell on making this "work". Let me know if you want me to elaborate on this further and explain why some of the scenarios you mentioned above "worked" and some "didn't work". It has to do with how the three steps of restore, build and run handle TargetFramework and TargetFrameworks.
2. I think it makes sense that the option `--framework` should only apply to multi-targeted apps only. This magic seems reasonable to me, and the alternative of forcing every service obey `--framework` seems even more troublesome. I'm okay with this behaviour as long as we add a test for it and it's more clear in the option description.

tebeco: for `2.` i'll wait for @jkotalik and @glennc answer so we got all the input before going wild on the branch

personal opinion:
current implementation was the easiest at the moment I did the PR and also seems logic for this very specific scenario
But after a bit of time I'm wondering if it should not just apply to every single project.
Why ?
well this would be kinda like `dotnet build` or `dotnet sln add` will throw if there's a conflict where the tools would solve that "conflict" on it own with an arbitrary logic.

I'll wait for a consensus and see if I got time to prepare the code for the "apply all" just in case

tebeco:
friendly ping to @jkotalik @glennc about the question raised previously in the review

src/tye/UndeployHost.cs
tebeco: I suppose I also remove this from the Signature ?
```csharp
public static async Task UndeployAsync(OutputContext output, FileInfo path, string @namespace, bool interactive, bool whatIf, ApplicationFactoryFilter? filter = null)
```

JunTaoLuo: Yup the change looks good.

tebeco: done

tebeco:
sorry to ping back again on that @jkotalik @glennc 
could you take a look at that thread ?
https://github.com/dotnet/tye/pull/567#discussion_r482436710

src/Microsoft.Tye.Core/ApplicationFactory.cs
JunTaoLuo: Ah I forgot to update this thread but we met last week and agreed that the proposal "the option --framework should only apply to multi-targeted apps only" is preferred, so let's continue with that in mind.

tebeco:
@JunTaoLuo 
I think that since MsBuild is now running "Out of proc" the `BuildProperties` does not list anymore the `TargetFrameworks` from the `csproj` itself and not only list thing from the YAML
I'm not sure at all how to proceed with [this suggestion](https://github.com/dotnet/tye/pull/567#discussion_r496311231) yet

As I'm now blind ... should I just apply the framework blindly ? that would work I suppose

JunTaoLuo:
Hmm good point, some of this work might have changed given the MsBuild out of proc change I made. However, all the properties are still there, there's just a few more hoops to jump through to get them. I'll take a look and see if there's something we can do.

tebeco:
I'm hitting a strange MsBuild error on the all test related the this PR so i'll compare if something changed in the test "bootstrapping"
So far the "visible" side effect is that `ApplicationFactory.CreateAsync()` fails on

```cs
                    var msbuildEvaluationResult = await ProcessUtil.RunAsync(
                        "dotnet",
                        $"build " +
                            $"\"{projectPath}\" " +
                            $"/p:CustomAfterMicrosoftCommonTargets={Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!, "ProjectEvaluation.targets")} " +
                            $"/nologo",
                        throwOnError: false,
                        workingDirectory: directory.DirectoryPath);
```

with `stdout` :
```
C:\Users\foo\AppData\Local\Temp\0kgcti2m.jbl\multi-targetframeworks\multi-targetframeworks.csproj : error MSB4057: The target "MicrosoftTye_GetProjectMetadata" does not exist in the project.

Build FAILED.

C:\Users\foo\AppData\Local\Temp\0kgcti2m.jbl\multi-targetframeworks\multi-targetframeworks.csproj : error MSB4057: The target "MicrosoftTye_GetProjectMetadata" does not exist in the project.
    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:00.21
```

i've found out that the file `$"/p:CustomAfterMicrosoftCommonTargets={Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!` properly exists and contains this:
```xml
<Project>
    <Target Name="MicrosoftTye_GetProjectMetadata" DependsOnTargets="Restore;ResolveReferences;ResolvePackageDependenciesDesignTime;PrepareResources;GetAssemblyAttributes" >
        <PropertyGroup>
            <_MicrosoftTye_MetadataFile>$([System.IO.Path]::GetFullPath('$(IntermediateOutputPath)MicrosoftTye.ProjectMetadata.txt'))</_MicrosoftTye_MetadataFile>
            <_MicrosoftTye_ProjectFrameworkReference>@(FrameworkReference, '%3B')</_MicrosoftTye_ProjectFrameworkReference>
        </PropertyGroup>

        <ItemGroup>
            <_MicrosoftTye_ProjectMetadata Include="AssemblyInformationalVersion: $(AssemblyInformationalVersion)" />
            <_MicrosoftTye_ProjectMetadata Include="InformationalVersion: $(InformationalVersion)" />
            <_MicrosoftTye_ProjectMetadata Include="Version: $(Version)" />
            <_MicrosoftTye_ProjectMetadata Include="TargetFrameworks: $(TargetFrameworks)" />
            <_MicrosoftTye_ProjectMetadata Include="RunCommand: $(RunCommand)" />
            <_MicrosoftTye_ProjectMetadata Include="RunArguments: $(RunArguments)" />
            <_MicrosoftTye_ProjectMetadata Include="TargetPath: $(TargetPath)" />
            <_MicrosoftTye_ProjectMetadata Include="PublishDir: $(PublishDir)" />
            <_MicrosoftTye_ProjectMetadata Include="AssemblyName: $(AssemblyName)" />
            <_MicrosoftTye_ProjectMetadata Include="IntermediateOutputPath: $(IntermediateOutputPath)" />
            <_MicrosoftTye_ProjectMetadata Include="TargetFramework: $(TargetFramework)" />
            <_MicrosoftTye_ProjectMetadata Include="_ShortFrameworkIdentifier: $(_ShortFrameworkIdentifier)" />
            <_MicrosoftTye_ProjectMetadata Include="_ShortFrameworkVersion: $(_ShortFrameworkVersion)" />
            <_MicrosoftTye_ProjectMetadata Include="_TargetFrameworkVersionWithoutV: $(_TargetFrameworkVersionWithoutV)" />
            <_MicrosoftTye_ProjectMetadata Include="FrameworkReference: $(_MicrosoftTye_ProjectFrameworkReference)" />
            <_MicrosoftTye_ProjectMetadata Include="ContainerBaseImage: $(ContainerBaseImage)" />
            <_MicrosoftTye_ProjectMetadata Include="ContainerBaseTag: $(ContainerBaseTag)" />
            <_MicrosoftTye_ProjectMetadata Include="MicrosoftNETPlatformLibrary: $(MicrosoftNETPlatformLibrary)" />
            <_MicrosoftTye_ProjectMetadata Include="_AspNetCoreAppSharedFxIsEnabled: $(_AspNetCoreAppSharedFxIsEnabled)" />
            <_MicrosoftTye_ProjectMetadata Include="UsingMicrosoftNETSdkWeb: $(UsingMicrosoftNETSdkWeb)" />
        </ItemGroup>

        <WriteLinesToFile
            File="$(_MicrosoftTye_MetadataFile)"
            Lines="@(_MicrosoftTye_ProjectMetadata)"
            Overwrite="true"
            WriteOnlyWhenDifferent="true"/>

        <Message Text="Microsoft.Tye metadata: $(MicrosoftTye_ProjectName): $(_MicrosoftTye_MetadataFile)" Importance="High"/>

    </Target>
</Project>
```

wyhich contains the `MicrosoftTye_GetProjectMetadata` targets ... I have no idea yet what's going on

JunTaoLuo:
I took a deeper look and it seems like the logic at https://github.com/dotnet/tye/blob/bbef222754e024e10d3df0f456d450eeb05a85ab/src/Microsoft.Tye.Core/ApplicationFactory.cs#L127 doesn't quite work for multi-targeted projects.

Specifically, it seems like imports for multi-targeting projects rely on `CustomAfterMicrosoftCommonCrossTargetingTargets` instead of `CustomAfterMicrosoftCommonTargets`. However, just switching to that property doesn't resolve the issue due to some nuances between inner build and outer build. I'll need to look into this further.

@jkotalik I think this would block 0.5.0 since it seems like msbuild out of proc would break all multi-targeted projects.

tebeco:
I also noticed that the change I did in the PR might have to be re-written, It look like to have an "certain" impact over lots of E2E testing (probably because of that custom `Props`

As I had trouble understanding it, I went to discord `Dotnet Evolution` (https://aka.ms/dotnet-discord)
There's both a `tye` channel and an MsBuild channel

I got help on the MsBuild part, but I would love that someone take a look at this discussion here:
https://discordapp.com/channels/732297728826277939/732310858776182866/764579073799880724

so my side question is:
As `dotnet/tye` use `Directory.Build.props`, why not use `Directory.Build.targets` ? (which supports multi TFM)

JunTaoLuo:
FYI I took a stab at fixing the issue, turns out to be simpler than I expected. Can you try incorporating the change https://github.com/tebeco/tye/pull/1? I can't directly push to your branch so I opened a PR 😄 .

tebeco:
> I can't directly push to your branch so I opened a PR 😄

I'll take a look to be sure I enabled maintainer to push to branch. I usually make sure it's possible.
Sorry for the inconvenience

JunTaoLuo:
D'oh looks like there are a few more test failures I missed. Btw it's no problem, I can just open PRs since I'm not sure if there's an option to allow maintainers to push to your fork.

tebeco:
Easiest way for me was to send you an invite to my fork directly as a collaborator

tell me if it gave you access properly ?

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -125,6 +128,7 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
                         $"build " +
                             $"\"{projectPath}\" " +
                             $"/p:CustomAfterMicrosoftCommonTargets={Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!, "ProjectEvaluation.targets")} " +
+                            $"/p:CustomAfterMicrosoftCommonCrossTargetingTargets={Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!, "ProjectEvaluation.targets")} " +
```
JunTaoLuo, line 131: This change is only for producing a better error message. Theoretically, multi-tfm projects must have a TFM specified for evaluation. However, without evaluating a multi-tfm project without an explicit TFM, there's no way to even know that a project is multitargeted in the first place. I've also updated ProjectEvaluation.targets to handle a multi-tfm project and at least be able to resolve the `TargetFrameworks` property so the rest of the logic would be able to tell it's a multi-targeted project.

src/tye/ProjectEvaluation.targets
```diff
@@ -1,15 +1,21 @@
 ﻿<Project>
-    <Target Name="MicrosoftTye_GetProjectMetadata" DependsOnTargets="Restore;ResolveReferences;ResolvePackageDependenciesDesignTime;PrepareResources;GetAssemblyAttributes" >
+    <PropertyGroup>
+        <MicrosoftTye_GetProjectMetadata_DependsOn>Restore;ResolveReferences;ResolvePackageDependenciesDesignTime;PrepareResources;PrepareResources;GetAssemblyAttributes</MicrosoftTye_GetProjectMetadata_DependsOn>
+        <MicrosoftTye_GetProjectMetadata_DependsOn Condition="'$(IsCrossTargetingBuild)' == 'true'">Restore</MicrosoftTye_GetProjectMetadata_DependsOn>
```
JunTaoLuo, line 4: For reference, the other targets such as `GetAssumblyAttributes` are not available for a cross-targeting build (i.e. the outer build) that occurs during the first evaluation of a multi-targeted project.

test/E2ETest/testassets/projects/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too/multi-targetframeworks-and-targetframework-too.csproj
JunTaoLuo: Just to circle back, the `TargetFramework` property should be removed here.

test/E2ETest/TyeRunTests.cs
```diff
@@ -900,6 +904,130 @@ public async Task DockerFileChangeContextTest()
...
+                var responseContent = await backendResponse.Content.ReadAsStringAsync();
+                Assert.Contains(".NET Core 3.1", responseContent);
+            });
+        }
+
+        [ConditionalFact]
+        [SkipIfDockerNotRunning]
+        public async Task RunIgnorePluralTargetFrameworksWhenSingularFormExistsTest()
+        {
+            using var projectDirectory = CopyTestProjectDirectory("multi-targetframeworks-and-targetframework-too");
```
JunTaoLuo, line 1004: This scenario isn't valid, let's remove the test and the test project.

src/Microsoft.Tye.Core/ApplicationFactory.cs
JunTaoLuo: Does the logic here need to be updated to ensure we only override the TargetFramework when it's a multi-targeted app?

src/Microsoft.Tye.Core/StandardOptions.cs
```diff
@@ -46,16 +46,16 @@ public static Option Tags
...
-                return new Option("--force", "Force overwrite of existing files")
-                {
-                    Argument = new Argument<bool>(),
-                };
-            }
-        }
+        public static Option Framework =>
+                    new Option(new string[] { "-f", "--framework" })
+                    {
+                        Description = "The target framework to run for. The target framework must also be specified in the project file.",
```
JunTaoLuo, line 52: We should mention that this option is only applied to multi-targeted projects.

JunTaoLuo: We should also mention that this option overrides the TFM specified in the tye.yaml for multi-targeted projects. Essentially we should be clear that this option is used as the TFM for all multi-targeted projects in the tye-yaml specification.

JunTaoLuo:
I think we are very close to getting this merged. Can you make sure that the TFM override is only applied to multi-tfm projects and not single-tfm projects? Let's also add a test for that.

src/Microsoft.Tye.Core/ApplicationFactory.cs
tebeco: ok so now I see this I'm kinda sure i would probably not found out this ^^

test/E2ETest/TyeRunTests.cs
tebeco: Yeah I remember we spoke about that.
Do you want me to remove it or di you want to do it ?

src/Microsoft.Tye.Core/StandardOptions.cs
tebeco: you mean chaing the `Description` ?
I'm usually very bad for such thing, If you have any idea internally (I mean with the maintaining team, feel free to push an update on that part)

tebeco:
> I think we are very close to getting this merged. Can you make sure that the TFM override is only applied to multi-tfm projects and not single-tfm projects? Let's also add a test for that.

fair enought, thx a lot for the time you took on this PR

src/tye/ProjectEvaluation.targets
tebeco: I have to be honest, I have no idea if I have something to do / change with that statement now ^^

src/Microsoft.Tye.Core/ApplicationFactory.cs
JunTaoLuo: No problem 😄 it's not something all that easy to track down. I'm adding a comment mostly for information purposes. Now I think about it, I should add some comments.

src/tye/ProjectEvaluation.targets
JunTaoLuo: No worries, I just included some comments in case you were wondering what this is for.

src/Microsoft.Tye.Core/StandardOptions.cs
JunTaoLuo: Okay I'll write up something.

JunTaoLuo:
I'll merge the PR once you add the test. We're so close!

src/Microsoft.Tye.Core/ApplicationFactory.cs
tebeco: can't agree more

JunTaoLuo:
@tebeco Thanks for the contribution! I ended up adding the test so we can merge this since we are thinking of another release soon and we needed a few of the fixes in this PR.

tebeco:
yeah i réalisé i procrastinated a bit on that
(dense week ^^)

I apologize for making it on you :(
I'm terribly sorry if that delayed the 0.5.x :s

I can't thank you enough for all the effort and the help you provided

PR #568, opened 2020-06-30
Add tye.yml VS Solution Items on init

### LINKED ISSUE #372: Should we add tye.yaml as a solution item when running `tye init`
This seems like it would save someone a step or two when they are using VS.

spboyer:
Closes #372 

- Adds tye.y[a]ml file to sln files for VS users.
- NoOps if there are no sln files
- if multiple sln files are found assumes the first


tebeco:
Out of curiosity
when doing `dotnet build|run|...` it fails if there's multiple match
From a developer point of view "First sln match" is
* not always going to be the expected one
* tampering with unwanted sln

I understand that it should not "fail"
Maybe a warning
> we did not edited any solution file as ... multiple sln founds

jkotalik:
Agree with @tebeco here, I think we should not add it to any slns when multiple.

src/tye/InitHost.cs
```diff
@@ -105,5 +110,30 @@ private static void ThrowIfTyeFilePresent(FileInfo? path, string yml)
...
             }
         }
+
+        private static void AddToSolutionItems(FileInfo? path)
+        {
+            var fileGuid = Guid.NewGuid().ToString("B").ToUpper();
+
+            var slnFiles = Directory.GetFiles(path!.DirectoryName, "*.sln");
+            // noop if there are no sln files
+            if (slnFiles.Length > 0)
```
jkotalik, line 120:
```suggestion
            if (slnFiles.Length == 1)
```

src/tye/InitHost.cs
```diff
@@ -105,5 +110,30 @@ private static void ThrowIfTyeFilePresent(FileInfo? path, string yml)
...
+            var fileGuid = Guid.NewGuid().ToString("B").ToUpper();
+
+            var slnFiles = Directory.GetFiles(path!.DirectoryName, "*.sln");
+            // noop if there are no sln files
+            if (slnFiles.Length == 1)
+            {
+                // assume the first one is the one wanted.
+                var sln = slnFiles[0];
+                //format
+                var insertItem = @$"Project(""{fileGuid}\"") = ""Solution Items"", ""Solution Items"", ""{fileGuid}""
```
JunTaoLuo, line 125: A few things here:
1. Why is there a backslash `\` after fileGuid? Looks like that's not needed since you already have `""` which should resolve correctly to a quotation mark.
2. How did you come up with this format for insert Item? Specifically, from my understanding of [this doc](https://docs.microsoft.com/en-us/visualstudio/extensibility/internals/solution-dot-sln-file?view=vs-2019#file-body) the two Guids should not be the same, one is a project Guid, one is a project type Guid. The current logic creates a random Guid and uses it for both, which doesn't seem right. Have you tested this logic? What does the sln file look like after you open it?
3. This logic assumes it's safe to insert a folder called `Solution Items`, what if that folder already exists in the sln file?

JunTaoLuo: I tested it out and it seems like adding this section worked:
```
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{D85F58E9-CD8F-45D7-B2F2-ADAE5812E9C0}"
+       ProjectSection(SolutionItems) = preProject
+               tye.yaml = tye.yaml
+       EndProjectSection
+EndProject
```
Still needs logic to resolve whether `Solution Items` already exists in the sln file.

spboyer: Thanks @JunTaoLuo for the comments here. The `\` was a miss on the last edit, will fix this and I'll make a quick change to check for the existing "Solution Items".

src/tye/InitHost.cs
```diff
@@ -105,5 +110,49 @@ private static void ThrowIfTyeFilePresent(FileInfo? path, string yml)
...
+            {
+                var sln = slnFiles[0];
+
+                var lines = File.ReadAllLines(sln).ToList<string>();
+                var insertAt = 0;
+                var exists = false;
+
+                lines.ForEach(l =>
+                {
+                    if (l.IndexOf("(SolutionItems)") > 0)
```
JunTaoLuo, line 127: This logic doesn't work, `(SolutionItems)` may be specified in other itemgroups. For example in gRPC projects, often protobuf files are included as solution items under `Proto` directory e.g. [here](https://github.com/grpc/grpc-dotnet/blob/master/Grpc.DotNet.sln#L20-L35), this logic will add tye.yaml to that directory which wouldn't make any sense. You'll need to look for a project with the name `Solution Items`.

src/tye/InitHost.cs
```diff
@@ -105,5 +110,49 @@ private static void ThrowIfTyeFilePresent(FileInfo? path, string yml)
...
+                if (exists)
+                {
+                    insertItem = "        tye.yaml = tye.yaml";
+                    lines.Insert(insertAt + 1, insertItem);
+                }
+                else
+                {
+                    var fileGuid = Guid.NewGuid().ToString("B").ToUpper();
+
+                    insertItem = @$"Project(""{fileGuid}"") = ""Solution Items"", ""Solution Items"", ""{fileGuid}""
```
JunTaoLuo, line 144: No this is still wrong, the two Guids have different purposes and should not be the same. Again, how are you verifying this change? I think if you opened the sln file, there will be a load error.

spboyer: The `SolutionItems` is a good call out here. Testing process has been using the simple microservices.sln example from the docs. Opening the sln in VS works fine. There is no loading error.

Ideally, there should be `dotnet sln add` functionality in the dotnet CLI to support this beyond **proj files.

JunTaoLuo: When I test this locally, I see
image

It's because the project type Guid is invalid if you just use a randomly generated Guid.

JunTaoLuo: And I agree, maybe file an issue here https://github.com/dotnet/sdk/issues for `dotnet sln` to have the ability to add solution file? Also this is why I mentioned that it would be better to use the Microsoft.DotNet.Cli.Sln.Internal library if we plan to update our TFM.

spboyer: Resolved the Guid issue. There is a section Guid, project references should have the same in the first and then an identifier for the second. Solution items are their own section. Taking a dependency on on dotnet cli commands would be preferred over Microsoft.DotNet.Cli.Sln.Internal in the future. 

Added issue :https://github.com/dotnet/sdk/issues/12454

spboyer:
Closing for now. This is getting ugly and not providing value.

Issue #565, opened 2020-06-26
Tye run wont mount a volume or logPath when path have spaces


aljaff94:
### Describe the bug
When mount a volume or logPath with  a path have a space like "C:\Users\aljaff94\Documents\Projects\My Project", the mounting will not work and throw error 

### Further technical details

- Tye Version 0.3
- Platform: Windows 10 Version 2004
- Docker With WSL2

### Screenshot

![image](https://user-images.githubusercontent.com/5650955/85877706-3e0fc680-b7e0-11ea-94c7-e8fb0dc4e0d4.png)

PR #569, opened 2020-07-02
fix for bug #565

suddenelfilio:
surrounded the volume statement with quotation marks so that spaces don't fail the run


jkotalik:
We should test this on Windows before merging; I'll clone this locally and try it.

jkotalik:
Verified.

Issue #563, opened 2020-06-26
Support podman based distributions


tmds:
Distributions like Fedora, Red Hat Enterprise Linux provide `podman` instead of `docker`.

`podman` has high compatibility with `docker` but the implementation is different. `podman` doesn't have a daemon that runs as `root`. This allows regular users to launch containers without elevating their privileges.

It would be nice if `tye` supports `podman` based distributions and is able to work in a rootless mode. (The network creation currently performed by `tye` requires root.)

I can help implement this.

cc @davidfowl

davidfowl:
I think this makes sense to do. We need to figure out what to do about some of the options in the tye.yaml. They have the name docker 😄. I'm also interested to see how compatible the APIs are (I know what podman is but I haven't looked deeply enough to understand the differences). Maybe throw up a draft PR with the implementation so we can learn more and then we can iterate from there.

PR #570, opened 2020-07-02
Add podman support

tmds:
Creating custom network fails with rootless podman because it is a priviledged operation.


tmds:
Contributes to https://github.com/dotnet/tye/issues/563.

This fixes the main issue I see when using `tye run` on Fedora with `podman`.

I've minimized the `podman` specific bits, so also for `docker` this changes the way the network is configured.

cc @davidfowl

tmds:
Using the "host network" doesn't permit port-mapping.

To allow port-mapping, a "pod network" can be used instead. This is to launch all containers in the same pod, and use "localhost" for inter-container communication.

Both the "host network" and "pod network" approach don't allow multiple containers to bind to the same containerPort. This is a limitation of the rootless network implementation used by podman.

@davidfowl I guess the role of "Microsoft.Tye.Proxy" containers is to allow containers to get access to an ASP.NET container which is running directly on the host (that is: not in a container)? I don't understand how that works. To what address does the ASP.NET application bind?

I will look at using a "pod network" instead of the "host network" next week.
If different modes make sense, they could be made available as sentinel values set for the network name in tye.yaml.

davidfowl:
The idea is being able to talk to software running directly on the host without running that thing in a container and being able to use the host name (I don’t want to muck with host files).

tmds:
How does the proxy container connect to the software running on the host?

I guess it uses the `host.docker.internal` hostname?

According to the Docker docs this:

> connect to the special DNS name host.docker.internal, which resolves to the internal IP address used by the host. This is for development purpose and will not work in a production environment outside of Docker Desktop for Windows.

I assume this is an internal IP address on some virtual bridge.

An ASP.NET Core app by default binds to a loopback address (like `127.0.0.1`). I'd be surprised if the loopback address is accessible from that bridge.

Does it work like this?

If the app would bind not bind to a specific interface (like `0.0.0.0`), then it would be accessible on the bridge. But it would also be accessible to other machines on the (physical) network.

davidfowl:
It works fine on both windows and OS X. Linux however is where host.docker.internal doesn’t work and we use the IP address. We configure how the application binds so we bind to the appropriate address so that it all works

tmds:
@davidfowl this is up for review. All tests pass except those with `SkipIfPodmanAttribute`. The reasons for skipping are in a comment.

```
  Running tests: /home/tmds/repos/tye/artifacts/bin/Microsoft.Tye.E2ETests/Debug/netcoreapp3.1/Microsoft.Tye.E2ETest.dll [netcoreapp3.1|x64]
  Running tests: /home/tmds/repos/tye/artifacts/bin/Microsoft.Tye.UnitTests/Debug/netcoreapp3.1/Microsoft.Tye.UnitTests.dll [netcoreapp3.1|x64]
  Running tests: /home/tmds/repos/tye/artifacts/bin/Microsoft.Tye.Extensions.Configuration.Tests/Debug/netcoreapp3.1/Microsoft.Tye.Extensions.Configuration.Tests.dll [netcoreapp3.1|x64]
  Tests succeeded: /home/tmds/repos/tye/artifacts/bin/Microsoft.Tye.Extensions.Configuration.Tests/Debug/netcoreapp3.1/Microsoft.Tye.Extensions.Configuration.Tests.dll [netcoreapp3.1|x64]
  Tests succeeded: /home/tmds/repos/tye/artifacts/bin/Microsoft.Tye.UnitTests/Debug/netcoreapp3.1/Microsoft.Tye.UnitTests.dll [netcoreapp3.1|x64]
  Tests succeeded: /home/tmds/repos/tye/artifacts/bin/Microsoft.Tye.E2ETests/Debug/netcoreapp3.1/Microsoft.Tye.E2ETest.dll [netcoreapp3.1|x64]
```

Is it possible to add a Fedora VM to CI, to run these tests on a `podman` based system?

tmds:
Using the "host network" seems the best we can do for now.
Podman doesn't support creating user-space networks.
And localhost access is disable also, so "host.docker.internal" won't work anytime soon.

@davidfowl @rynowak @jkotalik can you take a look at the PR?

Is it feasible to add a Fedora VM to the CI so tests run on a podman based system?

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -291,7 +311,23 @@ async Task RunDockerContainer(IEnumerable<(int ExternalPort, int Port, int? Cont
...
+                    foreach (var serviceName in localhostServices!)
+                    {
+                        hostNetworkArgs += $" --add-host {serviceName}:127.0.0.1";
+                    }
+                }
+
+                // Workaround podman issue: https://github.com/containers/libpod/issues/6508
+                // Fixed in podman v2.
+                bool isPodman = await DockerDetector.Instance.IsPodman.Value;
+                string restartArg = isPodman ? "always" : "unless-stopped";
```
davidfowl, line 329: What does this mean for shutdown?

tmds: This affect what happens when the system restarts.
For "docker", "unless-stopped" means only containers which weren't stopped will be started at boot.
"podman" doesn't start containers at boot. With podman v2, "always" and "unless-stopped" are aliases.

davidfowl:
So podman uses the docker CLI commands but is daemonless? This change basically disables docker networking when podman is installed?

tmds:
> This change basically disables docker networking when podman is installed?

Yes, it's because rootless podman doesn't support creating networks.
It's the best we can do for now. When podman gets some more network features (there are open request), we can improve the implementation.

Because we're using the host network, there is no port forwarding. All ports are shared between the containers and host. So the main limitation when using podman is that we can't use the same container port in different containers.
Most images have a fixed container port, so running the same image multiple times doesn't work. For ASP.NET images there is no issue, because the container port can be controlled using ASPNETCORE_URLS.

davidfowl:
The podman support is a little too implicit for my liking. I think it needs to be opt in...

davidfowl:
@tmds Is it possible to run docker and podman side by side or does podman replace docker generally?

tmds:
> I think it needs to be opt in...

I think we need to detect and handle it under the hood, because otherwise `tye` just fails.
As `podman` gets better support for rootless networking, we can improve the implementation.

> Is it possible to run docker and podman side by side or does podman replace docker generally?

`podman` replaces `docker`.

> Is it feasible to add a Fedora VM to the CI so tests run on a podman based system?

Can we look into this? It is the best way to detect regressions on `podman` systems. I'll help fix regressions that turn up.

tmds:
@davidfowl @rynowak @jkotalik can you take a look?

It seems worthwhile to add this to tye and be aware of limitations and regressions on podman-based systems like RHEL and Fedora.

jkotalik:
Apologies, I'll take a look at this shortly.

src/Microsoft.Tye.Hosting/Model/Application.cs
```diff
@@ -29,6 +29,9 @@ public Application(FileInfo source, Dictionary<string, Service> services)
 
         public string? Network { get; set; }
 
+        // All services and application run on the container host.
+        public bool UseHostNetwork { get; set; }
```
jkotalik, line 33: I think you need to parse this in config, right? Do you intend for this to be part of tye.yaml and/or a command line arg to tye run?

jkotalik: Ah nvrm, you use podman existing as that check.

Little bit confusing, should this variable be called IsPodman for now?

src/Microsoft.Tye.Hosting/PortAssigner.cs
```diff
@@ -72,22 +92,23 @@ static int GetNextPort()
                         binding.Name ?? binding.Protocol);
                 }
 
+                // Set ContainerPort for the first http and https port.
+                // For ASP.NET we'll match the Port when UseHostNetwork. ASPNETCORE_URLS will configure the application.
+                // For other applications, we use the default ports 80 and 443.
```
jkotalik, line 97: Why should we default to 80 and 443 for non-aspnet services?

src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs
```diff
@@ -79,18 +79,18 @@ private async Task TransformProjectToContainer(Service service, ProjectRunInfo p
                 IsAspNet = project.IsAspNet
             };
 
-            dockerRunInfo.VolumeMappings.Add(new DockerVolume(source: project.PublishOutputPath, name: null, target: "/app"));
+            dockerRunInfo.VolumeMappings.Add(new DockerVolume(source: project.PublishOutputPath, name: null, target: "/app:z"));
```
jkotalik, line 82: Why was this changed?

jkotalik (approved):

tmds:
@jkotalik podman just got support for rootless networking in the 2.1.0-rc1 that was released 2 days ago. I'm going to redo the PR based on that version.

tmds:
I'm still looking into this. Some more features are required on podman side: being able to access localhost, and adding network aliases to containers.

jzabroski:
@jkotalik  Can you unmark this as approved given @tmds comments? https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/dismissing-a-pull-request-review

jkotalik:
@tdykstra going to close for now as this has gotten stale. Feel free to reopen when you get the chance to work on this.

tdykstra:
@jkotalik I think you meant to tag @tmds

thangchung:
@jkotalik  Is there any plan for supporting podman for `tye`? These days, many people run containers on podman due to licenses for `Docker for Desktop` in some of the small companies. That would be great if we can support it for others, instead of only Docker.

davidfowl:
@tmds AFAIK podman support was merged.

davidfowl:
See https://github.com/dotnet/tye/pull/1014

tmds:
> AFAIK podman support was merged.

Yes. @thangchung, it is part of v0.7.0 and higher.

Issue #540, opened 2020-06-14
fix initial build process in docs/developer_guide.md


KaiWalter:
### Describe the bug
I wanted to build and install a locally build version of Tye following the instruction in docs/developer_guide.md but came across 2 issues.

### To Reproduce - missing package

1. clone repo (commit: `c016f38ce03385a02cba5eab63f71b0af2d01bc9`)
2. execute `./build.cmd`

expected: find required packages in `./artifacts/packages/Debug/Shipping`
actual: folder not present

3. fixed `./build.cmd` and added `-pack` to

```bat
@echo off
powershell -ExecutionPolicy ByPass -NoProfile -command "& """%~dp0eng\common\Build.ps1""" -restore -build -pack %*"
```
4. repeated step 2

expected: find required packages in `./artifacts/packages/Debug/Shipping`
actual: find required packages in `./artifacts/packages/Debug/Shipping`

5. execute (adjust for package version) `dotnet install tye -g --version "0.3.0-dev" --add-source ./artifacts/packages/Debug/Shipping`

error: _Could not execute because the specified command or file was not found._

6. execute `dotnet tool install tye -g --version "0.3.0-dev" --add-source ./artifacts/packages/Debug/Shipping`

error: _error NU1101: Unable to find package tye. No packages exist with this id in source(s): bedrockframework, C:\Users\kai\src\Microsoft\tye\artifacts\packages\Debug\Shipping, dotnet-eng, dotnet-tools, featherhttp, nuget_

7. execute `dotnet tool install microsoft.tye -g --version "0.3.0-dev" --add-source ./artifacts/packages/Debug/Shipping`

SUCCESS!

#### dotnet --info

```
.NET Core SDK (reflecting any global.json):
 Version:   3.1.301
 Commit:    7feb845744

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19041
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.1.301\

Host (useful for support):
  Version: 3.1.5
  Commit:  65cd789777

.NET Core SDKs installed:
  3.1.300 [C:\Program Files\dotnet\sdk]
  3.1.301 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.App 3.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[...]

PR #579, opened 2020-07-14
Update developer guide

JunTaoLuo:
Addresses https://github.com/dotnet/tye/issues/540

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,4 +13,4 @@ while [[ -h $source ]]; do
 done
 
 scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
-"$scriptroot/eng/common/build.sh" --pack --build --restore $@
+"$scriptroot/eng/common/build.sh" --build --restore $@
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -13,4 +13,4 @@ while [[ -h $source ]]; do
 done
 
 scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
-"$scriptroot/eng/common/build.sh" --pack --build --restore $@
```
JunTaoLuo, line 16: It's odd that --pack is specified in the default build.sh, we don't do this elsewhere. Is there a reason for this in this repo?

jkotalik: not sure, but we'd probably need to update the azure-pipelines yaml to add -pack if we remove it.

JunTaoLuo: azure-pipelines uses eng/common/cibuild.sh which passes --pack so it's not necessary to update anything. I took a look and it seems nothing else in the repo references or uses build.sh directly.

jkotalik: Alright, update it then.

JunTaoLuo: Update what?

jkotalik: Oh, you already removed pack... 😄

Issue #527, opened 2020-06-04
Tye Host v0.2 hangs when only launching executables


jgorman-exp:
### Describe the bug
The Tye host just hangs after "Dashboard running on..." when the config file only specifies executables to launch.  This occurs in 0.2.0-alpha.20258.3 and does not occur in 0.1.0-alpha.20209.5.  It is preventing us from using 0.2 for some of our deployment workflows.

### To Reproduce
`tye run tye.yaml`

tye.yaml:
```
name: test
services:
- name: test-app
  executable: C:\Windows\System32\systeminfo.exe
```

### Further technical details
```
> tye --version
0.2.0-alpha.20258.3+bf3bc0913b37e77f5f3ef14f409b3feb794cc447

> tye run -v debug tye.yaml
Loading Application Details...
Launching Tye Host...

[10:59:55 INF] Executing application from C:\Source\Xpresso\dev-build-scripts\tye.yaml
[10:59:55 INF] Dashboard running on http://127.0.0.1:8000
```

Platform:  Windows

davidfowl:
Are you sure it isn't running the app and quitting over and over again?

jgorman-exp:
@davidfowl Doesn't seem to be.  Version 0.2 never makes it to the "Launching service..." log, and there are no logs for the service in the dashboard.  Version 0.1 does run the executable, prints logs in the dashboard and shows an increasing number of restarts on the dashboard's home page.  Screenshot is of the dashboard for v0.2, bottom of the terminal window is v0.2 output, top of the terminal window is part of v0.1 output, for comparison.

![image](https://user-images.githubusercontent.com/6437309/83956080-fa442580-a80e-11ea-8efa-bb0b1bb93d25.png)

For the actual executable I'm running (not the executable in this minimal repro), the action the executable takes is never executed.  Thanks for taking a look.

davidfowl:
Take a look at the dashboard and look at the logs for that specific service.

jgorman-exp:
@davidfowl I included the Tye dashboard and the logs for that service in the screenshot above--there are no logs for the service, just a black box.

JunTaoLuo:
This is a bug caused by https://github.com/dotnet/tye/commit/11ada3f1f715551adbb37b98c4f95bb5398ce3cf. The executable isn't being launched anymore. I'll try out a fix.

PR #583, opened 2020-07-15
Project with executable services only should launch

JunTaoLuo:
Fixes https://github.com/dotnet/tye/issues/527. WIP since I'm looking to add a test.


src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -144,19 +144,19 @@ private async Task BuildAndRunProjects(Application application)
                     _logger.LogInformation("Building projects failed with exit code {ExitCode}: \r\n" + buildResult.StandardOutput, buildResult.ExitCode);
                     return;
                 }
+            }
 
-                foreach (var s in application.Services)
+            foreach (var s in application.Services)
```
jkotalik, line 149: Ahah good one.

jkotalik:
IMO I don't think you need a test for this.

JunTaoLuo:
Making a fix without a regression test? Absolutely barbaric, but I won't complain.

PR #584, opened 2020-07-15
Fix some metric on dashboard

jkotalik:
And other small nits

### DIFF
--- a/samples/frontend-backend/frontend/Program.cs
+++ b/samples/frontend-backend/frontend/Program.cs
@@ -14,9 +14,6 @@ namespace Frontend
         {
             Activity.DefaultIdFormat = ActivityIdFormat.W3C;
             CreateHostBuilder(args).Build().Run();
-
-            
-
         }
 
         public static IHostBuilder CreateHostBuilder(string[] args) =>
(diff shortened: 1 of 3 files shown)


samples/frontend-backend/frontend/Program.cs
```diff
@@ -14,9 +14,6 @@ public static void Main(string[] args)
         {
             Activity.DefaultIdFormat = ActivityIdFormat.W3C;
             CreateHostBuilder(args).Build().Run();
-
```
davidfowl, line 17: This is the best part of this change wtf.

src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -397,6 +397,15 @@ async Task RunApplicationAsync(IEnumerable<(int ExternalPort, int Port, string?
                     {
                         _logger.LogError(0, ex, "Failed to launch process for service {ServiceName}", replica);
 
+                        if (!_options.Watch)
```
davidfowl, line 400: That doesn't seem like a nit

samples/frontend-backend/frontend/Program.cs
rynowak: @JamesNK

src/Microsoft.Tye.Hosting/ProcessRunner.cs
jkotalik: Yeah, should have also mentioned this.

samples/frontend-backend/frontend/Program.cs
jkotalik: uhhh dotnet watch testing stuff.......

JamesNK: Stop right there! You are under arrest by the whitespace police

image

Issue #561, opened 2020-06-26
deploy failed. Unsupported TFM net5.0


JamesNK:
```
C:\Development\Temp\microservice> dotnet --info
.NET SDK (reflecting any global.json):
 Version:   5.0.100-preview.7.20312.14
 Commit:    ee6dbf09d4
C:\Development\Temp\microservice> tye --version
0.3.0-alpha.20319.3+4c7d799624984bba1f0622a0aef92b37a9ece5a8
C:\Development\Temp\microservice> tye deploy --interactive
Loading Application Details...
Enter the Container Registry (ex: 'example.azurecr.io' for Azure or 'example' for dockerhub): jamesnk.azurecr.io
Processing Service 'frontend'...
    Applying container defaults...
Drats! 'deploy' failed:
        Unsupported TFM net5.0.
```

I'm guessing a global.json targeting 3.1 will resolve.

JamesNK:
global.json didn't fix it. The fix was changing `<TargetFramework>net5.0</TargetFramework>` to `<TargetFramework>netcoreapp3.1</TargetFramework>`

Is there a bug here where tye wasn't picking up `5.0.100-preview.7.20312.14`? It is a nightly build of the SDK.

Complete dotnet --info:
```
C:\Development\Temp\microservice> dotnet --info
.NET SDK (reflecting any global.json):
 Version:   5.0.100-preview.7.20312.14
 Commit:    ee6dbf09d4

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18363
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\5.0.100-preview.7.20312.14\

Host (useful for support):
  Version: 5.0.0-preview.7.20311.11
  Commit:  b179e19840

.NET SDKs installed:
  1.1.11 [C:\Program Files\dotnet\sdk]
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.500 [C:\Program Files\dotnet\sdk]
  2.1.502 [C:\Program Files\dotnet\sdk]
  2.1.504 [C:\Program Files\dotnet\sdk]
  2.1.505 [C:\Program Files\dotnet\sdk]
  2.1.507 [C:\Program Files\dotnet\sdk]
  2.1.508 [C:\Program Files\dotnet\sdk]
  2.1.600 [C:\Program Files\dotnet\sdk]
  2.1.602 [C:\Program Files\dotnet\sdk]
[...]

PR #585, opened 2020-07-16
Add support for 5.0 and change sample to 5.0

jkotalik:
Fixes https://github.com/dotnet/tye/issues/561

### DIFF
--- a/src/Microsoft.Tye.Core/DockerfileGenerator.cs
+++ b/src/Microsoft.Tye.Core/DockerfileGenerator.cs
@@ -101,7 +101,7 @@ namespace Microsoft.Tye
                 container.BaseImageName = "mcr.microsoft.com/dotnet/core/runtime";
             }
 
-            if (string.IsNullOrEmpty(container.BaseImageTag) && project.TargetFrameworkName == "netcoreapp")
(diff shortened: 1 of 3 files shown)


jkotalik:
There are two flaky tests we gotta fix 😟

src/Microsoft.Tye.Core/DockerfileGenerator.cs
```diff
@@ -112,7 +112,7 @@ public static void ApplyContainerDefaults(ApplicationBuilder application, Dotnet
             }
 
             container.BuildImageName ??= "mcr.microsoft.com/dotnet/core/sdk";
-            container.BuildImageTag ??= "3.1";
+            container.BuildImageTag ??= project.TargetFrameworkName == "net" ? "5.0" : "3.1";
```
JamesNK, line 115: What happens when net6.0 is out?

jkotalik: Going to play around with setting this to the TargetFrameworkVersion. I think the idea before was that if someone has a 2.1 app, the 3.1 sdk would still be able to build it. However, now that we are dealing with 5.0, I'd rather not for all builds to happen with a preview SDK, hence the condition.

Issue #436, opened 2020-05-03
Tye hangs out for a while doing `Loading Application Details...` if kubernetes is unreachable


rynowak:
We execute `kubectl cluster-info` to test if the user has a valid kubernetes cluster for us to talk to. However if the cluster is unreachable, we'll hang out for a few minutes displaying just:

```txt
Loading Application Details...
```

It's the underlying kubectl command that's taking a while, and there's not much we can do if your cluster is unreachable - but we could provide a better UI experience.

PR #586, opened 2020-07-16
Add logging for KubectlDetector logic

JunTaoLuo:
Fixes https://github.com/dotnet/tye/issues/436

### DIFF
--- a/src/shared/KubectlDetector.cs
+++ b/src/shared/KubectlDetector.cs
@@ -7,50 +7,57 @@ using System.Threading.Tasks;
 
 namespace Microsoft.Tye
 {
-    internal class KubectlDetector
+    internal static class KubectlDetector
     {
-        private static bool? _kubectlInstalled = null;
-        private static bool? _kubectlConnectedToCluster = null;
+        private static Lazy<Task<bool>> _kubectlInstalled = new Lazy<Task<bool>>(DetectKubectlInstalled);
+        private static Lazy<Task<bool>> _kubectlConnectedToCluster = new Lazy<Task<bool>>(DetectKubectlConnectedToCluster);
 
-        public static async Task<bool> IsKubectlInstalledAsync(OutputContext output)
+        public static Task<bool> IsKubectlInstalledAsync(OutputContext output)
         {
-            if (!_kubectlInstalled.HasValue)
+            if (!_kubectlInstalled.IsValueCreated)
             {
                 output.WriteInfoLine("Verifying kubectl installation...");
-                try
-                {
-                    // Ignoring the exit code and relying on Process to throw if kubectl is not found
-                    // kubectl version will return non-zero if you're not connected to a cluster.
-                    await ProcessUtil.RunAsync("kubectl", "version", throwOnError: false);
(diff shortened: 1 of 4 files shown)


src/shared/KubectlDetector.cs
```diff
@@ -9,46 +9,48 @@ namespace Microsoft.Tye
 {
     internal class KubectlDetector
     {
-        public static KubectlDetector Instance { get; } = new KubectlDetector();
+        private static bool? _kubectlInstalled = null;
```
JunTaoLuo, line 12: Curious why the whole Lazy construction was here? I looked at all the use cases and it doesn't seem like we have any thread safety concerns. I've simplified it to just be a bool, let me know if there is a reason to use Lazy instead.

jkotalik: I mean in general I think the best practice here is to use lazy, as that is thread safe.

What was the motivation for removing Lazy?

jkotalik: To pass in the output context?

JunTaoLuo: Nah, I just removed it to simplify. I can add the lazy back but it felt unnecessary, this never gets called in parallel.

JunTaoLuo:
FYI this is the new output when calling deploy
```
(tye) C:\gh\tye\src\tye [master ≡ +0 ~3 -0 !]> dotnet run -- deploy C:\gh\tye\samples\frontend-backend -i
Loading Application Details...
Verifying kubectl installation...
Verifying kubectl connection to cluster...
Enter the Container Registry (ex: 'example.azurecr.io' for Azure or 'example' for dockerhub): juntaoluo
Processing Service 'backend'...
    Applying container defaults...
    Compiling Services...
    Publishing Project...
    Building Docker Image...
        Created Docker Image: 'juntaoluo/backend:0.4.0-dev'
    Pushing Docker Image...
        Pushed docker image: 'juntaoluo/backend:0.4.0-dev'
    Validating Secrets...
    Generating Manifests...
Processing Service 'frontend'...
    Applying container defaults...
    Compiling Services...
    Publishing Project...
    Building Docker Image...
        Created Docker Image: 'juntaoluo/frontend:0.4.0-dev'
    Pushing Docker Image...
        Pushed docker image: 'juntaoluo/frontend:0.4.0-dev'
    Validating Secrets...
    Generating Manifests...
Deploying Application Manifests...

        Verifying kubectl installation...
        Verifying kubectl connection to cluster...
        Writing output to 'C:\Users\johluo\AppData\Local\Temp\tmp6864.tmp'.
        Deployed application 'frontend-backend'.
Time Elapsed: 00:00:28:08
```

src/shared/KubectlDetector.cs
rynowak: ewww a "best practice"

My original reasoning for using lazy was to avoid thread starvation or running multiple kubectl processes if we ever *do* run it in parallel. It was easy enough to avoid ever being a problem, and usually stuff that does build/publish grows parallel execution before long.

PR #587, opened 2020-07-16
Start finding func rather than downloading it

jkotalik:
TODO:
- Logs
- Fixing test (need to make sure func is installed).
- Doc updates


test/E2ETest/TyeRunTests.cs
```diff
@@ -75,12 +76,27 @@ public async Task FrontendBackendRunTest()
             });
         }
 
-        [Fact]
+        [Fact(Skip = "Need to figure out how to install func before running")]
```
JunTaoLuo, line 79: A new build step in the azure-pipelines yaml?

jkotalik: Yeah eventually, or something line that.

Issue #543, opened 2020-06-16
tye run continues when build errors happen


timheuer:
### Describe the bug
When you have a code bug that prevents build succeeding and you run `tye run` the build error will occur and fail build, `run` will fail of course, but the CLI doesn't exit with any exit code.

### To Reproduce
Follow the getting started tutorials but put a code error (omit a semicolon line ending) that would cause a build fail.
Execute `tye run`
EXPECT: build fails, and tye CLI exits with some exit code != 0
ACTUAL build fails, tye CLI stays running but nothing actually runs

- Include the output of `tye --version`
0.2.0-alpha.20258.3+bf3bc0913b37e77f5f3ef14f409b3feb794cc447
- The platform (Linux/macOS/Windows)
Windows

jkotalik:
I think this behavior is expected, however we could make it clearer what is going on. We tried to make the tye host match kubernetes in behavior with regards to starting process. If a process constantly fails to start in k8s, the other services will still run normally. 

Another example would be that we recently added watch support in tye. Commonly a change is saved and it causes the build to fail for an individual service. Making tye exit in that scenario wouldn't make sense.

When you were running `tye run` and the build failed, when you went to the dashboard, were other services running?

timheuer:
The matched behavior seems odd. I ask myself what do I expect run to do? If the project fails to build I expect that run wouldn’t succeed then and exit. The output didn’t get to the point of telling me the dashboard URI. Build fail message then basically a Console.Readline kind of feel. To me all I saw was a fail message on build and nothing to tell me that a dashboard still would exist in this state. 

To me if the experience of the is to discover, build, orchestrate the full solution then if anything fails in that startup I think it should fail the run.

jkotalik:
I think I'm seeing your viewpoint now; that isn't a great experience for build errors.

I think on first build, if any projects fail to build, we can exit tye. This would only be for build, not running the process.

timheuer:
Yes correct.  Thanks for understanding.

PR #588, opened 2020-07-16
Make tye exit if dotnet fails to start projects

jkotalik:
Fixes https://github.com/dotnet/tye/issues/543.

image


src/Microsoft.Tye.Hosting/TyeBuildException.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public TyeBuildException(string? message) : base(message)
+        {
+        }
+
+        public TyeBuildException(string? message, Exception? innerException) : base(message, innerException)
+        {
+        }
+
+        protected TyeBuildException(SerializationInfo info, StreamingContext context) : base(info, context)
```
JunTaoLuo, line 25: Nit: Not all of these constructors are needed right?

src/Microsoft.Tye.Hosting/TyeBuildException.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Tye.Hosting
+{
+    [Serializable]
```
JunTaoLuo, line 10: nit: is this required?

src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -257,7 +265,7 @@ private static bool IsPortInUseByBinding(Application application, int port)
             return false;
         }
 
-        private static AggregateApplicationProcessor CreateApplicationProcessor(ReplicaRegistry replicaRegistry, HostOptions options, Microsoft.Extensions.Logging.ILogger logger)
+        private static AggregateApplicationProcessor CreateApplicationProcessor(ReplicaRegistry replicaRegistry, HostOptions options, Microsoft.Extensions.Logging.ILogger logger, IHostApplicationLifetime lifetime)
```
JunTaoLuo, line 268: nit: this change isn't needed anymore?

src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -11,8 +11,10 @@
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Build.Globbing.Extensions;
```
JunTaoLuo, line 14: I'm guessing some of these imports are no longer used?

src/Microsoft.Tye.Hosting/TyeBuildException.cs
jkotalik: These are autogenerated when creating an exception type, I'd rather not modify it. IIRC it's a requirement for exceptions in the runtime repo to have these constructors as well.

PR #591, opened 2020-07-16
Fix static file loading when using ingress with host

areller:
I've encountered a problem when using an ingress host rule to route to UI applications.

Example:

```
- host: ui.example.com
  service: appC-ui
```

The problem was that static files (`.css`, `.js`, etc..) wouldn't load

Turns out that `MapFallback` doesn't route to static files.


jkotalik:
Looks like some tests are still failing.

areller:
@jkotalik looks like it's complaining about the new project in testassets, probably because it targets net5.0
I'll downgrade

jkotalik:
Yeah, we were planning on adding 5.0 support this release, but we hit some major roadblocks that we need to investigate with regards to msbuild loading.

PR #600, opened 2020-07-19
Graceful termination in Windows

areller:
Before this change, Tye would `CloseMainWindow` but that didn't have any effect because `CreateNoWindow` was set to `true` in the `ProcessStartInfo`. 

Now, when in Windows, I run the process with an invisible window and when I call `CloseMainWindow`, I wait 5 seconds for the process to exit gracefully (just like in Unix), before it's forcefully killed.

If `CloseMainWindow` fails (host can run from an environment where spawning windows is not possible), I default to killing the process (just like it has behaved before for Windows).

This solves a problem that I was having, I had an executable (parent) that launched another executable (child) and when Tye killed the parent, it would the child hanging.

### DIFF
--- a/src/Microsoft.Tye.Core/ProcessUtil.cs
+++ b/src/Microsoft.Tye.Core/ProcessUtil.cs
@@ -14,8 +14,14 @@ namespace Microsoft.Tye
 {
     public static class ProcessUtil
     {
+        #region Native Methods
+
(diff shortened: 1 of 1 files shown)


areller:
Do you have any thoughts about this one? @davidfowl @jkotalik

JunTaoLuo (approved):
LGTM

PR #608, opened 2020-07-25
Setup default RunCommand in case MSBuild project RunCommand is undefined

alexfdezsauco:
Sometimes a project RunCommand can't be resolved

image

Such situation causes that process launch failed.

image

### DIFF
--- a/src/Microsoft.Tye.Core/ProjectReader.cs
+++ b/src/Microsoft.Tye.Core/ProjectReader.cs
@@ -240,6 +240,12 @@ namespace Microsoft.Tye
             // Figure out if functions app.
             // If so, run app with function host.
             project.RunCommand = projectInstance.GetPropertyValue("RunCommand");
+            if (string.IsNullOrEmpty(project.RunCommand) && projectInstance.GetPropertyValue("OutputType")?.ToLower() == "exe")
+            {
+                var outputPath = Path.Combine(Directory.GetCurrentDirectory(), projectInstance.GetPropertyValue("OutputPath"), project.TargetFrameworks[0], $"{projectInstance.GetPropertyValue("AssemblyName")}.exe");
+                project.RunCommand = Path.GetFullPath(new Uri(outputPath).LocalPath);
+            }
+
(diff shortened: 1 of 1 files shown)


davidfowl:
When?

alexfdezsauco:
> When?

Good question. I have a project configuration where the project instance RunCommand property is empty, but not sure why.
I started to get the `Cannot start process because the file name has not been provided` error when I try to hosted it on tye. So, make take that Tye could survive to such situation by trying with a default RunCommand.

tebeco:
I know this will sound stupid but does the project build ?
The behavior of `tye` when the udnerlying service is ` project` and it fails on internal `dotnet restore` or `dotnet build` can be "strange" and ends up in an error that look like yours
because the binary could not be built, then there was no file provided to run the process

alexfdezsauco:
@tebeco Yes, it builds. I will try to extract the repo (but it will take longer than the time it took me to find this problem). With this default RunCommand, I'm able to continue to work and host the solution on Tye.

Well, now I noticed, that it requires a second `tye run`.

alexfdezsauco:
I will close this because it is not a definitive solution.  It requires to `tye run` in both the RunCommand  is emby but in the second one it runs.

alexfdezsauco:
@tebeco @davidfowl  Well I set <TargetFramework>netcoreapp3.1</TargetFramework> instead <TargetFrameworks>netcoreapp3.1</TargetFrameworks> and  RunCommand got the expected value. 

So, if TargetFramework is undefined even when TargetFrameworks have values the RunCommand will be empty at least on my configuration.

tebeco:
Welcome the the club, good news is that you just need to wait for this to be merged :
https://github.com/dotnet/tye/pull/567

also you can ALREADY edit you yaml by adding `buildProperties`:
```yaml
name: app-with-targetframeworks
services:
- name: multipletargetframeworks
  project: MultipleTargetFrameworks/MultipleTargetFrameworks.csproj
  buildProperties:
  - name: TargetFramework
    value: netcoreapp3.1
```

PR #642, opened 2020-08-28
Update tye version description

angelobelchior:


jkotalik:
The version there reflects the current version on nuget: https://www.nuget.org/packages/Microsoft.Tye/0.4.0-alpha.20371.1. We don't update it until we ship a new version on nuget.

PR #650, opened 2020-08-31
Fixed unquoted file paths

ckadluba:
Found and fixed two more cases of commands executed with an unquoted file path leading to problems with spaces in the path (see issue #647). The following commands were fixed:
* Parameter working directory when calling docker run.
* Parameter clone directory when calling git clone.


jkotalik:
Thanks!

PR #659, opened 2020-09-15
Fix quotations around working directory

JunTaoLuo:
This should fix the ubuntu builds


davidfowl (approved):
Is this the only thing to quote?

jkotalik (approved):

PR #664, opened 2020-09-17
Improve nginx-ingress experience

glennc:
I wanted to investigate the bugs I heard about ingress not working after an undeploy/deploy and look at reporting the ingress IPs after deploying with ingress. 

Running this code showed me that after deploying it can take a while to actually assign an IP. K8s docs say it can take "up to a minute or two". So I am thinking we want to wait for it to take and maybe show some progress when running in interactive mode. We can put it in here like I have but I was thinking it seems like it's time to add a `StatusStep` to the end of `DeployCommand.cs` that has a set of things we check on and report after deployment has happened.

What do you think?

### DIFF
--- a/samples/dockercompose/ingress.yml
+++ b/samples/dockercompose/ingress.yml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: ingress-basic
--- a/src/Microsoft.Tye.Core/DeployApplicationKubernetesManifestStep.cs
+++ b/src/Microsoft.Tye.Core/DeployApplicationKubernetesManifestStep.cs
@@ -62,36 +62,37 @@ namespace Microsoft.Tye
             output.WriteInfoLine($"Deployed application '{application.Name}'.");
             if (application.Ingress.Count > 0)
             {
-                foreach(var ingress in application.Ingress)
+                foreach (var ingress in application.Ingress)
                 {
                     using var ingressStep = output.BeginStep($"Retrieving details for {ingress.Name}...");
 
                     var done = false;
 
-                    Action<string> complete = line => {
+                    Action<string> complete = line =>
+                    {
                         done = line != "''";
-                        if(done) 
+                        if (done)
                         {
                             output.WriteInfoLine($"IngressIP: {line}");
                         }
                     };
 
                     var retries = 0;
-                    while(!done && retries < 60)
+                    while (!done && retries < 60)
                     {
                         var ingressExitCode = await Process.ExecuteAsync(
                             "kubectl", 
-                            $"get ingress {ingress.Name} -o jsonpath='{{..ip}}'", 
+                            $"get ingress {ingress.Name} -o jsonpath='{{..ip}}'",
                             System.Environment.CurrentDirectory,
                             complete,
                             capture.StdErr);
 
-                        if(ingressExitCode != 0)
+                        if (ingressExitCode != 0)
                         {
                             throw new CommandException("'kubectl get ingress' failed");
                         }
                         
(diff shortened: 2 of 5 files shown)


jkotalik:
Fix the formatting!

glennc:
But it's a meme at this point, if I don't submit formatting errors y'all will think something is wrong.

jkotalik:
@glennc I assume you won't have time to finish this PR, want one of us to take it over?

src/Microsoft.Tye.Core/DeployApplicationKubernetesManifestStep.cs
```diff
@@ -59,6 +60,47 @@ public override async Task ExecuteAsync(OutputContext output, ApplicationBuilder
...
+                            capture.StdErr);
+
+                        if (ingressExitCode != 0)
+                        {
+                            throw new CommandException("'kubectl get ingress' failed");
+                        }
+
+                        if (!done)
+                        {
+                            await Task.Delay(2000);
```
JunTaoLuo, line 98: Sure we want to wait 2 seconds? That seems quite long, especially if there are 60 retries.

JunTaoLuo (approved):
Changes look good.

src/Microsoft.Tye.Core/DeployApplicationKubernetesManifestStep.cs
jkotalik: It takes at least 20 seconds from my trials

samples/dockercompose/ingress.yml
```diff
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
```
rynowak, line 1: FYI this is a Kubernetes 1.19 feature. The default on AKS is 1.18.10 for westus2 at the time of checking. I found this notable because there's a few features in 1.19 that Tye *could* take advantage of for ingress, but we didn't do it when I wrote this because the paint wasn't quite dry yet.

rynowak: Specifically in 1.19 `IngressClass` is a resource type on its own: https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class

This means that you can query the cluster for existing `IngressClass` definitions, and use them without being coupled to nginx-ingress.  The whole set of features related to detecting the presence of nginx-ingress would be replaced by a lookup of a documented resource type 😆 - and Tye can start leveraging other ingress controllers.

jkotalik: I thought it was 1.19... I'll try verifying this with 1.18 to see what happens.

jkotalik: And *very cool*

src/Microsoft.Tye.Core/ValidateIngressStep.cs
```diff
@@ -140,29 +140,28 @@ public override async Task ExecuteAsync(OutputContext output, ApplicationBuilder
             else
             {
                 // If we get here then we should deploy the ingress controller.
+
+                // For some reason, the first time we apply the ingress controller, an exception is thrown
```
rynowak, line 144: It's not really *for some reason* - this is plot hole in how CRDs + validating webhooks work. Basically if you create a new CRD + a new webhook to validate it - what should the behavior be before the webhook starts? The author of the webhook can choose to be strict, and say no one gets to create one of these without my approval. There's no policy for "wait 30 sec to see if the webhook starts, and try again".

That's the problem you're likely seeing. Unfortunately helm doesn't solve it either :-/

jkotalik: Yeah that's the error I'm seeing.

Issue #657, opened 2020-09-13
docker-compose like environment variables syntax


areller:
### What should we add or change to make your life better?

There are 3 changes I would like to have

1. Allow to specify environment variables using the [docker-compose syntax](https://docs.docker.com/compose/environment-variables/#set-environment-variables-in-containers)

Example:

```
name: demo-envvars
services:
        - name: console
          project: console/console.csproj
          env:
                  - SOME_VAR_1=SOME_VALUE
                  - SOME_VAR_2="SOME VALUE WITH SPACES"
```

The wouldn't replace the old syntax (having to explicitly write `name` and `value`), but only provide an alternative

2. Allow to pass external environment variables to tye, per the [docker-compose syntax](https://docs.docker.com/compose/environment-variables/#pass-environment-variables-to-containers)

You should be able to specify the names of the environment variables you'd want to pass from the external environment (shell/etc...). Tye would read the values from the environment and pass them into the service

Example:

```
name: demo-envvars
services:
        - name: console
          project: console/console.csproj
          env:
                  - SOME_VAR_1
                  - SOME_VAR_2
```

(This should also work with the existing env vars syntax, if you specify a `name` without a `value`)

3. Allow to specify the path to an env file, per the [docker-compose syntax](https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option)

The env file would have this syntax

```
SOME_VAR_1=SOME_VALUE
SOME_VAR_2="SOME VALUE WITH SPACES"
```

and to reference it in `tye.yaml` you would do

```
name: demo-envvars
services:
        - name: console
          project: console/console.csproj
          env_file:
                  - my_vars_1.env
                  - my_vars_2.env
```

(You would be able to specify multiple files)

### Why is this important to you?

1. The current syntax for environment variables feels overly verbose, compared to docker-compose's syntax, especially, when dealing with 5~10+environment variables.

2. It would make it much more convenient to copy variables from docker-compose to tye or vice-versa

PR #669, opened 2020-09-21
Docker compose environment variable syntax

areller:
This PR addresses this issue https://github.com/dotnet/tye/issues/657

1. The `env` field in a service now accepts string of environment variables where the key and value are separated by a `=` sign.
2. If an element in `env` is either missing a `value` key or is a string that misses a `=` sign, the value of the environment variable is taken from the operating system/shell.
3. There is a new field under service, `env_file` which is a collection of strings which represent file paths. These file paths are formatted as explained in https://docs.docker.com/compose/env-file/

### DIFF
--- a/src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
+++ b/src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
@@ -2,7 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using Microsoft.Tye.ConfigModel;
 using YamlDotNet.RepresentationModel;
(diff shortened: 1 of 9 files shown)


areller:
@jkotalik @davidfowl @JunTaoLuo any thoughts about this?

davidfowl:
I like it! We used to have this but then we had the idea of config sources

src/Microsoft.Tye.Core/Serialization/ConfigServiceParser.cs
```diff
@@ -457,6 +483,66 @@ private static void HandleServiceConfigurationNameMapping(YamlMappingNode yamlMa
...
+            var nodeValue = YamlParser.GetScalarValue(yamlScalarNode);
+            var keyValueSeparator = nodeValue.IndexOf('=');
+
+            if (keyValueSeparator != -1)
+            {
+                var key = nodeValue.Substring(0, keyValueSeparator).Trim();
+                var value = nodeValue.Substring(keyValueSeparator + 1)?.Trim();
+
+                config.Name = key;
+                config.Value = value?.Trim(new[] { ' ', '"' }) ?? string.Empty;
```
davidfowl, line 497: Do we have tests for this case?

Issue #594, opened 2020-07-16
Run msbuild out of proc


JunTaoLuo:
We currently run msbuild in proc which gets into trouble when we resolve 5.0 sdk. The tasks used by the 5.0 SDK needs 5.0 assemblies which are not available when msbuild runs in proc.

ckadluba:
I ran into this problem too with .NET 5 SDK (5.0.0-preview.7.20366.1) installed.

```
PS C:\Source\github\tye> dotnet --list-runtimes
Microsoft.AspNetCore.All 2.1.16 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
Microsoft.AspNetCore.App 2.1.16 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
Microsoft.AspNetCore.App 3.1.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
Microsoft.AspNetCore.App 3.1.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
Microsoft.AspNetCore.App 5.0.0-preview.7.20365.19 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
Microsoft.NETCore.App 2.1.16 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
Microsoft.NETCore.App 3.1.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
Microsoft.NETCore.App 3.1.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
Microsoft.NETCore.App 5.0.0-preview.7.20364.11 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
Microsoft.WindowsDesktop.App 3.1.2 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
Microsoft.WindowsDesktop.App 3.1.6 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
Microsoft.WindowsDesktop.App 5.0.0-preview.7.20366.1 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

`tye run` on the voting sample produced the following error.

```
PS C:\Source\github\tye\samples\voting> tye run
Loading Application Details...
Drats! 'run' failed:
        Failed to load project: 'C:\Source\github\tye\samples\voting\vote\vote.csproj'.

[...]

PR #674, opened 2020-09-21
Use out-of-proc MSBuild evaluation

JunTaoLuo:
Fixes https://github.com/dotnet/tye/issues/594

I have tested to ensure that the original issue where installing a 5.0 SDK would break apps has been fixed with this change. With this change, supporting new TFMs wouldn't require tye itself to be re-targeted.

So far testing the perf with frontend-backend sample gives:

Before:
backend.csproj
Evaluation Took: 1500.9242ms
frontend.csproj
Evaluation Took: 463.2118ms

After:
backend.csproj
Evaluation Took: 1733.3304ms
frontend.csproj
Evaluation Took: 1766.8971ms

So there is a performance penalty with running out of proc, especially with additional projects.

### DIFF
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -135,7 +135,8 @@ namespace Microsoft.Tye
                     if (msbuildEvaluationResult.ExitCode != 0)
                     {
                         output.WriteDebugLine($"Evaluating project failed with exit code {msbuildEvaluationResult.ExitCode}:" +
-                            $"{Environment.NewLine}{msbuildEvaluationResult.StandardError}");
+                            $"{Environment.NewLine}Ouptut: {msbuildEvaluationResult.StandardOutput}" +
+                            $"{Environment.NewLine}Error: {msbuildEvaluationResult.StandardError}");
                     }
 
                     var msbuildEvaluationOutput = msbuildEvaluationResult
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -121,12 +121,12 @@ namespace Microsoft.Tye
                     output.WriteDebugLine("Restoring and evaluating projects");
 
                     var msbuildEvaluationResult = await ProcessUtil.RunAsync(
-                        "dotnet", 
+                        "dotnet",
                         $"build " +
                             $"\"{projectPath}\" " +
                             $"/p:CustomAfterMicrosoftCommonTargets={Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!, "ProjectEvaluation.targets")} " +
-                            $"/nologo", 
-                        throwOnError: false, 
+                            $"/nologo",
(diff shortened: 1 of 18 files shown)


src/Microsoft.Tye.Core/Microsoft.Tye.Core.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
```
jkotalik, line 4: Why downgrade these?

jkotalik (approved):
Change itself looks good, @glennc @davidfowl I think this is a price we'll need to pay for now to support .NET 5.

davidfowl:
I'm not sure I want to pay this price. I'd like to discuss different approaches, like possibly doing more computation of all projects in a single pass rather than trigger an invocation per project.

davidfowl (changes requested):
Lets try to see if we can do this in a bigger batch.

src/Microsoft.Tye.Core/Microsoft.Tye.Core.csproj
JunTaoLuo: The TFMs were upgraded to support "net5.0" projects. Now that's no longer needed so I've reverted them to netcoreapp3.1 which is LTS.

JunTaoLuo:
Updated numbers:

After:
backend.csproj and frontend.csproj: 2207.3488ms

I'll collect some numbers with more projects tomorrow.

src/Microsoft.Tye.Core/Microsoft.Tye.Core.csproj
jkotalik: But 5.0 is cool and new.

src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -74,6 +77,87 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
...
+                    var msbuildEvaluationResult = await ProcessUtil.RunAsync(
+                        "dotnet",
+                        $"build " +
+                            $"\"{projectPath}\" " +
+                            $"/p:CustomAfterMicrosoftCommonTargets={Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!, "ProjectEvaluation.targets")} " +
+                            $"/nologo",
+                        throwOnError: false,
+                        workingDirectory: directory.DirectoryPath);
+
+                    // If the build fails, we're not really blocked from doing our work.
```
jkotalik, line 132: What can cause the build to fail here? Would the failure to evaluate a project be shown when we throw a CommandException?

jkotalik (approved):

src/Microsoft.Tye.Core/ApplicationFactory.cs
JunTaoLuo: I didn't run into any cases where this fails but to be defensive, I've added the details of the output to the debug logs.

Issue #575, opened 2020-07-10
tye rewrite Dapr Environment variable


rspaulino:
I'm using tye locally(tye run) to Run Dapr locally as well and use service to service invocation from Dapr.

`$"http://localhost:{Environment.GetEnvironmentVariable("DAPR_HTTP_PORT")}/v1.0/invoke/backend/method/"`

this work running when using dapr run or deploy to k8s. I wrote a method to print all environment variable to see why when using tye is coming empty and it look like tye is rewriting the environment variable. 

`$"http://localhost:{Environment.GetEnvironmentVariable("SERVICE__FRONTEND-DAPR__HTTP__PORT")}/v1.0/invoke/backend/method/"`

and this is the output of the Environment variable from tye local:

```
[frontend_e61771b1-9]: SERVICE__BACKEND-DAPR__HTTP__HOST = localhost
[frontend_e61771b1-9]: SERVICE__FRONTEND-DAPR__HTTP__HOST = localhost
[frontend_e61771b1-9]: BACKEND-DAPR_METRICS_SERVICE_HOST = localhost
[frontend_e61771b1-9]: BACKEND-DAPR_METRICS_SERVICE_PROTOCOL = http
[frontend_e61771b1-9]: SERVICE__FRONTEND__HTTPS__PROTOCOL = https
[frontend_e61771b1-9]: SHLVL = 1
[frontend_e61771b1-9]: FRONTEND_HTTPS_SERVICE_PROTOCOL = https
[frontend_e61771b1-9]: SERVICE__BACKEND-DAPR__METRICS__PROTOCOL = http
[frontend_e61771b1-9]: __CF_USER_TEXT_ENCODING = 0x1F5:0x0:0x0
[frontend_e61771b1-9]: FRONTEND-DAPR_HTTP_SERVICE_PROTOCOL = http
[frontend_e61771b1-9]: SERVICE__BACKEND__PROTOCOL = http
[frontend_e61771b1-9]: SERVICE__BACKEND-DAPR__GRPC__PORT = 53423
[frontend_e61771b1-9]: BACKEND_HTTPS_SERVICE_PORT = 53419
[frontend_e61771b1-9]: BACKEND_HTTPS_SERVICE_HOST = localhost
[frontend_e61771b1-9]: SERVICE__FRONTEND__HOST = localhost
[frontend_e61771b1-9]: TERM_SESSION_ID = 147A436C-43F2-48DB-AE05-1CAE65830EBC
[...]

PR #696, opened 2020-10-11
Set DAPR_HTTP_PORT when launching project

rynowak:
Fixes: #575

Tye was missing code to set DAPR_HTTP_PORT when launching a project.
This variable is normally set by dapr when using `dapr run ...` to
launch a project. However Tye doesn't use `dapr run ...` so we need to
emulate this behavior.

We already had code to set up DAPR_GRPC_PORT which is used by the
`DaprClient` class. DAPR_HTTP_PORT is used by the actor runtime in
addition to behing available for arbitrary use (as reported in the
issue).

### DIFF
--- a/src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
+++ b/src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
@@ -139,7 +139,7 @@ namespace Microsoft.Tye.Extensions.Dapr
                     };
                     project.EnvironmentVariables.Add(daprGrpcPort);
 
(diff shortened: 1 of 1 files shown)


rynowak:
@jkotalik

davidfowl (approved):
LGTM. We should work out how to test this so we catch regression proactively with new versions

jkotalik (approved):

rynowak:
> LGTM. We should work out how to test this so we catch regression proactively with new versions

Add it to the list of things to create integration tests for?

PR #700, opened 2020-10-13
Add line break for better debug experience

davidhenley:
Without this you get `Registered .NET SDK.Loading project...` with no space or line break between the two statements when verbosity is set to `Debug`.


JunTaoLuo (approved):

JunTaoLuo:
Thanks!

Issue #683, opened 2020-09-30
Tye tries to use "core branded" docker image for .NET 5.0


eiximenis:
### Describe the bug
Using `tye build` or `tye deploy` for building or deploying a .NET5 project ends with error, because tye wants to use the `mcr.microsoft.com/dotnet/core/aspnet:5.0` docker image. However .NET5 docker images are under `mcr.microsoft.com/dotnet/aspnet:5.0` (no core).

### To Reproduce
Just create any project using NET 5 and use tye build or tye deploy.

Here is `tye build -v debug` sample output:

```
Restoring and evaluating projects
Resolved metadata for service webclient at D:\dotnet2020\src\WebClient\obj\Debug\net5.0\MicrosoftTye.ProjectMetadata.txtResolved metadata for service gamemanagerservice at D:\dotnet2020\src\GameManagerService\obj\Debug\net5.0\MicrosoftTye.ProjectMetadata.txt
Resolved metadata for service statisticsservice at D:\dotnet2020\src\StatisticsService\obj\Debug\net5.0\MicrosoftTye.ProjectMetadata.txt
Restore and project evaluation took: 2679.6643ms
Found application version: 1.0.0
RunCommand=D:\dotnet2020\src\WebClient\bin\Debug\net5.0\WebClient.exe
RunArguments=
TargetPath=D:\dotnet2020\src\WebClient\bin\Debug\net5.0\WebClient.dll
PublishDir=bin\Debug\net5.0\publish\
AssemblyName=WebClient
IntermediateOutputPath=obj\Debug\net5.0\
Found target framework: net5.0
Parsed target framework name: net
Parsed target framework version: 5.0
Found shared frameworks: Microsoft.NETCore.App, Microsoft.AspNetCore.App
IsAspNet=True
Evaluation Took: 4.3168ms
Found application version: 1.0.0
RunCommand=D:\dotnet2020\src\GameManagerService\bin\Debug\net5.0\GameManagerService.exe
RunArguments=
TargetPath=D:\dotnet2020\src\GameManagerService\bin\Debug\net5.0\GameManagerService.dll
PublishDir=bin\Debug\net5.0\publish\
AssemblyName=GameManagerService
IntermediateOutputPath=obj\Debug\net5.0\
Found target framework: net5.0
Parsed target framework name: net
Parsed target framework version: 5.0
Found shared frameworks: Microsoft.NETCore.App, Microsoft.AspNetCore.App
IsAspNet=True
[...]

PR #702, opened 2020-10-13
Resolve base docker images correctly for .NET 5.0+ projects

JunTaoLuo:
Fixes https://github.com/dotnet/tye/issues/683. Note that the naming scheme is dependent on the TFM of the project: https://hub.docker.com/_/microsoft-dotnet-core

### DIFF
--- a/src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs
+++ b/src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs
@@ -123,7 +123,7 @@ namespace Microsoft.Tye.Hosting
             serviceDescription.RunInfo = dockerRunInfo;
         }
 
-        private string DetermineContainerImage(ProjectRunInfo project)
+        private static string DetermineContainerImage(ProjectRunInfo project)
         {
             var baseImageTag = !string.IsNullOrEmpty(project.ContainerBaseTag) ? project.ContainerBaseTag : project.TargetFrameworkVersion;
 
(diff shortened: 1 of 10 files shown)


src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs
```diff
@@ -123,19 +123,38 @@ private async Task TransformProjectToContainer(Service service, ProjectRunInfo p
...
-                baseImage = project.IsAspNet ? "mcr.microsoft.com/dotnet/core/aspnet" : "mcr.microsoft.com/dotnet/core/runtime";
-            }
+                Version? baseImageVersion = null;
+                try
+                {
+                    baseImageVersion = new Version(baseImageTag);
+                }
+                catch
+                {
+                    _logger.LogInformation($"Could not determine version of docker base image for tag: {baseImageTag}");
```
jkotalik, line 144: Shouldn't you still let this throw?

jkotalik (approved):

src/Microsoft.Tye.Hosting/TransformProjectsIntoContainers.cs
JunTaoLuo: Curious if we should, currently it falls back to 2.1/3.1 but I dunno if that makes sense. If we throw we should probably advise selecting an explicit base image via the option.

JunTaoLuo: Actually, that sounds better to me. I'm gonna make it throw.

davidfowl (approved):

codeputer:
Awesome...can we pull this code via the update?

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

JunTaoLuo:
@codeputer Yes, you can get this change by installing our nightly, see https://github.com/dotnet/tye/blob/master/docs/getting_started.md#working-with-ci-builds

codeputer:
Thanks… I’m still getting this error on the microservice program…

This it the directory structure… I need to review/understand the dockerfile and why the copies are not working correctly – hints on where to look are appreciated.
[cid:image002.jpg@01D6A21E.FFB35630]

GeorgeS2019:
I still have problem getting[ the CI version](https://github.com/dotnet/tye/blob/master/docs/getting_started.md#working-with-ci-builds) to work with the  .NET5 (not .NETCore 3.1) version

PR #708, opened 2020-10-21
Add host option to enable remote access 

alexfdezsauco:
Add host option to enable remote access


jkotalik:
Sorry for the delay on reviewing this. 

Is this host option something that would apply to all services running?

Right now, I'm trying to figure out how this would play with bindings specific hosts, as I believe you discovered in this PR. Should bindings specific hosts be preferred to the one passed in on commandline?

alexfdezsauco:
Yes, Is something that would apply to all services. I think that have an issue, it doesn't count correctly when a service restart (not sure why).

The host passed in on command line should override all bindings (I think).

jkotalik:
Hm. I'm hesitant to accept this as a default option on commandline. It seems fairly niche and I'm concerned about defaulting to always using the passed in host.

@JunTaoLuo thoughts?

jkotalik:
After discussing internally, we don't think we'd take this PR at this time due to usability issues and the command line issue being niche. You can file an issue and we can continue discussing a better solution for this.

alexfdezsauco:
Ok, thanks. I'm using this in this way for demos & workshops. So, looks like I have to maintain my own fork of this to use --host feature.

PR #734, opened 2020-10-29
Fix resolution of TFM overrides for multi-tfm projects

JunTaoLuo:
The main issue I hit during verification was that the override logic we had didn't work as intended.

Imagine the following scenario: 
1. The project is cross-targeting netcoreapp3.1 and net5.0
2. There is configuration in tye.yaml that specifies the TargetFramework to be net5.0
3. There is a command line argument --framework net3.1

Since we want to optimize evaluation speed, we opted for batching. Ideally, we would be able to send all the information, the TFM from tye.yaml and the hint from the command line, to MSBuild and have it sort out which TFM to restore/build for during project evaluation. 

However, I spent some time looking into the SDK and it looks like there's no chance for us to customize the logic. The SDK determines whether it's a cross-targeting build immediately when it's imported and there's no customization available for us to hook into before that's done. As such I'm changing the approach here to evaluate projects in two batches. The first evaluate all projects and determines which are cross-targeting. The second batch evaluates all cross-targeting projects using the hint from the CLI if provided. 

As a fallout from that, I think the build properties in the tye.yaml should take precedence over cli args. I feel that the cli args should only be relevant if a cross-targeting project is lacking a TFM. I could be convinced otherwise here though.

As for testing, I've been testing with some of the scenarios in our tutorials with modifications for multi-tfm projects. I'm also looking to add a test for the scenario that wasn't covered when I found the bug.

### DIFF
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -83,77 +83,55 @@ namespace Microsoft.Tye
                 var projectServices = services.Where(s => !string.IsNullOrEmpty(s.Project));
                 var projectMetadata = new Dictionary<string, string>();
 
-                using (var directory = TempDirectory.Create())
+                var msbuildEvaluationResult = await EvaluateProjectsAsync(
+                    projects: projectServices, 
+                    configRoot: config.Source.DirectoryName!, 
+                    output: output);
+                var msbuildEvaluationOutput = msbuildEvaluationResult
+                    .StandardOutput
+                    .Split(Environment.NewLine);
+
+                var multiTFMProjects = new List<ConfigService>();
+
+                foreach (var line in msbuildEvaluationOutput)
                 {
-                    var projectPath = Path.Combine(directory.DirectoryPath, Path.GetRandomFileName() + ".proj");
-
-                    var sb = new StringBuilder();
-                    sb.AppendLine("<Project>");
-                    sb.AppendLine("    <ItemGroup>");
-
-                    foreach (var project in projectServices)
+                    var trimmed = line.Trim();
+                    if (trimmed.StartsWith("Microsoft.Tye metadata: "))
                     {
-                        var expandedProject = Environment.ExpandEnvironmentVariables(project.Project!);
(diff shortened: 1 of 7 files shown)


jkotalik:
> As a fallout from that, I think the build properties in the tye.yaml should take precedence over cli args. I feel that the cli args should only be relevant if a cross-targeting project is lacking a TFM. I could be convinced otherwise here though.

Hm, again to me it seems more intuitive for the commandline to override all, but I'm not that particular here.

jkotalik:
>The first evaluate all projects and determines which are cross-targeting. The second batch evaluates all cross-targeting projects using the hint from the CLI if provided.

Is there a perf hit with it?

JunTaoLuo:
Yes there's a perf hit since we are potentially doing two rounds of evaluation. It depends on how many cross-targeting projects there are. If there are none, there's no additional evaluation time. If every project is cross-targeting, the evaluation is doubled. 

An alternative is that we just scratch the CLI arg and have users specify the target framework in the tye.yaml only. If we do that, we'll just do one round of evaluation and fail with a message to specify the TFM in the tye.yaml. This presents two problems: 1. There could be a lot of projects that each need a TFM build property, which leads to a lot of repetition. 2. This enforces the use of tye.yaml, whereas currently you would be able to run using a sln file.

jkotalik:
Okay, I'm less worried if this only applies to if there are cross targets.

tebeco:
This pb is why I had trouble before the `MsBuild out of proc` and even more after.
One way would be to have an accessor to the CLI `args` and check for the presence of `--framework` before build properties are evaluated, and directly generates "intermediate" project for MsBuild with that value.
So that would avoid that extra Hoop maybe

It could be something like this (but it means overrding / mutating `TargetFrameworks`
Or/And rely on `Directory.Build.targets` to properly override the value (because it's too soon in `Directory.Build.props`)
```xml
<TargetFrameworks Condition=" '$(TfmFromArgs)' == '' ">xxxx</TargetFrameworks>
<TargetFramework Condition=" '$(TfmFromArgs)' != '' ">$(TfmFromArgs)</TargetFrameworks>
```

Issue #353, opened 2020-04-10
Open dashboard on browser after `tye run`


giggio:
### What should we add or change to make your life better?

After `tye run`, open the dashboard on a browser. Enabling this should be configurable, in a user setting and/or project setting.

### Why is this important to you?

It's one less step on the development workflow.

rynowak:
Thanks for the idea. It looks like someone's already working on the basic functionality. https://github.com/dotnet/tye/pull/350

> Enabling this should be configurable, in a user setting and/or project setting.

We should talk more about this part @jkotalik @davidfowl 

I like the idea of per-user options via env-vars as well as via command-line - allowing you to set env-vars to provide options like `--dashboard` makes things less repetative.

So like `$env:TYE_RUN_ARGS="--dashboard"` in your profile would toggle on this behavior all of the time.

rynowak:
There's a new release of the kubernetes client that hopefully has the fix.

rynowak:
We like the environment variable solution for making this configurable.

qpooqp:
Hello

I tried to solve this issue (initially just for myself to learn some stuff) and i think that i came up with decent solution which you could use. 

**Brief description**
It is generic solution to apply default options to commands.
Environment variable is choosen by convention as `TYE_{command name uppercase}_ARGS`.
It applies all non already applied options from env-var to command.
It can be disabled with option `--no-default`.

**Examples**
User creates env-var: `TYE_RUN_ARGS="--v debug --dashboard"`
- User runs command `tye run --v quiet` which will become `tye run --v quiet --dashboard`
Option `-v` is already applied so we will not reapply it. Option `--dashboard` is applied.
- User runs command `tye run` which will become `tye run --v debug --dashboard`
Both options are missing so we will apply them
- User runs command `tye run --no-default`  which stays as `tye run --no-default`
User specified the `--no-default` option which disables this feature
- User runs command `tye deploy`  which stays as  `tye deploy`
Env-var was not found for deploy command (expected `TYE_DEPLOY_ARGS`).

Is it something what you was thinking of? 
I would create a PR if you would like to check it out more.

giggio:
@qpooqp that would work for me!

PR #736, opened 2020-10-30
Added functionality to specify default options for commands

qpooqp:
This is functionality which allows users to specify default options for commands in environment variables and this default options will be automatically applied.

More description with some examples is in #353.

### DIFF
--- a/src/tye/Program.DefaultOptionsMiddleware.cs
+++ b/src/tye/Program.DefaultOptionsMiddleware.cs
@@ -1,0 +1,64 @@
+﻿using System;
+using System.Collections.Generic;
+using System.CommandLine;
+using System.CommandLine.Invocation;
+using System.CommandLine.Parsing;
+using System.Linq;
+
+namespace Microsoft.Tye
+{
+    public static partial class Program
+    {
+        private const string DefaultOptionsEnvVarPrefix = "TYE_";
+        private const string DefaultOptionsEnvVarPostfix = "_ARGS";
+
+        private static void DefaultOptionsMiddleware(InvocationContext context)
+        {
+            if (context.ParseResult.CommandResult.Parent != context.ParseResult.RootCommandResult)
+            {
+                return;
+            }
+
+            var commandName = context.ParseResult.CommandResult.Command.Name;
+
+            var rawDefaultOptions = Environment.GetEnvironmentVariable(DefaultOptionsEnvVarPrefix + commandName.ToUpper() + DefaultOptionsEnvVarPostfix);
+            if (string.IsNullOrWhiteSpace(rawDefaultOptions))
+            {
+                return;
+            }
+
+            var originalParseResult = context.ParseResult;
+            var originalOptionResults = GetCommandOptions(originalParseResult.CommandResult);
+            var originalTokens = StringifyTokens(originalParseResult.Tokens);
+
+            if (originalOptionResults.Any(option => option.Option.Name == StandardOptions.NoDefaultOptions.Name))
+            {
+                return;
+            }
+
+            var defaultParseResult = context.Parser.Parse($"{commandName} {rawDefaultOptions}");
+            var defaultOptionResults = GetCommandOptions(defaultParseResult.CommandResult);
+            var additionalTokens = GetAdditionalOptionsTokens(originalOptionResults, defaultOptionResults);
+
+            context.ParseResult = context.Parser.Parse($"{originalTokens} {additionalTokens}");
+
+            static string StringifyTokens(IEnumerable<Token> tokens)
+            {
+                return string.Join(" ", tokens.Select(t => t.Value));
+            }
+
+            static IEnumerable<OptionResult> GetCommandOptions(CommandResult commandResult)
+            {
+                return commandResult.Children.OfType<OptionResult>();
+            }
+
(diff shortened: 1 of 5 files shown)


src/tye/Program.DefaultOptionsMiddleware.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System;
```
jkotalik, line 1: nit: add copyright header.

src/tye/Program.DefaultOptionsMiddleware.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace Microsoft.Tye
+{
+    public static partial class Program
+    {
+        private const string DefaultOptionsEnvVarPrefix = "TYE_";
+        private const string DefaultOptionsEnvVarPostfix = "_ARGS";
+
+        private static void DefaultOptionsMiddleware(InvocationContext context)
+        {
+            if (context.ParseResult.CommandResult.Parent != context.ParseResult.RootCommandResult)
```
jkotalik, line 17: Is this a common check for these types of middleware? May be worth a comment why you check that.

src/tye/Program.DefaultOptionsMiddleware.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            var commandName = context.ParseResult.CommandResult.Command.Name;
+
+            var rawDefaultOptions = Environment.GetEnvironmentVariable(DefaultOptionsEnvVarPrefix + commandName.ToUpper() + DefaultOptionsEnvVarPostfix);
+            if (string.IsNullOrWhiteSpace(rawDefaultOptions))
+            {
+                return;
+            }
+
+            var originalParseResult = context.ParseResult;
```
jkotalik, line 30: I think an example here in comments would be useful. Took me a while to understand what was going on here.

src/tye/Program.DefaultOptionsMiddleware.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Linq;
+
+namespace Microsoft.Tye
+{
+    public static partial class Program
+    {
+        private const string DefaultOptionsEnvVarPrefix = "TYE_";
+        private const string DefaultOptionsEnvVarPostfix = "_ARGS";
+
+        private static void DefaultOptionsMiddleware(InvocationContext context)
```
jkotalik, line 15: Is it feasible to add a test for this middleware? Is there a mock invocation context you could have to parse?

src/tye/Program.DefaultOptionsMiddleware.cs
qpooqp: Well, not exactly.
I'm checking if the currently executing command is second level (probably not official name but i'm not sure how to name it) command. That means that the command is subcommand of root.
For example:
 - Input `tye run` is described like this
    - `tye` is root command
    - `run` is second level command
 - Input `tye build`
    - `tye` is root command
    - `build` is second level command

This check is not necessary at the moment. I was just trying to be safe for the future.
If there will be added third level command and the name of the command will be same as one of the second level (for example `tye build run` - `run` is third level and also second level command), then it may add unintended options. So i rather disable it for the moment.

If you know that there will not be third level commands, then i can remove this chcek. 
Othervise i will put some comment there.

src/tye/Program.DefaultOptionsMiddleware.cs
qpooqp: I think that i can test it. Will add some tests.

src/tye/Program.DefaultOptionsMiddleware.cs
qpooqp: I added some comments to the code, but here if brief description.

- Line 24 - Get value from env-var for current command - if nothing found then exit
- Line 31 - Get all applied options in original input
- Line 32 - Recreate original input
- L 34 - Check if option `--no-default` is applied in original options - if yes, then exit
Example: `tye run --no-default` - this will not apply default options
- L 39 - Parse default options to validate them
- L 40 - Get valid default options
- L 41 - Filter only the default options which are not already applied in original input
- L 43 - Set new parse result (original input plus default options)

qpooqp:
Can i have a question about the file `src/Microsoft.Tye.Core/StandardOptions.cs`?

The file is in project `Microsoft.Tye.Core`, but it is used only in project `tye`.
Shouldnt it be rather in project `tye`?

image

jkotalik:
> StandardOptions

Go ahead an try moving it, see if anything breaks 😄 . I'm fine to move it.

test/UnitTests/DefaultOptionsMiddlewareTests.cs
```diff
@@ -0,0 +1,130 @@
...
+using System.CommandLine.Builder;
+using System.CommandLine.Invocation;
+using System.CommandLine.IO;
+using System.CommandLine.Parsing;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.Tye.UnitTests
+{
+    public class DefaultOptionsMiddlewareTests
```
jkotalik, line 13: Nice tests!

jkotalik (approved):
LGTM besides a small nit to move file locations. Thanks!

jkotalik:
Thanks!

PR #738, opened 2020-10-31
Update ingress apiVersion now that it's not supposed to be beta anymore

### LINKED ISSUE #737: Update Ingress to v1

tye currently uses v1beta1, v1 is out.  Let's upgrade: 

`apiVersion: networking.k8s.io/v1`

Because v1 is now ready and I want to use the latest.

azsdke2e
azsdke2e1

tebeco:
fixes #737 

This PR is just a "find and replace" from `extensions/v1beta1` to `networking.k8s.io/v1` for the ingress part
@jongio can you confirm if I missed something ? It looks too simple to be just this in the PR and you might have seen something else ^^

I would also love double check if that URL here is to be updated too it look like URI from inside the cluster to reach the ingress(es)
```
                // We need to then wait for the webhooks that are created by ingress-nginx to start. Deploying an ingress immediately
                // after creating the controller will fail if the webhook isn't ready.
                //
                // Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": 
                // Post https://ingress-nginx-controller-admission.ingress-nginx.svc:443/networking.k8s.io/v1/ingresses?timeout=30s:
                // dial tcp 10.0.31.130:443: connect: connection refused
                //
                // Unfortunately this is the likely case for us.
```

### DIFF
--- a/samples/voting/ingress.yml
+++ b/samples/voting/ingress.yml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: ingress-basic
--- a/src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
+++ b/src/Microsoft.Tye.Core/KubernetesManifestGenerator.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Tye
             var root = new YamlMappingNode();
 
             root.Add("kind", "Ingress");
-            root.Add("apiVersion", "extensions/v1beta1");
+            root.Add("apiVersion", "networking.k8s.io/v1");
 
             var metadata = new YamlMappingNode();
             root.Add("metadata", metadata);
--- a/src/Microsoft.Tye.Core/ValidateIngressStep.cs
+++ b/src/Microsoft.Tye.Core/ValidateIngressStep.cs
@@ -162,7 +162,7 @@ namespace Microsoft.Tye
                 // after creating the controller will fail if the webhook isn't ready.
                 //
                 // Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": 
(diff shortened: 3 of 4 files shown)


jongio:
Nice, Thanks @tebeco 

The structure of the file also has breaking changes.  https://kubernetes.io/docs/concepts/services-networking/ingress/

So, need to update the Ingress file that is generated as well.

tebeco:
right i was a bit too optimist ^^

jkotalik:
Thanks @tebeco for this PR.

I think we may need to have some kubectl version checks here to determine what ingress type we should deploy. OR, we could just require that tye needs to use X version of kubernetes.

Can you check which version ingress was added out of beta (I think it 1.18, but not sure)? We can decide whether we need to support both ingress types based on that.

tebeco:
g

tebeco:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#ingress-v1-networking-k8s-io

tebeco:
it seems my description of links up it wrong 😑 
I have no idea why I have something about 1.14 🤣🙃

will update that after proper research

tebeco:
@jkotalik 

I'm moving out in the next weeks, I have very few time to take a dive at this one right now, 
I can let this PR like this as a Draft, but i kinda want to insist on the fact that if anyone want to re-create a PR from scratch, or anytnig else, please feels free to do so.

I don't want anyone to feel like i'm blocking anything here.

jkotalik:
We'll take over this PR now 😄

tebeco:
hello @jkotalik 

yes please, I should have insisted more on the fact that it's very difficult / nearly immpossible for me to continue on.
Feel free to either cherry pick / close / re-open (or continue as owner have access to it probably)

jkotalik:
closing in favor of https://github.com/dotnet/tye/pull/664

Issue #729, opened 2020-10-27
Ingress to AKS not working


jongio:
### Describe the bug

I'm not able to get ingress to AKS working.

### To Reproduce

Here's my tye.yaml ingress:

```yaml
ingress:
  - name: 1080
    bindings:
      - port: 1080
    rules:
      - path: /
        service: memealyzernetwebapp
  - name: 2080
    bindings: 
      - port: 2080
    rules:
      - path: /images
        service: memealyzernetapi
      - path: /image
        service: memealyzernetapi
      - path: /config
        service: memealyzernetapi
      - path: /imagehub
        service: memealyzernetapi
```

When I run `tye generate` I do not see any Ingress settings generated.

### Further technical details

- Include the output of `tye --version`
0.5.0-alpha.20520.1+739de2eeb94f0246f50fa4993e953abb4b0da18a
- If possible rerun the command with `-v debug` and include the output
- The platform (Linux/macOS/Windows)
WSL2

azsdke2e
azsdke2e0

jongio:
This works locally, but not for AKS.

Here's what I have in Ingress/tye.yaml
```yaml
ingress:
  - name: ingress
    rules:
      - path: /
        service: memealyzernetwebapp
      - path: /images
        service: memealyzernetapi
        preservePath: true
      - path: /image
        service: memealyzernetapi
        preservePath: true
      - path: /config
        service: memealyzernetapi
        preservePath: true
      - path: /imagehub
        service: memealyzernetapi
        preservePath: true
      - path: /api/negotiate
        service: memealyzernetfunction
        preservePath: true
    tags:
      - dev
      - staging
      - prod
```
Here's what is generated:
```yaml
kind: Ingress
apiVersion: extensions/v1beta1
metadata:
  name: ingress
  annotations:
    kubernetes.io/ingress.class: 'nginx'
    nginx.ingress.kubernetes.io/rewrite-target: '/$2'
  labels:
    app.kubernetes.io/part-of: 'memealyzer'
spec:
  rules:
  - http:
      paths:
      - backend:
          serviceName: memealyzernetwebapp
          servicePort: 1080
        path: /()(.*)
      - backend:
          serviceName: memealyzernetapi
          servicePort: 2080
        path: /()(.*)
      - backend:
          serviceName: memealyzernetapi
          servicePort: 2080
        path: /()(.*)
      - backend:
          serviceName: memealyzernetapi
          servicePort: 2080
        path: /()(.*)
      - backend:
[...]

PR #747, opened 2020-11-03
Adds preservePath for ingress

jkotalik:
Porting from https://github.com/dotnet/tye/pull/463. This will keep the path if a path is given for a particular redirect rule.

Fixes https://github.com/dotnet/tye/issues/729


JunTaoLuo (approved):
We don't have tests for these kind of changes?

jkotalik:
We could add a test, but this is blocking @jongio, so I'll file an issue for follow up.

PR #752, opened 2020-11-04
Fix nginx ingress path for capture group.

jkotalik:
Small mistake in https://github.com/dotnet/tye/commit/df18a1d31bac543de05c498b1f2054031157a3c2, I forgot that we need to specify the path in the capture group.


jkotalik:
Verified this is working.

davidfowl:
Oops 😄

JunTaoLuo:
Looks good, post hoc :shipit:

PR #755, opened 2020-11-05
Updated Voting sample deployment steps

sughosneo:
Improved Voting sample deployment steps by : 

- Added missing `postgres.yaml` for easy access. 
- Updated `Readme.md` to capture all the steps.


samples/voting/README.md
```diff
@@ -9,11 +9,44 @@ The project should be immediately runnable by calling `tye run` from the directo
...
 A few things need to be configured before deploying to Kubernetes.
 
-- Setting up Redis. A connection string needs to be provided to connect to Redis. You can follow our tutorial on [setting up redis in your cluster](../../docs/tutorials/hello-tye/02_add_redis.md).
-- Setting up postgresql. A connection string eventually needs to be provided to tye for postgresql.
-- Deploying the ingress.yaml by calling:
+- [Create an Azure container registry using the Azure portal](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-get-started-portal) or you can use your public [docker hub account](https://hub.docker.com/)
 
+- `Redis` can be deployed using the below command :
+
+    ```powershell
```
davidfowl, line 16: Why powershell?

sughosneo: Hi @davidfowl , 
Yes, instead of marking it as specific shell. It would be better to use only "triple backticks" to show the command text. I have made the changes and updated the PR for your further review.

Thanks.

samples/voting/README.md
```diff
@@ -9,11 +9,44 @@ The project should be immediately runnable by calling `tye run` from the directo
...
     ```
+
+- Once the deployment is done, you need to keep a note of the below connection strings :
+
+    -  `redis:6379`
+    -  `Server=postgres;Port=5432;User Id=postgres;Password=pass@word1;`
+
+>! NOTE: If you want, you can modify the password in `postgres` yaml.
+
+- You need to install [NGINX Ingress Controller](https://kubernetes.github.io/ingress-nginx/) in your Kubernetes cluster.
```
jkotalik, line 32: Actually, we should be able to remove this step here, as tye now supports deploying ingresses, including installing the NGINX ingress controller. Can you do the following:
- Remove the ingress.yaml and translate it to the equivalent tye.yaml ingress definition. See https://github.com/dotnet/tye/blob/master/samples/apps-with-ingress/tye.yaml as an example.
- Remove these bullets about deploying the ingress.

samples/voting/README.md
```diff
@@ -9,11 +9,44 @@ The project should be immediately runnable by calling `tye run` from the directo
...
-- Deploying the ingress.yaml by calling:
+- [Create an Azure container registry using the Azure portal](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-get-started-portal) or you can use your public [docker hub account](https://hub.docker.com/)
+
+- `Redis` can be deployed using the below command :
 
     ```
-    kubectl apply -f ingress.yml
+    kubectl apply -f https://raw.githubusercontent.com/dotnet/tye/master/docs/tutorials/hello-tye/redis.yaml
+    ```
+- `Postgresql` can be installed by the following command :
```
jkotalik, line 19: Can you also mention that you can install postgres and redis with an external service, like a managed redis instance?

samples/voting/README.md
```diff
@@ -9,11 +9,44 @@ The project should be immediately runnable by calling `tye run` from the directo
...
+- You need to install [NGINX Ingress Controller](https://kubernetes.github.io/ingress-nginx/) in your Kubernetes cluster. 
+  You can also use a package manger like `helm` to [create an ingress controller](https://docs.microsoft.com/en-us/azure/aks/ingress-basic#create-an-ingress-controller) to expose your application outside of your Kubernetes cluster. 
+  
+  And then you need to expose those services by using ingress yaml. 
     
+    ```
+    kubectl apply -f ingress.yml
+    ```
+
+- After that, run `tye deploy --interactive --namespace default` to do all the deployment.
```
jkotalik, line 41: Why --namespace default?

samples/voting/README.md
```diff
@@ -9,11 +9,44 @@ The project should be immediately runnable by calling `tye run` from the directo
...
+
+>! NOTE: Fill in the connection string for both `redis` and `postgres` when it's prompted.
+
+Once the deployment is complete you should be able to browse the apps by the below Urls:
+
+- vote : http://\<Cluster Public IP Or DNS\>/vote
```
jkotalik, line 47: I'd recommend them running kubectl get ingress to get the IP. 

Also maybe should add a note about how the ingress controller may take awhile to update the public IP listed.

jkotalik (approved):
Some nits, but overall good.

docs/tutorials/hello-tye/postgres.yaml
```diff
@@ -0,0 +1,47 @@
...
+      labels:
+        app.kubernetes.io/name: postgres
+        app.kubernetes.io/part-of: microservice
+    spec:
+      containers:
+        - name: postgres
+          image: postgres
+          resources:
+            requests:
+              cpu: 100m
```
jkotalik, line 24: Why these limits specifically?

samples/voting/README.md
sughosneo: @jkotalik - Yes, correct. Have done the following changes : 

- Removed `ingress.yaml` and updated  `tye` yaml  to deploy ingress. 
- Removed the ingress deploy section.

samples/voting/README.md
sughosneo: Rephrased the `redis` and `postgres` installation steps to include the information related to managed service.

samples/voting/README.md
sughosneo: Well this strange while using `tye deploy --interactive` I am getting below error while deploying to both local k8s and AKS. 

image

But when I provide the namespace value everything seems to be fine. That's the reason I had included that `--namespace` parameter.

samples/voting/README.md
sughosneo: Yes, agreed. Updated this step.

docs/tutorials/hello-tye/postgres.yaml
sughosneo: This is to set requested configuration for the container. That's the reason it has been set as `request` not as `limit`. This is also to keep it synced with `redis.yml`. Do you want me to remove that ? Thoughts ?

sughosneo:
> Some nits, but overall good.

Hi @jkotalik , thanks for your review and comments. I have tried to incorporate most of the changes. But windows build is failing for the last commit. When I checked the log it says  `XUnit : error : Tests failed: F:\workspace\_work\1\s\artifacts\log\Release\Microsoft.Tye.E2ETests_netcoreapp3.1_x64.log [netcoreapp3.1|x64] [F:\workspace\_work\1\s\test\E2ETest\Microsoft.Tye.E2ETests.csproj]`

When I ran test `E2ETest` in my local I could see that one of the test case has failed with following error - `E2ETest.TyeRunTests.SingleProjectWithDocker_UsesCorrectBaseImage(projectName: "single-project-5.0", baseImage: "mcr.microsoft.com/dotnet/aspnet:5.0")` that could be because I don't have .NET 5 runtime in my local. Are these related by any chance ? Thoughts ?

samples/voting/tye.yaml
```diff
@@ -1,4 +1,13 @@
 name: VotingSample
+ingress:
+  - name: ingress
+    bindings:
+      - port: 8080
+    rules:
+      - path: /vote(/|$)(.*)
```
davidfowl, line 7: Did you test this?

sughosneo: Yes, @davidfowl , I have tested this in my local Kubernetes cluster. Please find below details for your reference : 

image

And then browsed the sample with 

image

Is there anything missing in the `tye.yaml`, which I should consider ?

Thanks.

davidfowl: Did you run it locally with tye run?

sughosneo: No, I used deploy option `tye deploy --interactive --namespace default` to deploy this sample in my local Kubernetes cluster. Also, please note, I had to use `namespace` parameter. Without that I was getting following error.

image


`

samples/voting/README.md
jkotalik: Ah, did you set your current namespace in your kube context to teractive?

docs/tutorials/hello-tye/postgres.yaml
jkotalik: I would personally just remove these limits.

jkotalik (approved):

samples/voting/tye.yaml
jkotalik: Can you double check this works with tye run locally as well?

sughosneo: I have tested the sample with tye version `0.5.0-alpha` and it works fine in local.

docs/tutorials/hello-tye/postgres.yaml
sughosneo: Removed the resource limits from the `postgres.yaml`

samples/voting/README.md
sughosneo: No, actually I found out that kubectl current context wasn't even set to any namespace. That's the reason it was throwing that error. I am not sure from where the name `teractive` was coming up in that error message. But when I set my current kubectl context to namespace `default` then I was able to deploy the sample with `tye deploy --interactive`. 

I have updated the `Readme` file as well. If you could review that again.

samples/voting/README.md
sughosneo: Updated this step already.

samples/voting/tye.yaml
sughosneo: Not sure why the build fails on this PR. Any thoughts on that ?

JunTaoLuo (approved):

samples/voting/tye.yaml
davidfowl: AFAIK we don't support regex like this in the rules in the local ingress

samples/voting/README.md
sughosneo: After resetting my local kubernetes cluster I was able to deploy it to the local K8s without any issue with help of `tye deploy -i` command.  So it seems to be the problem of my local environment. Updated the readme file.

samples/voting/tye.yaml
sughosneo: I see, updated `tye.yaml`. Thank you for the clarification.

jkotalik:
Thanks!

PR #763, opened 2020-11-08
Add note about Dapr placement container requirement

dasiths:
The Tye documentation for Dapr integration (https://github.com/dotnet/tye/blob/master/docs/recipes/dapr.md) says all that's required is
```
extensions:
- name: dapr
```

but I need to also add this to the tye.yaml file to make it work
```
- name: placement
  image: daprio/dapr
  args: ./placement
  bindings:
    - port: 50005
```

I updated the documentation but please review and make sure it's correct.

### DIFF
--- a/docs/recipes/dapr.md
+++ b/docs/recipes/dapr.md
@@ -93,6 +93,17 @@ extensions:
 - name: dapr
 ```
 
+:warning: If you're running Dapr in Windows you will need to make sure the Dapr placement container is exposed via port 50005.
+```yaml
+# To ensure that your are running a dapr placement container with the right binding port.
+# (50005 as HostPort)
+- name: placement
+  image: daprio/dapr
+  args: ./placement
+  bindings:
+    - port: 50005
+```
+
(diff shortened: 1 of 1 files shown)


davidfowl:
That seems like we should just fix that bug

dasiths:
> That seems like we should just fix that bug

I can raise it as a bug here but wasn't sure it was Tye or a Dapr issue. The current Dapr recipe sample in the Tye code base already has that yaml snippet included.

dasiths:
I think I know how to fix it. Let me give it a crack.

dasiths:
@davidfowl I created a PR and referenced this. Will close this now.

jkotalik:
@dasiths I didn't have this issue when I was using the dapr sample in our repo locally. 

image

What tye version are you on? Is this issue resolved when you update to the latest release?

jkotalik:
Nvrm, I didn't realize we updated our sample to include the placement service on a port now.

PR #767, opened 2020-11-09
[Dapr Extension] Allow defining custom placement container address

dasiths:
As referenced in https://github.com/dotnet/tye/pull/763

When running Dapr in Windows (with WSL2 backend) the Dapr placement container runs on 6050 port on the host. So it requires we do something like this in `Tye.yaml` (Example https://github.com/dotnet/tye/blob/master/samples/dapr/tye.yaml)

```yaml
# To ensure that your are running a dapr placement container with the right binding port.
# (50005 as HostPort)
- name: placement
  image: daprio/dapr
  args: ./placement
  bindings:
    - port: 50005
```

Edit: With @jkotalik suggestions the PR now does the following

- Creates a Dapr placement container automatically and uses it if one is not defined by you.
- Allows you to instruct Tye to use a particular host port for the above.
- Allows you to instruct Tye to reuse an existing Dapr placement container that's already running.
```yaml
extensions:
- name: dapr

  # If you want to use a different tag or container port
  # placement-image: daprio/dapr
  # placement-container-port: 50005

  # You can instruct Tye to not create the Dapr placement container on your behalf. This is required if you have Dapr running and want to use that container.
  # Doing a `docker ps` can show if its already running. If it's running then you can set 'exclude-placement-container: true' with `placement-port: xxxx` set to the host port of that container.
  # (i.e. In Windows + WSL2, Dapr uses 6050 as the host port)

  exclude-placement-container: true
  placement-port: 6050
```


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/tye?pullRequest=767) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/tye?pullRequest=767) before we can accept your contribution.<br/><br/>:x: dasiths [sign now](https://cla.dotnetfoundation.org/dotnet/tye?pullRequest=767)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/tye?pullRequest=767) it.</sub>

jkotalik:
Is the placement container something we always need to be running with dapr now? If so, I think always creating the placement service and allowing someone to specify the service to override settings may be better, rather than having `placement-address`.

@davidfowl thoughts?

dasiths:
> Is the placement container something we always need to be running with dapr now? If so, I think always creating the placement service and allowing someone to specify the service to override settings may be better, rather than having `placement-address`.
> 
> @davidfowl thoughts?

I'm not sure which side of the boundary the placement container instantiation ownership should sit. I feel like if it's already there created and running by Dapr we should just use it. Now that I think about it, some sort of auto detection would be a lot smarter.

jkotalik:
To sum up our thoughts:
- If there isn't a placement container in the tye.yaml, we should create one to be used in the dapr extension. This matches what we have for the dapr proxies.
- If there is a placement container, just use it.

The only question now is auto detection of the placement container. Probably just checking for the image name daprio/dapr ./placement. Also we could add an argument to the extension called `excludePlacementContainer` to avoid adding it.

dasiths:
> * If there isn't a placement container in the tye.yaml

I think we shouldn't have to define a placement container in tye.yaml in either case. If there is a way to elegantly detect the existing container instance then we simply use that, or we spin up our own.

I'm not sure if the current design allows for a certain node in yaml (i.e. extensions) to alter another (services) when hydrating the internal models. I haven't looked at the codebase for that part closely yet.

We can then have more fine grained control by using `exclude-placement-container` or `placement-address` to reuse something running on a different port perhaps. Thoughts?

dasiths:
@jkotalik  Let's consider these scenarios 
- Dapr placement not running at all. So we create it and point to port 50005 (or pick port dynamically) and use it.
- Dapr placement is defined in Tye.yaml. So we use it.
- Dapr is already running placement container on 50005 or 6050. We need to either auto detect that and use it or allow the user to specify that in Tye.yaml (My original idea with `placement-address`).

If you can verify my assumptions are correct I can try and make the necessary changes.

dasiths:
@jkotalik I've updated the PR to cover the scenarios captured in the discussion and also included a sample app to demonstrate it.

I can move the sample app to a separate PR if required. The magic in this PR happens in the `src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs`

The build was successful till my last commit but it's broken now due to something else in the pipeline not related to this PR. Once that is fixed the PR will pass the build and tests.

Please let me know if there are any more changes required. :)

jkotalik:
@dasiths yes I think you assumptions were correct! Apologies for not confirming earlier.

jkotalik:
@dasiths yes I think you assumptions were correct! Apologies for not confirming earlier.

Looks like potentially our format checker is affected by the set_env issue in github actions. I'll look into that.

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
```diff
@@ -20,6 +20,61 @@ public override Task ProcessAsync(ExtensionContext context, ExtensionConfigurati
 
             if (context.Operation == ExtensionContext.OperationKind.LocalRun)
             {
+                // default placement port number
+                var placementPort = 50005;
+                var isCustomPlacementPort = false;
+
+                var existingDaprPlacementDefinition = context.Application.Services.FirstOrDefault(s =>
+                    s is ContainerServiceBuilder serviceBuilder && serviceBuilder.Image == "daprio/dapr");
```
mtaghavi2005, line 28: In case of using ImageName + Tag, placement definition cannot be found.
image

jkotalik: The more I think about this, the more I don't like sniffing to determine if the placement port is already defined or not.

I think including an option called `excludePlacementService`, which default to false, is enough to have people either include or exclude it as required.

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
```diff
@@ -20,12 +20,67 @@ public override Task ProcessAsync(ExtensionContext context, ExtensionConfigurati
 
             if (context.Operation == ExtensionContext.OperationKind.LocalRun)
             {
+                // default placement port number
+                var placementPort = 50005;
```
jkotalik, line 24: Can this be a random port instead of a specific port?

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
```diff
@@ -20,12 +20,67 @@ public override Task ProcessAsync(ExtensionContext context, ExtensionConfigurati
...
+                          obj?.ToString() is string excludePlacementContainer && excludePlacementContainer == "true"))
+                    {
+                        context.Output.WriteDebugLine("Injecting Dapr placement service...");
+                        var daprPlacement = new ContainerServiceBuilder("placement", "daprio/dapr")
+                        {
+                            Args = "./placement",
+                            Bindings = {
+                                new BindingBuilder() {
+                                    Port = placementPort,
+                                    ContainerPort = 50005,
```
jkotalik, line 53: Is the container port always supposted to be 50005 for the placement?

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
```diff
@@ -20,12 +20,67 @@ public override Task ProcessAsync(ExtensionContext context, ExtensionConfigurati
...
+                    }
+                    else
+                    {
+                        context.Output.WriteDebugLine("Skipping injecting Dapr placement service because 'exclude-placement-container=true'...");
+                    }
+                }
+                else
+                {
+                    // use the port defined in the custom service definition is possible
+                    var definedPlacementPort = existingDaprPlacementDefinition.Bindings.FirstOrDefault(b => b.Port.HasValue);
```
jkotalik, line 68: Based on my suggestion earlier, I think if someone sets `excludePlacementService`, they must define placement-port so we know what that is.

jkotalik:
Sample looks great FYI.

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
```diff
@@ -51,12 +106,12 @@ public override Task ProcessAsync(ExtensionContext context, ExtensionConfigurati
 
                         // These environment variables are replaced with environment variables
                         // defined for this service.
-                        Args = $"-app-id {project.Name} -app-port %APP_PORT% -dapr-grpc-port %DAPR_GRPC_PORT% --dapr-http-port %DAPR_HTTP_PORT% --metrics-port %METRICS_PORT% --placement-address localhost:50005",
+                        Args = $"-app-id {project.Name} -app-port %APP_PORT% -dapr-grpc-port %DAPR_GRPC_PORT% --dapr-http-port %DAPR_HTTP_PORT% --metrics-port %METRICS_PORT% --placement-address localhost:{placementPort}",
```
mtaghavi2005, line 109: That would be great this issue #802 also could be addressed here.

thangchung: Should be `--placement-host-address`, instead of `--placement-address`

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
dasiths: I have removed sniffing. It will now use an auto assigned port and be agnostic about the user defined services. The only way to control the port is via the `placement-port` property.

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
dasiths: It will now use the next available port, same as the other parts of the system.

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
dasiths: Yes it's defined by Dapr. I did however introduce a new property called `placement-container-port` to allow the developer to set it if required. (i.e. If Dapr releases a new image with a new port). Similarly I now have a `placement-image` which allows the developers to define a custom image name if required. (i.e. `daprio/dapr:latest`)

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
dasiths: Yes I added this check now. It will ignore `exclude-placement-service` if not `placement-port` has been defined.

src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs
dasiths: @mtaghavi2005 @thangchung I incorporated this change here. Should now support Dapr rc1. :)

dasiths:
@jkotalik I have addressed all of the comments. Please let me know if you're happy with this revision. Cheers.

dasiths:
@jkotalik any feedback about this?

src/Microsoft.Tye.Hosting/PortAssigner.cs
```diff
@@ -30,24 +30,12 @@ public Task StartAsync(Application application)
...
-                    return ((IPEndPoint)socket.LocalEndPoint!).Port;
-                }
-
                 foreach (var binding in service.Description.Bindings)
                 {
                     // Auto assign a port
                     if (binding.Port == null)
                     {
-                        binding.Port = GetNextPort();
+                        binding.Port = NextPortFinder.Instance.GetNextPort();
```
jkotalik, line 38: Why have NextPortFinder as an instance rather than just a static method?

jkotalik (approved):
LGTM besides a few nits.

src/Microsoft.Tye.Core/NextPortFinder.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Net;
```
jkotalik, line 1: nit copyright

jkotalik:
@dasiths apologies, holidays in the states 😄

src/Microsoft.Tye.Hosting/PortAssigner.cs
dasiths: I followed the pattern from the `GitDetector` class but more than happy to make it static as it doesn't have any dependencies or logic in the constructor.

src/Microsoft.Tye.Core/NextPortFinder.cs
dasiths: done

dasiths:
> @dasiths apologies, holidays in the states 😄

No worries. Hope you had a good break Justin. Looking forward to holidays myself. It's another sweltering 40c December awaiting us in downunder. 😉 

I've updated the PR with the changes now.

dasiths:
@jkotalik Anything else required here?

thangchung:
How was it going on, guys? Look like `tye` wasn't upgrade accordingly with `dapr rc` right away so that it makes it hard to use `dapr` with `tie` respectively :(

jkotalik:
Thanks @dasiths!

PR #777, opened 2020-11-11
Update ApplicationFactory.cs

sirh3e:
Refactored while loop.
Use standard function


davidfowl:
Why? The older code was better.

sirh3e:
Why was the old code better in which sense?
There is a standard function for this behavior.

davidfowl:
Oh my bad, I thought the old code used TryDeque, this looks good 😄

sirh3e:
No worries

jkotalik (approved):

PR #779, opened 2020-11-11
Update CombineStep.cs

sirh3e:
Replaced string == "http" by Equals("http")

See this blog post: https://docs.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings#default-ordinal-comparisons


davidfowl:
This won't work because == also does a null check.

sirh3e:
Fair point

sirh3e:
But i must say that BindingBuilder.Protocol can be null makes no sense. You always use a protocol.

sirh3e:
And the class BindingBuilder is not a builder at all it is more likey a container.

PR #786, opened 2020-11-13
Refactored ConfigFileFinder.cs 

sirh3e:


src/Microsoft.Tye.Core/ConfigFileFinder.cs
```diff
@@ -23,12 +23,12 @@ public static bool TryFindSupportedFile(string directoryPath, out string? filePa
...
                 }
 
-                if (files.Length > 1)
-                {
-                    errorMessage = $"More than one matching file was found in directory '{directoryPath}'.";
-                    filePath = default;
-                    return false;
-                }
+                if (files.Length <= 1)
+                    continue;
```
jkotalik, line 27: nit prefer braces.

jkotalik (approved):

sirh3e:
Added braces

dasiths:
There is something wrong with the build pipeline. The Build Windows task timed out on this PR as well. (This happened with with #767 too).

src/Microsoft.Tye.Core/ConfigFileFinder.cs
```diff
@@ -23,12 +23,14 @@ public static bool TryFindSupportedFile(string directoryPath, out string? filePa
                     return true;
                 }
 
-                if (files.Length > 1)
+                if (files.Length <= 1)
```
JunTaoLuo, line 26: Isn't this condition only applicable when the length is 0? If so let's just make it `if (files.Length == 0)`

JunTaoLuo (approved):
One nit. Otherwise ready to merge.

sirh3e:
@JunTaoLuo Yeah, you're right. It was late when I made the change. 😄

src/Microsoft.Tye.Core/ConfigFileFinder.cs
```diff
@@ -8,27 +8,27 @@ namespace Microsoft.Tye
...
+        private static readonly string[] FileFormats = { "tye.yaml", "tye.yml", "*.csproj", "*.fsproj", "*.sln" };
 
         public static bool TryFindSupportedFile(string directoryPath, out string? filePath, out string? errorMessage)
         {
             foreach (var format in FileFormats)
             {
                 var files = Directory.GetFiles(directoryPath, format);
 
-                if (files.Length == 1)
+                switch (files.Length)
```
jkotalik, line 19: nit: move >1 case into default for switch statement.

PR #800, opened 2020-11-17
Initial docker-compose.yaml support

jkotalik:
Provides an initial implementation of docker-compose support.

Here is what is supported now:
- build element -> project
- images
- ports
- environment variables

The sample almost works, however we need to find a solution to connectionstring in the dockercompose format.

Other pain points include:
- tags vs labels
- extensions (should be able to have a section called x-extensionname)

Here is an example of what the docker compose file format would look like for the voting app;
```
version: "3"
services:
  redis:
    image: redis:alpine
    ports:
      - "6379"
  postgres: # TODO needed to switch name here for some reason
    image: postgres:9.4
    ports:
      - "5432"
    environment:
      POSTGRES_PASSWORD: "postgres"
  vote:
    build:
      context: ./vote
    ports:
      - 5000:80
  result:
    build:
      context: ./results
    ports:
      - 5001:80
  worker:
    build:
      context: ./worker
```


philliphoff:
One note, as I assume you're using `build.context` to locate the project file; as soon as the application becomes large enough to accumulate one or more common libraries shared between services, the Docker build contexts for each service will shift from the project directory to the solution directory (else the services could not be built while the Docker image is generated).  

Visual Studio, for example, always scaffolds Dockerfiles and Docker Compose YAML files assuming a solution-level build context for projects, given the likelihood of them using a shared library.

jkotalik:
> One note, as I assume you're using build.context to locate the project file; as soon as the application becomes large enough to accumulate one or more common libraries shared between services, the Docker build contexts for each service will shift from the project directory to the solution directory (else the services could not be built while the Docker image is generated).

> Visual Studio, for example, always scaffolds Dockerfiles and Docker Compose YAML files assuming a solution-level build context for projects, given the likelihood of them using a shared library.

Yeah that's what I'm currently doing, finding the first proj file in the context directory and using that.

Interesting point on the context; where would dockerfiles exist in that model? If we had the following structure, would dockerfiles live in each of the services at the sln level? And the build.context would be set to the sln level?
```
docker-compose.yaml
app.sln
serviceA/
    Dockerfile
serviceB/
    Dockerfile
sharedlib/
    
```

philliphoff:
Visual Studio scaffolds Dockerfiles alongside the project file, but writes them assuming a solution-level context.  So, for example, all of the paths for COPY/ADD statements in the Dockerfile are relative to the solution folder.  So, at least for VS-scaffolded files, using the `build.dockerfile` would likely work better as a hint for the project file.  That said, users are known to move Dockerfiles around--some prefer them at the solution level or in a separate deployment folder--so it's not a given.

jkotalik:
I'd rather not use dockerfile as the source of truth though, as the property name is confusing to anyone who is wanting to use tye without knowing about containers or docker files.

> The surest way might be a build.x-projectfile property if that were allowed by the new extensibility.

I didn't realize the x-<name> could be done at any depth. That seems like a much better path for a lot of this.

philliphoff:
Note that I don't *know* that `build.x-projectfile` is allowed.  (If it isn't, it seems like it ought to be.) It was just an idea.

philliphoff:
It looks like extension fields are allowed (as of 3.7) at the root of the service, which should be an ok place for a `x-projectfile` property.

jkotalik:
> It looks like extension fields are allowed (as of 3.7) at the root of the service, which should be an ok place for a x-projectfile property.

Cool, I'll give it a shot.

src/Microsoft.Tye.Core/ConfigModel/NameInferer.cs
```diff
@@ -21,7 +21,7 @@ internal static class NameInferer
                 return Path.GetFileNameWithoutExtension(fileInfo.Name).ToLowerInvariant();
             }
 
-            return fileInfo.Directory?.Parent?.Name.ToLowerInvariant();
+            return fileInfo.Directory?.Name.ToLowerInvariant();
```
JunTaoLuo, line 24: Why did this change?

src/Microsoft.Tye.Core/DockerCompose/DockerComposeParser.cs
```diff
@@ -0,0 +1,455 @@
...
+
+        public ConfigApplication ParseConfigApplication()
+        {
+            try
+            {
+                _yamlStream.Load(_reader);
+            }
+            catch (YamlException ex)
+            {
+                throw new TyeYamlException(ex.Start, "Unable to parse tye.yaml. See inner exception.", ex);
```
JunTaoLuo, line 46:
```suggestion
                throw new TyeYamlException(ex.Start, "Unable to parse docker-compose.yaml. See inner exception.", ex);
```

JunTaoLuo (commented):
Taking an initial look. I'm going to pull this branch and try a few things out.

src/Microsoft.Tye.Core/ConfigModel/NameInferer.cs
jkotalik: This was incorrect, it should be the current directory name for the name of the app, rather than the parent.

jkotalik:
> Taking an initial look. I'm going to pull this branch and try a few things out.

We can take that path, or we can merge this in and start working on PRs based on master. Whatever makes you most productive.

src/Microsoft.Tye.Core/DockerCompose/DockerComposeParser.cs
```diff
@@ -0,0 +1,455 @@
...
+                    binding.ContainerPort = int.Parse(portString);
+                }
+
+                // TODO how to specify protocol with docker compose. Using http for now.
+                binding.Protocol = "http";
+                service.Bindings.Add(binding);
+            }
+        }
+
+        private static void ParseVolumes(YamlMappingNode node, ConfigApplication app)
```
JunTaoLuo, line 342: Nit: This and ParseNetworks are not used yet.

JunTaoLuo (commented):
Looks pretty good so far. We can either merge as is and add more docker-compose support or iterate on this branch, your call since it's your PR.

src/Microsoft.Tye.Core/DockerCompose/DockerComposeParser.cs
jkotalik: Yeah there are a few here that aren't used, but I'd rather get all of the elements defined in this file and we can fill them in later.

jkotalik:
Let's merge and continue iterating in master/main.

PR #811, opened 2020-11-21
Refactore pipeline

sirh3e:


jkotalik:
@sirh3e can you please combine all of your changes into a single PR instead of multiple PRs? Will be much easier for use to review.

sirh3e:
@jkotalik sure but pls give one more day 😀

davidfowl:
@jkotalik are you sure it'll be easier? These are all non functional changes...

jkotalik:
Yeah I guess you are right.

src/Microsoft.Tye.Core/Pipeline.cs
```diff
@@ -38,9 +38,9 @@ public static void Configure<TApplication>(Action<TApplication> callback)
             foreach (var (type, delegates) in callbacks)
             {
                 var method = pipeline.GetType().GetMethod("Register")!.MakeGenericMethod(type);
-                for (var i = 0; i < delegates.Count; i++)
+                foreach (var @delegate in delegates)
```
jkotalik, line 41: nit: `@delegate` to `del`

sirh3e: Why?

jkotalik: Personally not a fan of using `@` variables.

sirh3e: Delegate is much cleaner then del. Del could mean to much

PR #822, opened 2020-11-21
Refactore console extensions

sirh3e:


src/Microsoft.Tye.Core/ConsoleExtensions.cs
```diff
@@ -39,14 +41,12 @@ public static void SetTerminalForegroundColor(this IConsole console, ConsoleColo
                 ((dynamic)console).ForegroundColor = color;
             }
 
-            if (IsConsoleRedirectionCheckSupported &&
-                !Console.IsOutputRedirected)
+            switch (IsConsoleRedirectionCheckSupported)
```
jkotalik, line 44: Isn't checking for Console.IsOutputRedirected redundant here? Remove the switch and just have an if for IsConsoleRedirectionCheckSupported

JunTaoLuo: Agreed

Issue #766, opened 2020-11-09
Evaluated project metadata file could not be found for service transactions


nickraphael:
Windows 10
Was was happily running tye last week.  Then avast antivirus seemed to mess it up and I had to reinstall.  Now I'm getting this error when I do 'tye run'...

Evaluated project metadata file could not be found for service transactions

### To Reproduce
Latest version.  Tye Run

PascalSenn:
@nickraphael  
It seems like the problem are the `/obj` files

fergalmoran:
> @nickraphael
> It seems like the problem are the `/obj` files.
> Try to remove these folders

This worked for me 😊

newuser:
Is it parsing for unsuccessful build exit code? Tutorial instruction has failed at this point because some error occured due environment issues.

jkotalik:
Which version of tye are you running in this example? Do you have a full stack trace available?

newuser:
Perhaps it is because of different system language, version was master branch.

syska:
Getting the same issue ... 

One simple project ... 

```
# tye application configuration file
# read all about it at https://github.com/dotnet/tye
#
# when you've given us a try, we'd love to know what you think:
#    https://aka.ms/AA7q20u
#
name: all
services:
- name: yyy-proxy
  project: yyy.Proxy/yyy.Proxy.csproj

```

```
tye run -v Debug
Loading Application Details...
Restoring and evaluating projects
Evaluating project failed with exit code 1:
Ouptut: MSBUILD : error MSB1008: Only one project can be specified.
Switch: Syska\.dotnet\tools\.store\microsoft.tye\0.5.0-alpha.20555.1\microsoft.tye\0.5.0-alpha.20555.1\tools\netcoreapp3.1\any\ProjectEvaluation.targets

For switch syntax, type "MSBuild -help"

Error:
Restore and project evaluation took: 326,3943ms
Drats! 'run' failed:
        Evaluated project metadata file could not be found for service yyy-proxy
```

I do wonder about the `Switch` part of the above ... 

It specified `Syska\.dotnet` ... but the real path is: `C:\Users\Mikael Syska\.dotnet\tools` ... notice the `space` in my name ... yes, it's my real name ... 

I'm using .NET 5 if it matters. Let me know if you need more details ...

Werner-Swart-83:
Followed the getting started page, but using .net5 getting same error.

```
tye run frontend -v Debug
Loading Application Details...
Restoring and evaluating projects
[...]

PR #827, opened 2020-11-26
#766 fix missing msbuild error code output

newuser:
MSBuild error wasn't written to console for some reason. Restore task fails and no section is available yet for parsing. Is it acceptable solution?

### DIFF
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -87,6 +87,10 @@ namespace Microsoft.Tye
                     projects: projectServices,
                     configRoot: config.Source.DirectoryName!,
                     output: output);
+                if (msbuildEvaluationResult.ExitCode != 0)
+                {
+                    throw new CommandException($"Unable to parse output.\r\n{msbuildEvaluationResult.StandardOutput}.\r\n Try buildMSBuild error.");
+                }
                 var msbuildEvaluationOutput = msbuildEvaluationResult
                     .StandardOutput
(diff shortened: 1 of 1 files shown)


src/Microsoft.Tye.Core/ApplicationFactory.cs
```diff
@@ -87,6 +87,10 @@ public static async Task<ApplicationBuilder> CreateAsync(OutputContext output, F
                     projects: projectServices,
                     configRoot: config.Source.DirectoryName!,
                     output: output);
+                if (msbuildEvaluationResult.ExitCode != 0)
+                {
+                    throw new CommandException($"Unable to parse output.\r\n{msbuildEvaluationResult.StandardOutput}.\r\n Try buildMSBuild error.");
```
jkotalik, line 92: @JunTaoLuo is this something we explicitly avoid throwing for?

JunTaoLuo:
If there was an issue with restore, we log to the output at the `Debug` level here: https://github.com/dotnet/tye/blob/1b5275a76a548deb56abbaa2aa049ac1a82467f0/src/Microsoft.Tye.Core/ApplicationFactory.cs#L540-L545. The reason is that in some cases, a restore failure doesn't prevent us from continuing with the command so we only log if you set the log level to debug.

JunTaoLuo:
I wonder if we should log that restore errored without all the details at the default level and include the details at Debug instead.

jkotalik:
@JunTaoLuo could you either close or merge this PR?

JunTaoLuo:
We don't want to throw an error here. Closing.

PR #831, opened 2020-11-29
fix #666, enable dotnet watch to refresh on razer file changes

### LINKED ISSUE #666: tye watch run does not monitor cshtml files
### Describe the bug
Watch is not monitoring cshtml files

### To Reproduce
1. Use the tutorial to get to the frontend/backend
2. Run `tye run --watch`
3. Change index.cshtml
EXPECT:
see change
ACTUAL:
no change

Changes occur in the .cs file but not in cshtml.

tye --version: 0.4.0-alpha.20371.1+d7623120d66b85bfeda8ab69eb5ff792df6b7243

hossambarakat:
Updated the DotNetWatch.targets to include the razor files


src/Microsoft.Tye.Hosting/Watch/assets/DotNetWatch.targets
```diff
@@ -26,6 +26,7 @@ Returns: @(Watch)
     </_CollectWatchItemsDependsOn>
     <_CollectWatchItemsDependsOn Condition=" '$(TargetFramework)' != '' ">
       _CoreCollectWatchItems;
+      $(CustomCollectWatchItems);
```
jkotalik, line 29: @pranavkm we should potentially talk about this; we copied the dotnet watch code into tye, and made a few modifications to instead run the executable rather than calling `dotnet run`. Is there any unification we could consider here?

jkotalik (approved):

jkotalik (approved):

jkotalik:
Thanks!

src/Microsoft.Tye.Hosting/Watch/assets/DotNetWatch.targets
pranavkm: VS scans for dotnet-watch assets in the SDK. Is this something we could do with Tye? Alternatively we could always ship a sources packages to make sure we're in sync.

PR #879, opened 2021-01-07
Refactore dockerfile generator

sirh3e:
Applied from SOLID the L part on on ContainerInfo.cs


jkotalik:
I don't see this as an improvement. Closing.

PR #881, opened 2021-01-07
use of linq

sirh3e:


davidfowl:
This isn't an improvement

jkotalik:
Closing as we don't see this as an improvement.

PR #882, opened 2021-01-07
use of linq and language feature

sirh3e:


jkotalik:
I don't think this is an improvement. Closing.

PR #884, opened 2021-01-08
allows specifying the Kubernetes imagePullSecrets

### LINKED ISSUE #508: How to set image imagePullSecrets?
How to set image imagePullSecrets?

thoemmi:
This PR is my attempt to allow specifying the Kubernetes [imagePullSecrets](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret) in the configuration.

The configuration schema is changed from
```yaml
registry: example.azurecr.io
```
to
```yaml
registry:
  name: example.azurecr.io
  pullSecret: acr-secret
```
(`pullSecret` is optional)

The old schema is still supported, so existing configurations won't break.

What do you think about these changes?

This PR would fix #508


jongio:
@thoemmi - Can you please have a look at my requirement here: https://github.com/dotnet/tye/issues/508#issuecomment-739997498

We often assemble solutions from many registries and pull images from more than one.  So we'd need to be able to have imagePullSecrets at the root level and then allow each service to provide their own pull secret.

jkotalik:
Hm, maybe we need to be able to set registries per service, if we want to set image pull secrets per service.

IMO, the scope of this PR is good, we can follow up if we want to specify registry and imagePullSecrets per service.

docs/reference/schema.md
```diff
@@ -93,6 +86,36 @@ Specifies the list of ingresses.
...
+- Determining how to tag the images
+- Determining where to push the images
+
+If this is not specified in configuration, interactive deployments will prompt for it.
+
+### Registry Example
+
+```yaml
+registry:
+  hostname: example.azurecr.io
```
jkotalik, line 103: Maybe `name` is better here instead of `hostname`.

src/Microsoft.Tye.Core/ConfigModel/ConfigRegistry.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.ComponentModel.DataAnnotations;
```
jkotalik, line 1: nit: copyright

src/Microsoft.Tye.Core/Serialization/ConfigRegistryParser.cs
```diff
@@ -0,0 +1,54 @@
+﻿using Microsoft.Tye.ConfigModel;
```
jkotalik, line 1: nit: copyright.

src/tye/Program.DeployCommand.cs
```diff
@@ -118,7 +118,9 @@ internal static void ApplyRegistry(OutputContext output, ApplicationBuilder appl
                 var registry = output.Prompt("Enter the Container Registry (ex: 'example.azurecr.io' for Azure or 'example' for dockerhub)", allowEmpty: !requireRegistry);
                 if (!string.IsNullOrWhiteSpace(registry))
                 {
-                    application.Registry = new ContainerRegistry(registry.Trim());
+                    var registryPullSecrets = output.Prompt("Enter the kubernetes credential secret to authorize pulling an image from the registry or leave empty", allowEmpty: true);
```
jkotalik, line 121: I don't think we should prompt for pull secrets always via interactive. Interactive is only for required items, and pull secrets is more of an optional parameter.

jkotalik (changes requested):
This looks good, pending a few core changes.

jongio:
@jkotalik - That's fine for now - eval per service later.

docs/reference/schema.md
thoemmi: `ConfigRegistry` and `ContainerRegistry` have a `Hostname` property. Shall I rename it to `Name` too?

thoemmi:
Looks to me like flaky tests broke the build 🤔

docs/reference/schema.md
thoemmi: @jkotalik I assume you missed my question 😐

jkotalik: If those are both hostname, I'm fine with keeping this as host name for now then.

thoemmi: Ok, than I guess I have addresses all your change requests.

thoemmi:
Unfortunately, Github does not reset the "change requested" badge, even if all comments have been resolved.
I.e. this PR is ready for review.
/cc @jkotalik

jongio:
@thoemmi - Thanks a lot for doing this!  Could you add more info to docs on how imagePullSecrets work and how someone should use with this new feature?  I'm imagining a sample which points to the imagePullSecrets docs with examples.

jkotalik:
Apologies for the delay in merging. Going to work on that now.

stefankip:
Tried this locally and works fine, just had a small 'huh' when I got the error saying `hostname` is incorrect ;-)

thoemmi:
Oops, after renaming the `registry.hostname``to `registry.name`, I haven't updated the schema documentation. Just pushed the fix.

@jongio Apologies for my delay. Shall I create an example in _./samples/_? And where's the best place I can add some additional documentation?

jongio:
@jkotalik for thoughts on where sample would go

Thanks!

dalestone:
Anyone looking at this?

remcoros:
We need this as well.

One team develops the UI, deploys on private ACR.
back-end devs only need to pull the webapp image from ACR. They use tye

Dona278:
Any updates?

ChristianWeyer:
Is this still 'on' @thoemmi? A really great and important feature! :-)
Thanks.

thoemmi:
@ChristianWeyer AFAIK it's on @jongio or @jkotalik to decide if anything more is required for this PR.

jkotalik:
@philliphoff

Dona278:
After more than one year after the creation of this pullrequest why is still unmerged?
In order to work with private container registry that stores custom images for developers which haven't coding permission I think that this is a must-to-have feature.
Maybe there is a sort of new "by design" confict?
@philliphoff can you provide a feedback please? Thanks

MohamedEbrahimMohsen:
Any updates regarding this pull request? or is there any workaround to add the **_imagePullSecrets_** secret?

davidfowl:
I just realized this is a massive breaking change. I think we'd need to support both formats to take this.

Mrxx99:
The pull request says the old scheme is still supported, so it shouldn't be a breaking change?

davidfowl:
> The old schema is still supported, so existing configurations won't break.

Missed that!

davidfowl (approved):

thoemmi:
Thanks for finally accepting my PR :sweat_smile:

davidfowl:
Just in time for the holidays!

PR #885, opened 2021-01-08
"Advertise" local Tye instances

### LINKED ISSUE #863: Idea: "advertise" local Tye instances

When running Tye locally, the Tye instance should "advertise" itself to the rest of the system as belonging to a "Tye" category of components.  Other components could then identify all running instances of Tye and obtain metadata about each one (e.g. its dashboard endpoint).

A proposal would be to mDNS (a.k.a. ZeroConf or Bonjour), already supported across multiple OS and application platforms.  After starting, each Tye instance would advertise itself as `<instance>._microsoft-tye._tcp` (i.e. a member of the general `_microsoft-tye._tcp` category of service).  Other components could use mDNS to query for all local Tye instances using the `_microsoft-tye._tcp.local` service name, which returns records (and metadata) for all advertised Tye instances.

Another option, likely more complicated, would be to have a central Tye "daemon" that tracks the starting/stopping of Tye instances and then have a CLI command and/or API to query the daemon to obtain its current list.

Developers may start more than one Tye instance at a time (e.g. switching between different applications, samples, etc.) and want to view/manage them as a set within their tools.  Tools that manage Tye instances are not always the same entity that starts/stops those instances; such tools need a way to identify existing running instances as well as correlate them with other context (e.g. workspaces).

philliphoff:
This is a PoC for using mDNS to "advertise" running instances of Tye, useful for tooling to locate and interact with those instances where such tooling may not have been the one to start the instances. Resolves #863.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/tye?pullRequest=885) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/tye?pullRequest=885) before we can accept your contribution.<br/><br/>:x: philliphoff [sign now](https://cla.dotnetfoundation.org/dotnet/tye?pullRequest=885)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/tye?pullRequest=885) it.</sub>

src/Microsoft.Tye.Hosting/Microsoft.Tye.Hosting.csproj
```diff
@@ -21,6 +21,7 @@
   <ItemGroup>
     <PackageReference Include="Bedrock.Framework" Version="0.1.38-alpha.gd25d5b37ad" />
     <PackageReference Include="FeatherHttp" Version="0.1.42-alpha.gf06a8747e7" />
+    <PackageReference Include="Makaretu.Dns.Multicast" Version="0.27.0" />
```
jkotalik, line 24: What's this external dependency? Are we safe to take a dependency on it?

jongio: License is MIT: https://github.com/richardschneider/net-mdns/blob/master/LICENSE

But should go through a review before being committed.

philliphoff: Yes, I wasn't sure what the policies/requirements for this org were regarding 3rd-party dependencies, hence this being more of a PoC to get feedback on the general idea and approach.  This was the only mDNS library I could find for .NET that supported the advertisement side (vs. the observing side), other than an ancient Xamarin implementation that no longer appears to be supported, and also seems dependent (on Windows) on 3rd-party components.

rynowak:
This is pretty dope 👍

jkotalik:
@philliphoff IIRC I think we were fine with taking this as an external dependency. Is there anything else that needs to happen before we merge this PR?

philliphoff:
I'm ok with it as-is, if it's gone though whatever 3rd-party OSS review is required on your side.  (Or is it now technically "our" side?) It'd be great to have, as I already have a corresponding branch of the...other thing that can't quite be mentioned yet...that uses it, and it'd be nice to get that committed prior to its release.

That said, other things have required my attention lately so I've not done any Tye-related development or testing in a couple months.

PR #888, opened 2021-01-12
Fixing some small things in DAPR samples

jkotalik:
Did some verification, found that on unix, dapr can install to /root/.dapr/bin if installed with sudo, so adding that as a backup path to find daprd.

### DIFF
--- a/samples/dapr/pub-sub/orders/orders.csproj
+++ b/samples/dapr/pub-sub/orders/orders.csproj
@@ -5,7 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Dapr.AspNetCore" Version="0.11.0-preview02" />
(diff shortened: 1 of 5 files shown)


rynowak:
BTW you can find the logic that Dapr uses here: https://github.com/dapr/cli/blob/7cf8fade1b9c99b11ea17ce3c8900bfb5a845060/pkg/standalone/common.go#L26 when you do `dapr run`.

I'm not sure that `dapr run` would look in `/root/...` when run normally, it just uses the users home directory. Installing Dapr as root isn't something we're telling people to do 😆

jkotalik:
Yeah I'm not surprised; I only accidentally installed it there when running `sudo dapr init`.

If you think it's a case we shouldn't cover, then I'll remove that logic.

JunTaoLuo:
I agree with @rynowak let's not handle the case where dapr is installed with sudo.

PR #891, opened 2021-01-13
Handle project evaluation target path with spaces

JunTaoLuo:
Fixes https://github.com/dotnet/tye/issues/766. 

The issue is that if tye is installed in a directory path with spaces, such as when user profile contained spaces, the project evaluation command was broken. The fix is to add quotation marks around the path.

It's a bit hard to construct an unit test for this since tye needs to be installed in a special directory but I've verified this manually.

### DIFF
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -516,16 +516,17 @@ namespace Microsoft.Tye
 
             output.WriteDebugLine("Restoring and evaluating projects");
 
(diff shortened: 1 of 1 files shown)


jkotalik:
> It's a bit hard to construct an unit test for this since tye needs to be installed in a special directory but I've verified this manually.

When would this ever happen? Where is tye installed by default?

jkotalik (approved):

JunTaoLuo:
By default the installation is in the user profile. The problem folks were reporting in the issue occur when the user profile contains spaces.

Issue #516, opened 2020-05-30
Tye deploy throws incorrect error saying dotnet not found


NileshGule:
### Describe the bug
I had an error in the path to .csproj file in Tye.yaml file. When I ran the **tye deploy** command, I got an error message saying "Failed to find dotnet. Make sure the .Net SDK is installed and on the PATH". I have the .NET SDK installed and it is set on the PATH correctly. 

When I corrected the path to .csproj file and reran the tye deploy command, it worked fine.

### To Reproduce
Introduce a wrong path in the tye.yaml file for the csproj file

Refer to the attached screenshot for exact error message
<img width="828" alt="tye deploy dotnet not found error" src="https://user-images.githubusercontent.com/1298374/83319465-4e7c6200-a271-11ea-8a17-3ce47269f883.PNG">

### Further technical details

- The platform Windows 10

jkotalik:
Can you get logs at a higher verbosity by running `tye deploy --verbosity debug`? 

My guess is that it's just a bad log message that needs to be improved.

NileshGule:
@jkotalik in the verbosity debug mode as well I get the same output

<img width="771" alt="tye deploy verbose output" src="https://user-images.githubusercontent.com/1298374/83650238-b2dd4f80-a5ea-11ea-89e7-bc6689bd805e.PNG">


Can we make the error handling robust by checking for invalid paths and throwing an error suggesting the File / Path does not exist instead of giving misleading error message regarding .Net SDK?

chen1tian:
use **tye init** command generate a new tye.yaml, then try again.
may be c# project path was wrong.

JunTaoLuo:
The problem here is that we use the [project path](https://github.com/dotnet/tye/blob/master/src/Microsoft.Tye.Core/ProjectReader.cs#L112) when resolving the SDK probably so we would be able take into account of global.json settings. We can add a log saying which project directory we are scanning to make this more obvious.

Alternatively if we don't mind not resolving the correct global.json setting from the point of view of the projects listed, we can just use the current working directory.

PR #892, opened 2021-01-13
Update SDK resolution and error message

JunTaoLuo:
Addresses https://github.com/dotnet/tye/issues/516

Update the error message to be a little clearer where SDK resolution failed. For some added spiciness, I also changed the logic to resolve the SDK from the current working directory instead of the project directory. I think that better suits the principle of least surprise as suggested by the linked issue.

### DIFF
--- a/src/Microsoft.Tye.Core/ProjectReader.cs
+++ b/src/Microsoft.Tye.Core/ProjectReader.cs
@@ -90,27 +90,22 @@ namespace Microsoft.Tye
                 {
                     output?.WriteDebugLine("Locating .NET SDK...");
 
-                    // It says VisualStudio - but we'll just use .NET SDK
-                    var instances = MSBuildLocator.QueryVisualStudioInstances(new VisualStudioInstanceQueryOptions()
-                    {
-                        DiscoveryTypes = DiscoveryType.DotNetSdk,
-
-                        // Using the project as the working directory. We're making the assumption that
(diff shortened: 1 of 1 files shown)


JunTaoLuo:
Actually on second thought I think it's probably best to fall back to global SDK instance if one cannot be resolved from the project instead.

src/Microsoft.Tye.Core/ProjectReader.cs
```diff
@@ -90,27 +90,33 @@ private static void EnsureMSBuildRegistered(OutputContext? output, FileInfo proj
...
-                    var instance = instances.SingleOrDefault();
+                        instance = MSBuildLocator
+                            .QueryVisualStudioInstances()
+                            .FirstOrDefault();
+                    }
+
                     if (instance == null)
                     {
-                        throw new CommandException("Failed to find dotnet. Make sure the .NET SDK is installed and on the PATH.");
+                        throw new CommandException($"Failed to resolve dotnet in {projectFile.Directory} or the PATH. Make sure the .NET SDK is installed and on the PATH.");
```
jkotalik, line 119:
```suggestion
                        throw new CommandException($"Failed to resolve dotnet in {projectFile.Directory} or the PATH. Make sure the .NET SDK is installed and is on the PATH.");
```

jkotalik (approved):

PR #893, opened 2021-01-13
Revert ingress update for 1.18 or lower

jkotalik:
Only have ingress changes for 1.19 or later.


src/shared/KubectlDetector.cs
```diff
@@ -30,22 +30,36 @@ public static Task<bool> IsKubectlConnectedToClusterAsync(OutputContext output)
...
             {
                 // Unfortunately, process throws
-                return false;
+                return null;
             }
         }
 
+        private static int GetVersion(int startIndex, string start, string end, string input)
+        {
+            var first = input.IndexOf(start, startIndex) + start.Length;
```
jkotalik, line 58: This is a hair hacky to parse a json object with string indexs, but I only needed two pieces of info from it.

JunTaoLuo: Wouldn't System.Text.Json work well here?

JunTaoLuo (approved):

src/shared/KubectlDetector.cs
jkotalik: Yeah, I realized I can output the kubectl version as json, making it easy to parse. Before, it wasn't valid json.

Issue #580, opened 2020-07-14
tye run --watch has locking issues with project used multiple times, and replicas


JamesNK:
Example tye.yaml:

```yaml
name: microservice
services:
- name: frontend
  project: frontend/frontend.csproj
- name: backend-1
  project: backend/backend.csproj
- name: backend-2
  project: backend/backend.csproj
```

Starting projects with `tye run --watch` succeeds, but when a file changes in the backend project the build fails with locking issues:

```
[11:09:50 INF] Building projects failed with exit code 1:
  Determining projects to restore...
  All projects are up-to-date for restore.
C:\Program Files\dotnet\sdk\3.1.300\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(424,5): error MSB4018: The "CreateAppHost" task failed unexpectedly. [C:\Development\Temp\microservice\backend\backend.csproj]
C:\Program Files\dotnet\sdk\3.1.300\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(424,5): error MSB4018: System.AggregateException: One or more errors occurred. (800703EE) (The process cannot access the file 'C:\Development\Temp\microservice\backend\obj\Debug\netcoreapp3.1\backend.exe' because it is being used by another process.) [C:\Development\Temp\microservice\backend\backend.csproj]
C:\Program Files\dotnet\sdk\3.1.300\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(424,5): error MSB4018:  ---> Microsoft.NET.HostModel.HResultException: 800703EE [C:\Development\Temp\microservice\backend\backend.csproj]
C:\Program Files\dotnet\sdk\3.1.300\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(424,5): error MSB4018:    at Microsoft.NET.HostModel.ResourceUpdater.ThrowExceptionForLastWin32Error() [C:\Development\Temp\microservice\backend\backend.csproj]
C:\Program Files\dotnet\sdk\3.1.300\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(424,5): error MSB4018:    at Microsoft.NET.HostModel.AppHost.HostWriter.<>c__DisplayClass2_0.<CreateAppHost>g__UpdateResources|1() [C:\Development\Temp\microservice\backend\backend.csproj]
[...]

PR #895, opened 2021-01-14
Ensure watched projects are not built in parallel to avoid file locking issues

JunTaoLuo:
Fixes https://github.com/dotnet/tye/issues/580.

I've experimented with several approaches and this seems the most general. One of the drawbacks is that if there are multiple services that are based on the same project file, re-running build on all of them will now be much slower. However, I'm skeptical of running build once and re-using the results since the files could have changed during the time the project was built the first time. Ideally we would track what file changes have been accounted for and whether builds need to be re-run but I don't think the file watcher has that capability at the momennt.

### DIFF
--- a/src/Microsoft.Tye.Hosting/Model/Application.cs
+++ b/src/Microsoft.Tye.Hosting/Model/Application.cs
@@ -3,11 +3,12 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using System.Threading;
+using System.Threading.Tasks;
 
 namespace Microsoft.Tye.Hosting.Model
 {
@@ -18,18 +19,6 @@ namespace Microsoft.Tye.Hosting.Model
             Source = source.FullName;
             ContextDirectory = source.DirectoryName!;
             Services = services;
-
-            foreach (var s in Services.Values)
-            {
-                if (s.Description.RunInfo is ProjectRunInfo projectRunInfo)
-                {
-                    var projectFileName = projectRunInfo.ProjectFile.FullName;
-                    if (!ProjectFileLocks.ContainsKey(projectFileName))
-                    {
-                        ProjectFileLocks[projectFileName] = new SemaphoreSlim(1, 1);
-                    }
-                }
-            }
         }
 
         public string Source { get; }
@@ -38,7 +27,7 @@ namespace Microsoft.Tye.Hosting.Model
 
         public Dictionary<string, Service> Services { get; }
 
-        internal Dictionary<string, SemaphoreSlim> ProjectFileLocks { get; } = new Dictionary<string, SemaphoreSlim>();
+        internal ConcurrentDictionary<string, TaskCompletionSource<ProcessResult>> ProjectProcesses { get; } = new ConcurrentDictionary<string, TaskCompletionSource<ProcessResult>>();
 
         public Dictionary<object, object> Items { get; } = new Dictionary<object, object>();
 
--- a/src/Microsoft.Tye.Hosting/Model/Application.cs
+++ b/src/Microsoft.Tye.Hosting/Model/Application.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Threading;
 
 namespace Microsoft.Tye.Hosting.Model
 {
@@ -17,6 +18,18 @@ namespace Microsoft.Tye.Hosting.Model
             Source = source.FullName;
             ContextDirectory = source.DirectoryName!;
             Services = services;
+
+            foreach (var s in Services.Values)
+            {
+                if (s.Description.RunInfo is ProjectRunInfo projectRunInfo)
+                {
+                    var projectFileName = projectRunInfo.ProjectFile.FullName;
+                    if (!ProjectFileLocks.ContainsKey(projectFileName))
+                    {
+                        ProjectFileLocks[projectFileName] = new SemaphoreSlim(1, 1);
+                    }
+                }
+            }
         }
 
         public string Source { get; }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Tye.Hosting/Model/Application.cs
```diff
@@ -17,6 +18,18 @@ public Application(FileInfo source, Dictionary<string, Service> services)
...
             Services = services;
+
+            foreach (var s in Services.Values)
+            {
+                if (s.Description.RunInfo is ProjectRunInfo projectRunInfo)
+                {
+                    var projectFileName = projectRunInfo.ProjectFile.FullName;
+                    if (!ProjectFileLocks.ContainsKey(projectFileName))
+                    {
+                        ProjectFileLocks[projectFileName] = new SemaphoreSlim(1, 1);
```
jkotalik, line 29: Why can't these just be locks instead of semaphoreslims?

JunTaoLuo: Builds are called via [`ProcessUtil.RunAsync`](https://github.com/dotnet/tye/pull/895/files#diff-956d150a381630f9458461ccd954c3cf99e0be29aafd92b81a4ab98635f80763R372). You can't lock when making async calls.

src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -359,9 +359,24 @@ async Task RunApplicationAsync(IEnumerable<(int ExternalPort, int Port, string?
...
+                                    var projectSemaphore = application.ProjectFileLocks[projectFile];
+                                    await projectSemaphore.WaitAsync();
+                                    try
+                                    {
+                                        _logger.LogDebug($"[{replica}] Building project {projectFile}:");
+                                        buildResult = await ProcessUtil.RunAsync("dotnet", $"build \"{service.Status.ProjectFilePath}\" /nologo", throwOnError: false, workingDirectory: application.ContextDirectory);
+                                    }
+                                    finally
+                                    {
+                                        _logger.LogDebug($"[{replica}] Finished Building project {projectFile}:");
```
pranavkm, line 376: This should go in the `try` block

src/Microsoft.Tye.Hosting/Model/Application.cs
```diff
@@ -25,6 +27,8 @@ public Application(FileInfo source, Dictionary<string, Service> services)
 
         public Dictionary<string, Service> Services { get; }
 
+        internal ConcurrentDictionary<string, TaskCompletionSource<ProcessResult>> ProjectProcesses { get; } = new ConcurrentDictionary<string, TaskCompletionSource<ProcessResult>>();
```
davidfowl, line 30: What is this?

JunTaoLuo: In case there exists multiple services that are based on the same project file, we want to trigger the build only once, instead of building the same project multiple times. I guess I can probably name this better so it's more obvious what this is.

src/Microsoft.Tye.Hosting/ProcessRunner.cs
```diff
@@ -359,9 +359,27 @@ async Task RunApplicationAsync(IEnumerable<(int ExternalPort, int Port, string?
...
-                                if (service.Description.RunInfo is ProjectRunInfo)
+                                if (service.Description.RunInfo is ProjectRunInfo projectRunInfo)
                                 {
+                                    var projectFile = projectRunInfo.ProjectFile.FullName;
+                                    var newProcess = new TaskCompletionSource<ProcessResult>();
+                                    var ongoingProcess = application.OngoingBuildProjectProcesses.GetOrAdd(projectFile, newProcess);
+
+                                    if (ongoingProcess != newProcess)
+                                    {
+                                        return (await ongoingProcess.Task).ExitCode;
```
jkotalik, line 370: Add a debug log here

rynowak:
> I've experimented with several approaches and this seems the most general.

Did you consider the approach of synthesizing an MSBuild project w/ p2ps and calling build on that? This is something I discussed with @rainersigwald in the past, and it lets MSBuild *just do its thing*. This way you get to leverage parallelism in the build engine without creating a new build engine 😆 I know the idea of generating a project seems like a hack, but IIRC it's a technique that's used in a few places already.

jkotalik (approved):
Pending verification that I mentioned off of github.

jkotalik:
> Did you consider the approach of synthesizing an MSBuild project w/ p2ps and calling build on that? This is something I discussed with @rainersigwald in the past, and it lets MSBuild just do its thing. This way you get to leverage parallelism in the build engine without creating a new build engine 😆 I know the idea of generating a project seems like a hack, but IIRC it's a technique that's used in a few places already.

Yeah isn't that what we already do for the initial build? Could we leverage that here?

rynowak:
> Yeah isn't that what we already do for the initial build? Could we leverage that here?

I'm not sure if y'all ever built that idea. I don't think I got past the idea stage with it, but the expert said its the best way to avoid MSBuild stepping on itself while keeping parallelism.

rainersigwald:
How are y'all building? Shelling out to `dotnet build` or using the MSBuild API? If the latter, could you do all your build submissions in the same beginbuild/endbuild pair? Then MSBuild should reuse build results so even if there's concurrent requests to build the same project we'll deduplicate them just like we would in a single build with a synthesized project that points to everything.

(Normally I'd dig in a bit deeper but I have ~1 hour of work time left before going on parental leave so not today!)

rynowak:
> (Normally I'd dig in a bit deeper but I have ~1 hour of work time left before going on parental leave so not today!)

And it sounds like not next week either. Congrats!

JunTaoLuo:
@rynowak @jkotalik just to be clear, we evaluate all services (i.e. restore and read project metadata) using a generated meta-project file, as described above. However, we build and run each service individually, as seen [here](https://github.com/dotnet/tye/blob/c50d480e3730dfcad301ef2705b0a3c478a0028e/src/Microsoft.Tye.Hosting/ProcessRunner.cs#L148). 

Conceptually, it's not quite that simple either. Since we run each service in a different process, they each have a different file watcher. When some of them change, we'll need each process to report back to a centralized place where we can then determine which service(s) need to be rebuilt by tracing through all their dependencies and then generate a meta-project file to rebuild. This might be worth exploring in the future, but I'm not sure it's worth redesigning in this PR.

JunTaoLuo:
It's unlikely I'll make progress on this PR.

onionhammer:
Has this made it into a release yet? Seems like if I tye run --watch and change a project referenced by two services they often clash and 1 then fails to restart (and the other 'wins')

Issue #898, opened 2021-01-15
Idea: add a "control plane" for running Tye applications


philliphoff:
### What should we add or change to make your life better?

Add a "control plane" of sorts to Tye applications to allow management while running, in particular shutdown, but also restart, increase/decrease service replica counts, etc. 

### Why is this important to you?

Tooling may need to manipulate instances of Tye applications that it didn't start (and therefore does not have access to its PID in order to kill, for example) but, also, would prefer a method of manipulation more direct and consistent cross-platform than, say, sending a SIGKILL signal to a process (on Linux).

> This could also be accomplished using the Tye "daemon" notion that's been discussed earlier.

jongio:
Maybe look at something like Octant for this: http://octant.dev/

philliphoff:
@jongio Sorry, I should have been more specific; this is more about Tye applications running locally than those deployed to Kubernetes.

jongio:
I gotcha.  My comment was about seeing if we can mod octant to work with Tye versus  building from scratch.

PR #899, opened 2021-01-15
PoC Add "control plane" method to shutdown Tye application

philliphoff:
Adds a dashboard "control plane" endpoint that allows shutting down the Tye application.

Related to #898.

### DIFF
--- a/src/Microsoft.Tye.Hosting/TyeDashboardApi.cs
+++ b/src/Microsoft.Tye.Hosting/TyeDashboardApi.cs
@@ -12,6 +12,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 using Microsoft.Tye.Hosting.Model;
 using Microsoft.Tye.Hosting.Model.V1;
 
@@ -36,6 +37,7 @@ namespace Microsoft.Tye.Hosting
         public void MapRoutes(IEndpointRouteBuilder endpoints)
         {
             endpoints.MapGet("/api/v1", ServiceIndex);
+            endpoints.MapPut("/api/v1/control/shutdown", ControlPlaneShutdown);
             endpoints.MapGet("/api/v1/services", Services);
             endpoints.MapGet("/api/v1/services/{name}", Service);
             endpoints.MapGet("/api/v1/logs/{name}", Logs);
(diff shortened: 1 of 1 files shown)


src/Microsoft.Tye.Hosting/TyeDashboardApi.cs
```diff
@@ -36,6 +37,7 @@ public TyeDashboardApi()
         public void MapRoutes(IEndpointRouteBuilder endpoints)
         {
             endpoints.MapGet("/api/v1", ServiceIndex);
+            endpoints.MapDelete("/api/v1/control", ControlPlaneShutdown);
```
JunTaoLuo, line 40: Is this the best way to communicate the shutdown intent? A few consideration here: 
1. What about the restart and modification of replica count mentioned in the issue? I wouldn't think that would make sense with a Delete verb
2. Should restart and replica count be separate endpoint? If so then I think this should not be named control but something like /api/v1/shutdown.

philliphoff: @JunTaoLuo  I'm not highly opinionated about the shape of a control plane, or whether it should be integrated with the data plane or exist as a completely separate path.  Originally I had a `/api/v1/control/shutdown` endpoint, but then couldn't decide whether `POST` or `PUT` (or both) made more sense.  Using `DELETE` at the root of the control plane endpoint seemed to be the simplest way to indicate "shut the whole thing down".

I agree that wouldn't make sense for controlling, say, the number of replicas of a service.  I imagine that would be a `PUT` on, say, a service-specific endpoint (e.g. `/api/v1/services/{name}` or `/api/v1/control/services/{name}`) passing a JSON body that represented what you wanted the new state of that service to be (e.g. with a different replica count).

JunTaoLuo: Hmm now I think about it, wouldn't it make more sense to map a different verb to the services endpoint? I think it a DELETE to `/api/v1/services` to signify shutting down all tye services and a DELETE to `/api/v1/services/{name}` to signify shutting down a particular tye service makes sense. Thoughts?

As far as replica modification, I think we can discuss that further when the PR covers that functionality is made (either this one or separately), but I agree PUT probably makes sense.

philliphoff: I *could* see that, though I think that leads to a semantic that a Tye application is only "running" if it has services running.  For example, if a user individually `DELETE`s every service in an application, does the Tye application (i.e. host) then completely shutdown (i.e. equivalent to a `DELETE` at the service root)?  Does it ever make sense to still have a "running" Tye application even if none of its services are running?  For example, to still allow grabbing logs, metrics, restarting, etc.?

I think it might be good to have a (`DELETE`) endpoint that isn't semantically tied to the services themselves, that more represents the shutdown of the host of those services.  That could even be as simple as using `/api/v1`.

jkotalik (approved):
To unblock work.

src/Microsoft.Tye.Hosting/TyeDashboardApi.cs
```diff
@@ -56,6 +59,15 @@ private Task ServiceIndex(HttpContext context)
             _options);
         }
 
+        private Task ControlPlaneShutdown(HttpContext context)
+        {
+            var lifetime = context.RequestServices.GetRequiredService<IHostApplicationLifetime>();
+
+            lifetime.StopApplication();
+
+            return context.Response.CompleteAsync();
```
davidfowl, line 68: This isn't necessary.

PR #902, opened 2021-01-19
Add X-Forwarded-* headers for request proxying in ingress nodes

FiniteReality:
Previously, Tye's ingress proxy wouldn't add X-Forwarded-* headers to its proxied requests, making anything which depends on the Host header for the request (e.g. OAuth redirect URIs, CORS origins, etc.) almost impossible. This PR adds those headers to proxied requests, so that reverse proxying more or less "just works" using more or less standard ASP.NET forwarded headers.

For example:
```cs
// ConfigureServices
services.Configure<ForwardedHeadersOptions>(options =>
    {
        options.ForwardedHeaders = ForwardedHeaders.All;
        options.KnownProxies.Add(IPAddress.Loopback);
        options.KnownProxies.Add(IPAddress.IPv6Loopback);
        options.RequireHeaderSymmetry = true;
    });

// Configure
app.UseForwardedHeaders();
```

I have not yet added these to proxied websocket requests, because I'm not 100% sure how proxying works for those :sweat_smile:

### DIFF
--- a/src/Microsoft.Tye.Hosting/Infrastructure/ProxyExtensions.cs
+++ b/src/Microsoft.Tye.Hosting/Infrastructure/ProxyExtensions.cs
@@ -204,14 +204,7 @@ namespace Microsoft.AspNetCore.Proxy
             }
 
             // Append Via header so clients are aware of reverse-proxying.
-            if (response.Headers.TryGetValue("Via", out var via))
-            {
-                via.Append($"{responseMessage.Version} Tye");
-            }
-            else
-            {
-                response.Headers["Via"] = new[] { $"{responseMessage.Version} Tye" };
-            }
+            response.Headers.AppendCommaSeparatedValues("Via", $"{responseMessage.Version} Tye");
 
             // SendAsync removes chunking from the response. This removes the header so it doesn't expect a chunked response.
             response.Headers.Remove("transfer-encoding");
--- a/src/Microsoft.Tye.Hosting/Infrastructure/ProxyExtensions.cs
+++ b/src/Microsoft.Tye.Hosting/Infrastructure/ProxyExtensions.cs
@@ -85,6 +85,11 @@ namespace Microsoft.AspNetCore.Proxy
                 }
             }
 
+            // Append request forwarding headers
+            requestMessage.Headers.TryAddWithoutValidation("X-Forwarded-For", context.Connection.LocalIpAddress.ToString());
+            requestMessage.Headers.TryAddWithoutValidation("X-Forwarded-Proto", request.Scheme);
+            requestMessage.Headers.TryAddWithoutValidation("X-Forwarded-Host", request.Host.ToUriComponent());
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.Tye.Hosting/Infrastructure/ProxyExtensions.cs
```diff
@@ -85,6 +85,11 @@ public static HttpRequestMessage CreateProxyHttpRequest(this HttpContext context
                 }
             }
 
+            // Append request forwarding headers
+            requestMessage.Headers.TryAddWithoutValidation("X-Forwarded-For", context.Connection.LocalIpAddress.ToString());
```
jkotalik, line 89: @Tratcher does YARP do anything by default for adding the x-forwarded headers?

Also, websockets. What would we need to add to make websockets work?

Tratcher: Yes, YARP adds x-forwarded headers by default.
https://microsoft.github.io/reverse-proxy/articles/transforms.html#defaults

x-forwarded headers should be applied equally to websocket requests.

Careful, I don't think TryAddWithoutValidation actually appends to existing values. Test it out.

jkotalik: We always want to append here, correct?

@FiniteReality could you validate what happens when TryAddWithoutValidation is called on an already existing value? And also look into how these could apply to websockets.

FiniteReality: Looking at [the source](https://source.dot.net/#System.Net.Http/System/Net/Http/Headers/HttpHeaders.cs,e7ab97274f238802), TryAddWithoutValidation appears to append to existing values if one is present.

I think applying these to WebSockets should be relatively straightfoward too - I'll look into it later today.

FiniteReality:
Turns out I misread the spec while implementing the [Via header](https://tools.ietf.org/html/rfc7230#section-5.7.1):
```
   A proxy MUST send an appropriate Via header field, as described
   below, in each message that it forwards.  An HTTP-to-HTTP gateway
   MUST send an appropriate Via header field in each inbound request
   message and MAY send a Via header field in forwarded response
   messages.
```
So, a `Via` header is not necessary on the response, but is required for requests forwarded by the proxy. Skimming through the YARP code, it doesn't seem to follow the standard here either. (Unless I'm missing something...)

jkotalik (approved):

jkotalik:
Thanks!

Tratcher:
@FiniteReality fair point, nobody's asked for Via support in YARP. While the header is informative, it doesn't serve much functional use, especially in a reverse-proxy scenario where the apps are more likely to be aware of their network topography.

mcrio:
Hi, thanks for the PR. It's very useful. Please can you help me understand when this change will be part of the `0.7.0-*` CI build?

PR #904, opened 2021-01-20
Fix #901 - Executes git clone, based on the tye.yaml directory.

### LINKED ISSUE #901: Relative `cloneDirectory` creates multiple sub-directories
### Describe the bug
Specifying a relative `cloneDirectory` when configuring a service with a `repository`, the folder structure is created such that the repository is cloned into a directory located at `.\cloneDirectory\cloneDirectory\service-name`

### To Reproduce
Create the following `tye.yaml` file and run it with `tye build`
```yml
name: ExampleStack
services:
  - name: example-service
    cloneDirectory: Services
    repository: https://github.com/dotnet/tye # just an example repository
```

**Expected output:**
The TYE repository is cloned into a folder named `Services`, relative to the directory where `tye build` was executed from.  
i.e. `.\Services\example-service`

**Actual output:**
The TYE repository is cloned into a `Services` folder that is nested inside another `Services` folder, relative to the directory where `tye build` was executed from.
i.e. `.\Services\Services\example-service`

### Further technical details

I believe this is due to the interaction of the following lines and relative paths.
- Using [`git clone`](https://github.com/dotnet/tye/blob/659baa7d731ff3200d8c4f2628a236fc52cf7d5e/src/Microsoft.Tye.Core/ApplicationFactory.cs#L276) to clone into [`clonePath`](https://github.com/dotnet/tye/blob/659baa7d731ff3200d8c4f2628a236fc52cf7d5e/src/Microsoft.Tye.Core/ApplicationFactory.cs#L276) with a working directory set to [`path` ](https://github.com/dotnet/tye/blob/659baa7d731ff3200d8c4f2628a236fc52cf7d5e/src/Microsoft.Tye.Core/ApplicationFactory.cs#L270)

So effectively what happens above is that `path` is set to `Services` and `clonePath` is set to `Services\example-service`.
`git clone` will then set it's working directory to already be inside the `Services` folder, and then clone the repository into `Services/example-service`, ending up with the `./Services/Services/example-service` structure as highlighted above.

I believe ensuring that a full path is always used for the `path` variable should resolve this, as this is what the default case is doing when no `cloneDirectory` is provided.

Something like:
```cs
var path = string.IsNullOrWhiteSpace(configService.CloneDirectory) ?
      Path.Join(rootConfig.Source.DirectoryName, ".tye", "deps") :
      Path.GetFullPath(configService.CloneDirectory);
```

samtrion:
With this fix, all variants should have been considered.

It was tested on a Windows system.

- Service without `cloneDirectory` property
- Service with absolute `cloneDirectory`property
- Service with relative `cloneDirectory`property

### DIFF
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -267,7 +267,7 @@ namespace Microsoft.Tye
                     else if (!string.IsNullOrEmpty(configService.Repository))
                     {
                         // clone to .tye folder
-                        var path = configService.CloneDirectory ?? Path.Join(rootConfig.Source.DirectoryName, ".tye", "deps");
+                        var path = configService.CloneDirectory ?? Path.Join(".tye", "deps");
                         if (!Directory.Exists(path))
                         {
                             Directory.CreateDirectory(path);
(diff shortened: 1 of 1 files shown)


jkotalik (approved):

jkotalik:
Thanks!

Issue #839, opened 2020-12-03
binding protocol is always tcp even if i  set udp 


rmznkm:
### Describe the bug

When i wanna port forward for 6831/upd it forward 6831/tcp 
 - name: jaeger
    image: jaegertracing/all-in-one:latest
    bindings:
      - name: "udp" 
        port: "6831"
        protocol: Udp
      - name: "tcp"
        port: "16686"

### To Reproduce
 - name: jaeger
    image: jaegertracing/all-in-one:latest
    bindings:
      - name: "udp" 
        port: "6831"
        protocol: Udp
      - name: "tcp"
        port: "16686"

### Further technical details
4af6aeaafed5        jaegertracing/all-in-one:latest            "/go/bin/all-in-one-…"   2 minutes ago       Up 2 minutes        5775/udp, 5778/tcp, 14250/tcp, **0.0.0.0:6831->6831/tcp**, 6831-6832/udp, 14268/tcp, 0.0.0.0:16686->16686/tcp     jaeger_3ea1d146-3

 tye --version
0.5.0-alpha.20555.1+fae47325b0c8d7dafcdec5d1248191b24b2adc23

- If possible rerun the command with `-v debug` and include the output
windows docker dekstop

jooooel:
👍 Reported the same thing here https://github.com/dotnet/tye/issues/658

Issue #658, opened 2020-09-14
Add support for publishing UDP ports


jooooel:
### What should we add or change to make your life better?
Possibility to publish UDP ports like you can in Docker `1234:1234/udp`.

### Why is this important to you?
I'm trying to run a project in Tye that uses OpenTelemetry and their Jaeger exporter. The Jaeger container requires you to publish UDP ports so that the Jaeger client can send data. This is not currently supported by Tye (AFAIK).

When trying something like this in Tye:

```
- name: jaeger
    image: jaegertracing/all-in-one:1.19
    bindings:
      - name: thrift_compact
        port: 6831/udp
      - name: thrift_binary
        port: 6832/udp
      - name: ui
        port: 16686
```

You get the following error:
`Error parsing tye.yaml: (31, 15): "port" value must be an integer.`

It seems that `port` must be an integer, and looking in `DockerRunner.cs` there seems to be no way of dealing with this. I would be happy to take a shot a providing a PR if you are interested?

jkotalik:
Right now, we support protocol definition via setting `protocol`. 
```yaml
- name: jaeger
    image: jaegertracing/all-in-one:1.19
    bindings:
      - name: thrift_compact
        port: 6831
        protocol: UDP
      - name: thrift_binary
        port: 6832
        protocol: UDP
      - name: ui
        port: 16686
```
That configuration would need to be understood by the Jaeger exporter though. You should be able to use [the following convention](https://github.com/dotnet/tye/blob/master/src/Microsoft.Tye.Extensions.Configuration/TyeConfigurationExtensions.cs#L17) to read the value from config and use it accordingly when configuring Jaeger

jooooel:
Thanks for getting back to me @jkotalik. I'm really no "docker port expert" 😬 but I don't think using `protocol` does what I need?

In the image below the two first outputs are when running the container through Tye. The third and fourth outputs are when running the container via the following command below the image (taken from the Jaeger getting started):

![image](https://user-images.githubusercontent.com/9303280/93513086-9a890a80-f925-11ea-9fb8-ef5f745bd253.png)

```
 docker run -d --name jaeger \
  -p 6831:6831/udp \
  -p 6832:6832/udp \
  -p 16686:16686 \
  jaegertracing/all-in-one:1.19
```

So to me it looks like running it via Tye doesn't publish the ports correctly. Looking at the command Tye produces (I removed a lot of stuff for clarity):

```
[...]

PR #915, opened 2021-01-23
Bind port in docker with udp flag when protocol is udp

ikkentim:
This change allows setting the protocol of a port of an image service to udp, which will bind the port with the udp flag in docker.

Notes: 
- I didn't know if I should handle it case-insensitive, it currently only works when the protocol is set to lowercase udp
- I couldn't find any unit tests or end-to-end tests in this area, so I didn't add any either
- Possibly it is preferred to add this to the port field instead of the protocol field (e.g. `port: 6831/udp`). if it is, I can update this PR accordingly

I needed this when I tried to use jeager:
```
name: jaeger-sample
services:
- name: jaeger
  image: jaegertracing/all-in-one:1.21
  bindings:
  - name: compact
    port: 6831
    protocol: udp
  - name: frontend
    port: 16686
```

This PR solves https://github.com/dotnet/tye/issues/658 and https://github.com/dotnet/tye/issues/839 .

### DIFF
--- a/src/Microsoft.Tye.Hosting/DockerRunner.cs
+++ b/src/Microsoft.Tye.Hosting/DockerRunner.cs
@@ -240,7 +240,7 @@ namespace Microsoft.Tye.Hosting
                     // These are the ports that the application should use for binding
 
                     // 1. Tell the docker container what port to bind to
-                    portString = docker.Private ? "" : string.Join(" ", ports.Select(p => $"-p {p.Port}:{p.ContainerPort ?? p.Port}"));
+                    portString = docker.Private ? "" : string.Join(" ", ports.Select(p => $"-p {p.Port}:{p.ContainerPort ?? p.Port}{(p.Protocol == "udp" ? "/udp" : string.Empty)}"));
 
                     if (docker.IsAspNet)
                     {


src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -240,7 +240,7 @@ async Task RunDockerContainer(IEnumerable<(int ExternalPort, int Port, int? Cont
                     // These are the ports that the application should use for binding
 
                     // 1. Tell the docker container what port to bind to
-                    portString = docker.Private ? "" : string.Join(" ", ports.Select(p => $"-p {p.Port}:{p.ContainerPort ?? p.Port}"));
+                    portString = docker.Private ? "" : string.Join(" ", ports.Select(p => $"-p {p.Port}:{p.ContainerPort ?? p.Port}{(p.Protocol == "udp" ? "/udp" : string.Empty)}"));
```
jkotalik, line 243:
```suggestion
                    portString = docker.Private ? "" : string.Join(" ", ports.Select(p => $"-p {p.Port}:{p.ContainerPort ?? p.Port}{(string.Equals(p.Protocol, "udp", StringComparison.OrdinalIgnoreCase) ? "/udp" : string.Empty)}"));
```

jkotalik (approved):

davidfowl:
Does it need to work for non docker services as well?

ikkentim:
I haven't looked into the kubernetes side of this change.

For project services I suppose this feature could be added as well (it would probably involve changing the proxy to support proxying udp), but I don't think there would be much demand for such a feature?

davidfowl:
> For project services I suppose this feature could be added as well (it would probably involve changing the proxy to support proxying udp), but I don't think there would be much demand for such a feature?

If there's udp support for docker then there should be udp support for projects as well. If it isn't supported for projects, then it needs to fail with a nice error message.

ikkentim:
@davidfowl I've thrown a CommandException when proxies are attempted to be configured with udp.

src/Microsoft.Tye.Hosting/DockerRunner.cs
```diff
@@ -79,6 +79,11 @@ public async Task StartAsync(Application application)
                         continue;
                     }
 
+                    if (string.Equals(binding.Protocol, "udp", StringComparison.InvariantCultureIgnoreCase))
```
jkotalik, line 82: I'm pretty sure this will always throw by default when specifying UDP, not just for proxies. This check is checking the bindings of a service.

I think the main things to verify are:
1. If a single container is running, works with UDP
2. If a single project is running, works with UDP
3. If multiple containers are running, works with UDP
4. If multiple containers and projects are running (docker proxy enabled), works with UDP
5. If an ingress is enabled, works with UDP.

ikkentim: The piece of code i added the check and exception in appear to be responsible for adding proxies for access from inside docker to services running outside of docker.

This will only throw for any service without DockerRunInfo and without IngressRunInfo, so I think only services with project or executable type ( ? ) for which it will set up a proxy in docker, so it won't throw when you bind a udp port for a container here.

I'll try and verify the scenarios you mention above.

davidfowl: Did you?

ikkentim: Apologies for the delay...

**Scenario 1**

> If a single container is running, works with UDP

docker container runs and port is bound @ `0.0.0.0:6831->6831/udp`

```yaml
name: scenario-1
services:
- name: jaeger-service
  image: jaegertracing/all-in-one:1.21
  bindings:
  - name: compact
    port: 6831
    protocol: udp
```

```
$ docker ps
CONTAINER ID   IMAGE                           COMMAND                  CREATED          STATUS          PORTS                                                                                   NAMES
1182ed35355f   jaegertracing/all-in-one:1.21   "/go/bin/all-in-one-…"   26 seconds ago   Up 23 seconds   5775/udp, 5778/tcp, 14250/tcp, 14268/tcp, 6832/udp, 16686/tcp, 0.0.0.0:6831->6831/udp   jaeger-service_68510150-b
```

**Scenario 2**

> If a single project is running, works with UDP

Project starts fine. But I get an error in my web app "Unrecognized scheme in server address 'udp://localhost:5050'. Only 'http://' is supported.", this can be fixed with PR #916.


```yaml
name: scenario-2
services:
- name: webapplication1
  project: WebApplication1/WebApplication1.csproj
  bindings:
  - name: udp
    port: 5050
    protocol: udp
```

**Scenario 3**

> If multiple containers are running, works with UDP

I'm assuming here that by "multiple containers", you mean multiple replicas of a container. This will not work because it would need to proxy the incoming udp packages to the correct service, and no udp proxy has been implemented.

```yaml
name: scenario-3
services:
- name: jaeger-service
  image: jaegertracing/all-in-one:1.21
  replicas: 2
  bindings:
  - name: compact
    port: 6831
    protocol: udp
```

```
$ tye run
Loading Application Details...
Launching Tye Host...

[21:38:54 INF] Executing application from C:\Users\Tim\source\repos\WebApplication1\tye.yaml
[21:38:54 INF] Dashboard running on http://127.0.0.1:8000
[21:38:54 INF] Mapping external port 6831 to internal port(s) 59186, 59187 for jaeger-service binding compact
Drats! 'run' failed:
        Proxy does not support the udp protocol yet.
```

If by "multiple containers" you mean different containers, this works just fine:

```yaml
name: scenario-3b
services:
- name: jaeger-service
  image: jaegertracing/all-in-one:1.21
  bindings:
  - name: compact
    port: 6831
    protocol: udp
- name: jaeger-service-2
  image: jaegertracing/all-in-one:1.21
  bindings:
  - name: compact
    port: 6832
    protocol: udp
```

```
$ docker ps
CONTAINER ID   IMAGE                           COMMAND                  CREATED          STATUS          PORTS                                                                                   NAMES
f66fbb9e6cec   jaegertracing/all-in-one:1.21   "/go/bin/all-in-one-…"   16 seconds ago   Up 14 seconds   5775/udp, 5778/tcp, 14250/tcp, 14268/tcp, 6831/udp, 16686/tcp, 0.0.0.0:6832->6832/udp   jaeger-service-2_6c6da33c-9
4c26836fb8dc   jaegertracing/all-in-one:1.21   "/go/bin/all-in-one-…"   16 seconds ago   Up 14 seconds   5775/udp, 5778/tcp, 14250/tcp, 14268/tcp, 6832/udp, 16686/tcp, 0.0.0.0:6831->6831/udp   jaeger-service_5235c5b2-8
```

**Scenario 4**

> If multiple containers and projects are running (docker proxy enabled), works with UDP

This won't work since the proxy doesn't support udp yet.

```yaml
name: scenario-4
services:
- name: webapplication1
  project: WebApplication1/WebApplication1.csproj
  replicas: 2
  bindings:
  - name: udp
    port: 5050
    protocol: udp
```

```
$ tye run
Loading Application Details...
Launching Tye Host...

[21:49:02 INF] Executing application from C:\Users\Tim\source\repos\WebApplication1\tye.yaml
[21:49:03 INF] Dashboard running on http://127.0.0.1:8000
[21:49:03 INF] Mapping external port 5050 to internal port(s) 59817, 59819 for webapplication1 binding udp
Drats! 'run' failed:
        Proxy does not support the udp protocol yet.
```

**Scenario 5**

> If an ingress is enabled, works with UDP.

Ingress works fine in the scenario below. The udp connections are not touched by ingres.

```yaml
name: scenario-5
services:
- name: webapplication1
  project: WebApplication1/WebApplication1.csproj
  args: --urls http://*:5000 # workaround for #916
  bindings:
  - name: http
    port: 5000
  - name: udp
    port: 5050
    protocol: udp
- name: webapplication2
  project: WebApplication2/WebApplication2.csproj
  args: --urls http://*:5001 # workaround for #916
  bindings:
  - name: http
    port: 5001
  - name: udp
    port: 5051
    protocol: udp
ingress:
- name: http
  bindings:
  - port: 8080
    protocol: http
  rules:
  - service: webapplication1
    path: /app1
  - service: webapplication2
    path: /app2
```

But when I add a docker container it won't work, because it tries to add a proxy container.

A comment in tye explains the reason for the proxy:
```cs
// Inject a proxy per non-container service. This allows the container to use normal host names within the
// container network to talk to services on the host
```

```yaml
name: scenario-5
services:
- name: jaeger-service
  image: jaegertracing/all-in-one:1.21
  bindings:
  - name: compact
    port: 6831
    protocol: udp
- name: webapplication1
  project: WebApplication1/WebApplication1.csproj
  args: --urls http://*:5000 # workaround for #916
  bindings:
  - name: http
    port: 5000
  - name: udp
    port: 5050
    protocol: udp
- name: webapplication2
  project: WebApplication2/WebApplication2.csproj
  args: --urls http://*:5001 # workaround for #916
  bindings:
  - name: http
    port: 5001
  - name: udp
    port: 5051
    protocol: udp
ingress:
- name: http
  bindings:
  - port: 8080
    protocol: http
  rules:
  - service: webapplication1
    path: /app1
  - service: webapplication2
    path: /app2
```

```
$ tye run
Loading Application Details...
Launching Tye Host...

[22:08:01 INF] Executing application from C:\Users\Tim\source\repos\WebApplication1\tye.yaml
[22:08:01 INF] Dashboard running on http://127.0.0.1:8000
[22:08:01 INF] Processing ingress rule: Path:/app1, Host:null, Service:webapplication1
[22:08:01 INF] Service webapplication1 is using (5000, http://localhost:5000/)
[22:08:01 INF] Processing ingress rule: Path:/app2, Host:null, Service:webapplication2
[22:08:01 INF] Service webapplication2 is using (5001, http://localhost:5001/)
[22:08:02 INF] Docker image jaegertracing/all-in-one:1.21 already installed
Drats! 'run' failed:
        Proxy does not support the udp protocol yet.
```

If the projects do not use udp (but the docker container still does) it works just fine because now there is no need to proxy udp
```yaml
name: scenario-5
services:
- name: jaeger-service
  image: jaegertracing/all-in-one:1.21
  bindings:
  - name: compact
    port: 6831
    protocol: udp
- name: webapplication1
  project: WebApplication1/WebApplication1.csproj
  bindings:
  - name: http
    port: 5000
- name: webapplication2
  project: WebApplication2/WebApplication2.csproj
  bindings:
  - name: http
    port: 5001
ingress:
- name: http
  bindings:
  - port: 8080
    protocol: http
  rules:
  - service: webapplication1
    path: /app1
  - service: webapplication2
    path: /app2
```

```
$ docker ps
CONTAINER ID   IMAGE                                   COMMAND                  CREATED          STATUS          PORTS                                                                                   NAMES
bbfd5f6e30c1   mcr.microsoft.com/dotnet/core/sdk:3.1   "dotnet Microsoft.Ty…"   19 seconds ago   Up 16 seconds                                                                                           webapplication2-proxy_f831c996-1
284767eceada   mcr.microsoft.com/dotnet/core/sdk:3.1   "dotnet Microsoft.Ty…"   19 seconds ago   Up 16 seconds                                                                                           webapplication1-proxy_f1f02015-8
43ee389c7bad   jaegertracing/all-in-one:1.21           "/go/bin/all-in-one-…"   19 seconds ago   Up 17 seconds   5775/udp, 5778/tcp, 14250/tcp, 14268/tcp, 6832/udp, 16686/tcp, 0.0.0.0:6831->6831/udp   jaeger-service_de537efe-4
```

---

It might be better to just skip udp bindings here instead of throwing an error, this allows single-replica project services with udp bindings to work alongside docker container services.

ikkentim: Please let me know if I should remove this error and skip the binding instead.

ikkentim:
Could this be reviewed? I would really appreciate if this issue could be fixed in an upcoming update, because it is a blocking issue for the projects I'm working on (along with #916 )

davidfowl:
It would be good if you could summarize the support for the scenarios that @jkotalik  mentioned:


> If a single container is running, works with UDP
If a single project is running, works with UDP
If multiple containers are running, works with UDP
If multiple containers and projects are running (docker proxy enabled), works with UDP
If an ingress is enabled, works with UDP.

ikkentim:
I've tested those scenarios in the a previous comment, to summarize:

What does work:
- A single container runs fine, with UDP
- A single project with UDP starts fine. But if it's a asp.net core app, you'll get an error `Unrecognized scheme in server address 'udp://localhost:5050'. Only 'http://' is supported.`, this can be fixed with PR #916.
- Multiple containers with UDP from different services (1 replica per service) works fine.
- Ingress works fine if you have just projects with udp+tcp/http

What doesn't work:
- Multiple (replica) containers with UDP won't work because that would need an implemented UDP proxy. You'll receive the error added in this PR `Proxy does not support the udp protocol yet.`
- I'm not entirely sure what's meant by "If multiple containers and projects are running (docker proxy enabled), works with UDP", I'm assuming you mean a proxy which has been added because of a replica count > 1. But UDP through a proxy won't work, you'll get the new error message added in this PR
- When you have projects with udp+tcp/http and you also add containers, this does not work because tye adds a proxy to allow the containers to talk to the projects by a host name.

---

All I really want (and most people probably want who run into this issue) is to be able to use a docker container which exposes a UDP port in tye (like jaegertracing).

ikkentim:
If it helps, we could limit UDP bindings to containers only, which would reduce the list of "what doesn't work"....

davidfowl:
I'll merge as is, can you file an issue labeled UDP support with the things you call out in your comment as bullet points?

Issue #776, opened 2020-11-11
Exception when stopping tye with CTRL+C


jongio:
### Describe the bug
Getting this:
[07:09:22 ERR] Error while shutting down
System.AggregateException: One or more hosted services failed to stop. (Cannot access a disposed object.
Object name: 'Bedrock.Framework.TimerAwaitable'.)
 ---> System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'Bedrock.Framework.TimerAwaitable'.
   at Bedrock.Framework.TimerAwaitable.Stop()
   at Bedrock.Framework.Server.StopAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.Internal.Host.StopAsync(CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at Microsoft.Extensions.Hosting.Internal.Host.StopAsync(CancellationToken cancellationToken)
   at Microsoft.Tye.Hosting.ProxyService.StopAsync(Application application) in /_/src/Microsoft.Tye.Hosting/ProxyService.cs:line 190
   at Microsoft.Tye.Hosting.AggregateApplicationProcessor.StopAsync(Application application) in /_/src/Microsoft.Tye.Hosting/AggregateApplicationProcessor.cs:line 39
   at Microsoft.Tye.Hosting.AggregateApplicationProcessor.StopAsync(Application application) in /_/src/Microsoft.Tye.Hosting/AggregateApplicationProcessor.cs:line 49
   at Microsoft.Tye.Hosting.TyeHost.StopAsync() in /_/src/Microsoft.Tye.Hosting/TyeHost.cs:line 319
Oh dear! Operation canceled.
### To Reproduce

tye run

let is start

CTRL+C

### Further technical details

- Include the output of `tye --version`
0.6.0-alpha.20559.2+c91e24cf224a8a74121614052885956092c5384f
- If possible rerun the command with `-v debug` and include the output
- The platform (Linux/macOS/Windows)

WSL2

.NET 5 

azsdke2e
azsdke2e2

davidhenley:
Same

```
[11:21:53 ERR] Error while shutting down
System.AggregateException: One or more hosted services failed to stop. (Cannot access a disposed object.
Object name: 'Bedrock.Framework.TimerAwaitable'.)
 ---> System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'Bedrock.Framework.TimerAwaitable'.
[...]

PR #983, opened 2021-03-16
Catch ODE from bedrock framework code

jkotalik:
Fixes https://github.com/dotnet/tye/issues/776

As we plan to remove bedrock.framework long term in favor of YARP, just going with the bandaid solution of catching the ODE from bedrock framework.

### DIFF
--- a/src/Microsoft.Tye.Hosting/ProxyService.cs
+++ b/src/Microsoft.Tye.Hosting/ProxyService.cs
@@ -185,7 +185,15 @@ namespace Microsoft.Tye.Hosting
 
             if (_host != null)
             {
-                await _host.StopAsync();
+                try
+                {
+                    await _host.StopAsync();
+                }
+                catch (ObjectDisposedException)
+                {
+                    // System.ObjectDisposedException: Cannot access a disposed object.
+                    // From Bedrock.Framework. As we plan on removing this long term, not going to directly fix in bedrock.
+                }
 
                 if (_host is IAsyncDisposable disposable)
                 {


davidfowl:
YARP doesn't replace bedrock 😄

davidfowl:
Can you reproduce this?

jkotalik:
> Can you reproduce this?

Infrequently, but I can occasionally.

> YARP doesn't replace bedrock 😄

Why wouldn't it? Don't we want the same behavior locally as we do on kubernetes? I don't think we need L4 capabilities...

jkotalik:
I investigated this for around a day in the past and wasn't able to get a repro with a locally built version of the package (release mode, with logs, etc). I think that's an appropriate time box for it.

davidfowl:
> Why wouldn't it? Don't we want the same behavior locally as we do on kubernetes? I don't think we need L4 capabilities...

We use L4 for replicas, which is what k8s does.

jkotalik:
>We use L4 for replicas, which is what k8s does.

Ah, you are talking about the case where there isn't an ingress. This code is still used I guess then.

Do you think long term we should keep this dependency on Bedrock.Framework?

davidfowl:
> Ah, you are talking about the case where there isn't an ingress. This code is still used I guess then.

No I mean when there are multiple replicas of a service, the address you get is proxy address. That's not an HTTP proxy and has nothing to do with the ingress.

jkotalik:
Yeah, I was agreeing with you.

jkotalik:
offline approval

Issue #428, opened 2020-04-30
Nginx sample is not working


mpcmarcos:
I'm testing with Mac the nginx sample and it's not working. The nginx image is stopped few seconds later to start.

Log:
2020/04/30 18:31:17 [emerg] 1#1: host not found in upstream "appA" in /etc/nginx/conf.d/default.conf:5
nginx: [emerg] host not found in upstream "appA" in /etc/nginx/conf.d/default.conf:5

rynowak:
I can reproduce this except the nginx image doesn't crash on me.

rynowak:
I was able to work around this by adding a resolver to `nginx.conf` this seems like a behavior difference between docker on windows and docker on macOS.

```
server {
    listen 80;

    location /A {
        resolver 127.0.0.11;
        proxy_pass http://appA/;
        proxy_http_version 1.1;
        proxy_set_header   Upgrade $http_upgrade;
        proxy_set_header   Connection keep-alive;
        proxy_set_header   Host $host;
        proxy_cache_bypass $http_upgrade;
        proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header   X-Forwarded-Proto $scheme;
    }

    location /B {
        resolver 127.0.0.11;
        proxy_pass http://appB/;
        proxy_http_version 1.1;
        proxy_set_header   Upgrade $http_upgrade;
        proxy_set_header   Connection keep-alive;
        proxy_set_header   Host $host;
        proxy_cache_bypass $http_upgrade;
        proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header   X-Forwarded-Proto $scheme;
    }
}
```

rynowak:
@davidfowl

PR #1014, opened 2021-04-19
Support podman

tmds:
This makes tye work with rootless podman.

It requires podman 3.1+.

Users need to update their `$HOME/.config/containers/containers.conf` to allow containers to access services on the host:
```
[engine]

network_cmd_options=["allow_host_loopback=true"]
```

All tests pass except `NginxIngressTest` which fails for me due to https://github.com/dotnet/tye/issues/428.

Fixes https://github.com/dotnet/tye/issues/563

@davidfowl @jkotalik ptal


tmds:
I ran these tests in a clean Fedora 33 VM.

I did the following configuration:
```sh
mkdir -p $HOME/.config/containers
cat >$HOME/.config/containers/containers.conf <<EOF
[engine]

network_cmd_options=["allow_host_loopback=true"]
EOF

sudo dnf -y update podman
```

Then ran:
```sh
./build.sh --test
```

All tests passed. For E2E:

```
 Assemblies Run
Microsoft.Tye.E2ETest.dll
Summary
Tests run: 81 — Skipped: 1, Run time: 1213.693s, Finished: 04/22/2021 09:44:38

Skipped tests
Skipped ❢  E2ETest.TyeRunTests.FrontendBackendAzureFunctionTest
```

src/Microsoft.Tye.Core/DockerDetector.cs
```diff
@@ -14,41 +18,107 @@ public class DockerDetector
...
+            AspNetUrlsHost = "localhost";
             try
             {
-                var result = await ProcessUtil.RunAsync("docker", "version", throwOnError: false, cancellationToken: new CancellationTokenSource(Timeout).Token);
-                return result.ExitCode == 0;
+                ProcessResult result;
+                try
+                {
+                    // try to use podman.
+                    result = ProcessUtil.RunAsync("podman", "version -f \"{{ .Client.Version }}\"", throwOnError: false, cancellationToken: new CancellationTokenSource(Timeout).Token).Result;
```
jkotalik, line 43: It's been a while, but is it possible for someone to have both podman and docker locally? If so, having a way to select which one should be used?

tmds: Yes, both can be installed side-by-side. In most cases, only one will be installed.
Do you want to make it configurable in `tye.yaml`, or in some other way?

tmds:
@jkotalik can you review the PR? Functionally I think it's fine, but you may want to suggest some refactorings. I hope this can be part of the next tye release.

src/Microsoft.Tye.Core/DockerDetector.cs
jkotalik: I think it needs to be configurable.

jkotalik:
LGTM besides an option to turn on and off.

tmds:
@jkotalik I've added an option to `tye.yaml` that allows to select `podman`/`docker`. ptal at the changes.

src/schema/tye-schema.json
```diff
@@ -13,6 +13,11 @@
             "description": "Dockerhub username or hostname of remote registry. Used for tagging images.",
             "type": "string"
         },
+        "engine": {
```
jkotalik, line 16: I think containerEngine is a better name here.

jkotalik (approved):
Pending a rename.

tmds:
I've renamed the property.

I also relaxed the podman version check because I don't want risk it making things unusable. Instead of failing when we can't parse the version, we assume the version requirement is met.

jkotalik:
Thanks!

tmds:
@jkotalik thank you too. Do you have a rough idea when you'll do the next release?

jkotalik:
Hoping for one relatively soon 👍 . There are a few key bugs we need to fix first (.NET 6 support).

PR #1020, opened 2021-04-28
use of method / function chaining

sirh3e:
refactored src/tye/Program.cs


jkotalik:
This doesn't make any functional change.

PR #1021, opened 2021-04-28
refactored HandleException use variable for context is less error prone

sirh3e:
refactored src/tye/Program.cs


jkotalik:
This doesn't make any functional change.

PR #1022, opened 2021-04-28
refactored removed nested if use of if pattern

sirh3e:
refactored src/tye/InitHost.cs


jkotalik:
This doesn't make any functional changes.

PR #1027, opened 2021-05-04
Don't consider it a fatal error when the container host is not accessible.

tmds:
For many use-cases there is no need to access the host.


tmds:
@jkotalik ptal

davidfowl (approved):

PR #1029, opened 2021-05-04
Fixing internal builds

jkotalik:


eng/Signing.props
```diff
@@ -33,6 +33,7 @@
 
   <ItemGroup>
     <!-- We don't produce font files. We rebundle some for using the web brower, so they do not need to be signed. -->
```
wtgodbe, line 35: Might be worth expanding this comment to include .js files

wtgodbe (approved):

eng/Signing.props
jkotalik: To confirm, this file is shipping as a part of tye right now. Is that okay?

wtgodbe: I don't know of a reason why it wouldn't be

JunTaoLuo (approved):

PR #1048, opened 2021-05-18
Specify ingress IP in Debug

Hylaean:
This allows the possibility to expose an ingress to multiple IP while running with tye run.
It might interest anybody running a service locally for the benefit of an emulator or device.
Current version only allows to expose localhost.

Ingresses bindings have a new option: ip. Setting an ip will allow binding to the ip. '*' will allow binding to all ips.
Syntax is the same as Kestrel as it is passed to Kestrel after parsing checks.

Note: unit test don't run in the windows pipeline because of firewall issues, they do run in linux and on windows, in Debug.

### DIFF
--- a/docs/reference/schema.md
+++ b/docs/reference/schema.md
@@ -514,6 +514,11 @@ The port of the binding.
 
 The protocol (`http` or `https`).
 
+#### `ip` (`string`)
+
+The optional IP adress to bind to. Can be '*' for all addresses.
+Default is localhost.
+
 ## IngressRule
 
 `IngressRule` elements appear in an array within the `rules` property of the `Ingress` element. Rules configure the routing behavior of the ingress proxy.
--- a/src/Microsoft.Tye.Core/CoreStrings.resx
+++ b/src/Microsoft.Tye.Core/CoreStrings.resx
@@ -153,6 +153,9 @@
   <data name="MustBeAnInteger" xml:space="preserve">
     <value>"{value}" value must be an integer.</value>
   </data>
+  <data name="MustBeAnIPAddress" xml:space="preserve">
+    <value>"{value}" value must be an IP address, * or localhost.</value>
+  </data>
   <data name="MustBePositive" xml:space="preserve">
     <value>"{value}" value cannot be negative.</value>
   </data>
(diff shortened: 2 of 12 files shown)


src/Microsoft.Tye.Core/CoreStrings.resx
```diff
@@ -153,6 +153,9 @@
   <data name="MustBeAnInteger" xml:space="preserve">
     <value>"{value}" value must be an integer.</value>
   </data>
+  <data name="MustBeAnIPAddress" xml:space="preserve">
+    <value>"{value}" value must be an IP address, * or localhost.</value>
```
philliphoff, line 157: Nit: add quotes around `*` and `localhost`?

src/Microsoft.Tye.Core/Serialization/ConfigIngressParser.cs
```diff
@@ -144,6 +146,17 @@ private static void HandleIngressBindingMapping(YamlMappingNode yamlMappingNode,
 
                         binding.Port = port;
                         break;
+                    case "ip":
+                        if (YamlParser.GetScalarValue(key, child.Value) is string ipString
+                            && (IPAddress.TryParse(ipString, out var ip) || ipString == "*" || ipString.Equals("localhost", StringComparison.OrdinalIgnoreCase) ))
+                        {
+                            binding.IPAddress = ip == IPAddress.Loopback || ip == IPAddress.IPv6Loopback ? null : ipString;
```
philliphoff, line 153: Why ignore the user's explicit setting if it happens to be the loopback address?

docs/reference/schema.md
```diff
@@ -514,6 +514,11 @@ The port of the binding.
 
 The protocol (`http` or `https`).
 
+#### `ip` (`string`)
```
philliphoff, line 517: Would it be possible to add this same definition to the [`tye.yaml` YAML schema](https://github.com/dotnet/tye/blob/main/src/schema/tye-schema.json)?

philliphoff:
@Hylaean Overall, looks good.  Just a couple of comments/questions and then a merge conflict that needs to be resolved.

src/Microsoft.Tye.Core/CoreStrings.resx
Hylaean: Done

philliphoff (approved):

Issue #947, opened 2021-02-17
Is there are way to set service dependencies 


jkears:
Within Docker compose we can set service dependencies to control start-up order.  
Is that something that is also available in Tye.yaml?

davidfowl:
Nope we discussed it but why do you need it?

gterdem:
@davidfowl I think it is a common scenario where an application service is depend on to an infrastructure service (redis, rabbitmq, mongodb, sqlserver etc) so the infrastructure should be guaranteed to run before app services. 

Since there is no way to start/stop/restart individual services (oh there is already a PR  https://github.com/dotnet/tye/pull/948 about it happens to be waiting your review 😜) it will be hard to master the chaos.

jkears:
@davidfowl I have a start-up scenario in which service A sends across registration-type data to secondary service.  I currently have this handled via a strong-consistency operation using intra-service communication.  It is possible to instead follow an eventual consistency pattern through event notification however my original thought was that if any registration operation failed, we would fail the service which would see it recycled and thus inherently retry as many times as would be necessary to complete the work.   The registration data in this case is Business Rule Definitions per each domain command.  The secondary service provides a centralized query point for authoring business rules for any command and any service.   The actual business rules are stored within each service and are added/updated or deleted on the secondary service.  

At a very high-level ...

![image](https://user-images.githubusercontent.com/1290937/109213408-8a9c0f00-777e-11eb-8d81-f2cfc7c547f1.png)

davidfowl:
The reason we didn't do dependencies and startup order is because we didn't think it would be replicated in the production environments so giving people that false sense that it works locally is bad.

jkears:
I suppose that somewhat makes sense.  
Either way, we have many ways to get around this so personally I am not all that fused.

More importantly, we are way more productive having Tye and Dapr than without, and so for that, many thanks!  

Tye & Dapr Rock!

ThomasSkyldahl:
Sorry for bumping the issue, but another great benefit for dependencies would be to enable starting single or multiple services and their dependencies
something like
`tye run --services catalogservice,orderservice`

this would bring up the two services and their dependencies

NMillard:
@davidfowl I guess that makes sense to not give the wrong impression.

I'm thinking that dependencies allow one to emulate what's in production since we'd expect a few services to already exist such as a database, rabbitmq, etc.

Not having the opportunity to do dependencies makes the application restart until the service is available, which is a bit annoying.

PR #1061, opened 2021-05-26
DotNetWatcher: don't wait for file changes to restart when the process exits

tmds:
When not using `--watch` an app will restart immediately when it terminates.
This makes `--watch` also restart apps that terminate.

cc @jkotalik

### DIFF
--- a/src/Microsoft.Tye.Hosting/Watch/DotNetWatcher.cs
+++ b/src/Microsoft.Tye.Hosting/Watch/DotNetWatcher.cs
@@ -69,29 +69,24 @@ namespace Microsoft.DotNet.Watcher
 
                     await Task.WhenAll(processTask, fileSetTask);
 
-                    if (processTask.Result.ExitCode != 0 && finishedTask == processTask && !cancellationToken.IsCancellationRequested)
-                    {
-                        // Only show this error message if the process exited non-zero due to a normal process exit.
-                        // Don't show this if dotnet-watch killed the inner process due to file change or CTRL+C by the user
-                        _logger.LogError($"watch: Exited with error code {processTask.Result}");
-                    }
-                    else
-                    {
-                        _logger.LogInformation("watch: {Replica} Exited", replica);
-                    }
-
                     if (finishedTask == cancelledTaskSource.Task || cancellationToken.IsCancellationRequested)
                     {
                         return;
                     }
-
-                    if (finishedTask == processTask)
+                    else if (finishedTask == processTask)
                     {
-                        // Now wait for a file to change before restarting process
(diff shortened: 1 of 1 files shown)


davidfowl:
It has to wait or you risk getting into a quick restart cycle when the app is in a broken state.

tmds:
The app migrates a database at startup. It's expected to fail until the database container is running.
Without `--watch` the app restarts until the database was migrated. With `--watch` it fails on the first attempt and then waits until a file changes.
I think this difference is not expected.

The containers are also configured to continue restarting.

davidfowl:
That difference by design.

davidfowl:
Sorry is this just about containers?

tmds:
It's about .NET applications.
`tye` restarts them when the exit. `tye --watch` does not. It's an unexpected difference.

containers are always restarted.

When the .NET apps fails to start because missing dependent services are not yet up, with `tye` it restarts until they are, with `tye --watch` it sits and waits for the user to make a move.

davidfowl:
That's by design. The .NET applications wait for a change until they restart to avoid the restart loop. That's how watch tools work. As an example, if I change the port and there's a binding conflict, I want to wait for the next change to potentially fix that before I re-run again.

tmds:
In this case, it fails to start because a dependent service is not yet up.

The not restarting is unexpected.

Maybe it can restart with a backoff timer?

davidfowl:
Maybe its time to bite the bullet and add dependencies.

cc @glennc

tmds:
> Maybe its time to bite the bullet and add dependencies.

Is this going to happen?

fwiw, I agree with what you said in https://github.com/dotnet/tye/issues/947#issuecomment-786190875:

> The reason we didn't do dependencies and startup order is because we didn't think it would be replicated in the production environments so giving people that false sense that it works locally is bad.

aladd04:
Dug a bit to find this but I'm glad we found something out there about this behavior. My team is running into this issue, intermittently, too. Are there any updates for this feature?

For localdev we have a redis instance that needs to be up and ready before our APIs can bootstrap. `tye run` works perfect every time, restarting the API until it's ready. But `tye run --watch` will wait for a file change if it initially fails because redis isn't ready.

Our current workaround is to have two different tye yamls, one for `tye run` and one for `tye run --watch`.  The `tye run --watch` variant marks redis as an external dependency instead of an image. Downside is we have to go outside of tye and manually start that redis container first. Then manually shut it down if we want to run the other command.

tmds:
> Dug a bit to find this but I'm glad we found something out there about this behavior. My team is running into this issue, intermittently, too. Are there any updates for this feature?

This PR fixes the issue by also restarting the app when using `--watch`.

It has not been merged because https://github.com/dotnet/tye/issues/947 offers a different solution. I don't think it is actively being worked on.

davidfowl:
Closing this

Issue #1035, opened 2021-05-06
Tye CLI not accessible in Terminal on M1 macOS


robertsundstrom:
I install the latest CI-build of Tye on an M1 Mac, but when entering the command it cannot be found.

When running ```./tye```from the ```~/.dotnet/tools```directory it works.

It seem it has been added to the ```$PATH``` It seems like ~ does not expand. (https://github.com/dotnet/sdk/issues/9415)

I have reinstalled the tools but no luck.

````
/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/share/dotnet:~/.dotnet/tools:/Library/Apple/usr/bin:/Library/Frameworks/Mono.framework/Versions/Current/Commands
````

```
robert@Roberts-MBP-2 tools % dotnet --list-sdks
3.0.103 [/usr/local/share/dotnet/sdk]
3.1.408 [/usr/local/share/dotnet/sdk]
5.0.202 [/usr/local/share/dotnet/sdk]
```

```
robert@Roberts-MBP-2 tools % ls -la
total 520
drwxr-xr-x   5 robert  staff     160 May  6 09:12 .
drwxr-xr-x  19 robert  staff     608 May  5 15:06 ..
drwxr-xr-x   5 robert  staff     160 May  6 09:12 .store
-rwxr--r--   1 robert  staff  165976 Feb 11 18:44 dotnet-format
-rwxr-xr-x   1 robert  staff   95244 May  6 09:12 tye
```

**Comment by robertsundstrom** [comment]
My workaround it to put this in ```~/.zshrc```: 

```
export PATH=$HOME/.dotnet/tools:$PATH
```

**Comment by jkotalik** [comment]
@philliphoff did you see this when you were using tye on your M1?

**Comment by philliphoff** [comment]
To be honest, I don't recall.  That said, I *do* have this in my `.zprofile`:

```bash
# Add Tye to PATH.
export PATH="$PATH:/Users/<user>/.dotnet/tools"
```

I have to imagine that I added it myself, given the other items/comments in the file.  Anyway, I know I've had trouble in the past with using `~/` as part of paths in `PATH`.  Looking at my `PATH`, I see both `~/.dotnet/tools` as well as the path I added via my `.zprofile`.  I'm not sure where `~/.dotnet/tools` came from; I don't see another shell configuration file with that anywhere.

philliphoff:
[...]

PR #1076, opened 2021-06-04
Add PATH note for Mac.

philliphoff:
Related to #1035.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -2,10 +2,10 @@
 
 Tye is a tool that makes developing, testing, and deploying microservices and distributed applications easier. Project Tye includes a local orchestrator to make developing microservices easier and the ability to deploy microservices to Kubernetes with minimal configuration.
 
-## Installing tye
+## Installing Tye
 
 1. Install [.NET Core 3.1](<http://dot.net>).
-2. Install tye via the following command:
+1. Install tye via the following command:
 
     ```text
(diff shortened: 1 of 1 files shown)


docs/getting_started.md
```diff
@@ -2,10 +2,10 @@
 
 Tye is a tool that makes developing, testing, and deploying microservices and distributed applications easier. Project Tye includes a local orchestrator to make developing microservices easier and the ability to deploy microservices to Kubernetes with minimal configuration.
 
-## Installing tye
+## Installing Tye
 
 1. Install [.NET Core 3.1](<http://dot.net>).
-2. Install tye via the following command:
+1. Install tye via the following command:
```
pratiksanglikar, line 8: I think it is still the second step, so "2." makes sense!

pratiksanglikar: git diff is going crazy showing this diff.
I checked the rich diff and it looks fine!

pratiksanglikar (approved):

docs/getting_started.md
pratiksanglikar: image

philliphoff: Markdown lets you use `1.` for every step and dynamically generates the correct number when shown, which makes it less likely that you'll end up with a mis-numbered list after future edits.  I try to switch to that format whenever I happen to see it.

pratiksanglikar: That's really smart! I didn't know that!

PR #1079, opened 2021-06-10
Add endpoint for Tye application metadata

### LINKED ISSUE #806: Idea: dashboard API that exposes application-level metadata
It would be helpful if Tye would expose general metadata about the application, such as the `name` property set in the Tye YAML (or inferred if started directly from a directory or project file).  Also, maybe the location of the YAML file used to run the application (if any).  Currently, a metadata is exposed about each individual service, but not for the application as a whole.  Such metadata could be used for correlating and differentiating Tye application instances across automation and other tooling.

philliphoff:
Adds a `GET` `api/v1/control` endpoint that returns metadata related to the Tye application as a whole (vs. service-level metadata available via other endpoints). In particular, it returns an ID unique to the instance of Tye, the configured/inferred application name, and the source of the configuration (e.g. `tye.yaml` or a project file).

This metadata is helpful for tooling such as [`vscode-tye`](https://github.com/microsoft/vscode-tye), for example, to correlate applications back to a given workspace for debugging purposes.

Resolves #806

### DIFF
--- a/src/Microsoft.Tye.Hosting/Model/Application.cs
+++ b/src/Microsoft.Tye.Hosting/Model/Application.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Tye.Hosting.Model
             ContainerEngine = containerEngine;
         }
 
-        public string ID { get; } = Guid.NewGuid().ToString();
+        public string Id { get; } = Guid.NewGuid().ToString();
 
         public string Name { get; }
 
--- a/src/Microsoft.Tye.Hosting/Model/Application.cs
+++ b/src/Microsoft.Tye.Hosting/Model/Application.cs
@@ -12,13 +12,18 @@ namespace Microsoft.Tye.Hosting.Model
 {
     public class Application
     {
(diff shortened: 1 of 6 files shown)


src/Microsoft.Tye.Hosting/Model/Application.cs
```diff
@@ -12,14 +12,19 @@ namespace Microsoft.Tye.Hosting.Model
...
+        public Application(string name, FileInfo source, Dictionary<string, Service> services, ContainerEngine containerEngine)
         {
+            Name = name;
             Source = source.FullName;
             ContextDirectory = source.DirectoryName!;
             Services = services;
             ContainerEngine = containerEngine;
         }
 
+        public string ID { get; } = Guid.NewGuid().ToString();
```
davidfowl, line 24: nit: Uppercase ID? Or Id?

philliphoff: Looks like the pattern within Tye is `Id`.  Updated.

pratiksanglikar (approved):
Looks good!

PR #1095, opened 2021-06-21
Add dockerfile build args documentation

danspark:
I had to dig a little bit to find out if it was possible to pass build arguments on a `tye.yaml` file. Turns out it is possible, it was made on #521, but I didn't find any docs explaining how. I just added a new section to the `schema.md`, let me know if there's anything I can add to make it more clear.

### DIFF
--- a/src/schema/tye-schema.json
+++ b/src/schema/tye-schema.json
@@ -383,6 +383,13 @@
                     "description": "The Dockerfile to use.",
                     "type": "string"
                 },
+                "dockerFileArgs": {
+                    "description": "Build arguments to use when building the image.",
+                    "type": "array",
+                    "items": {
+                      "type": "object"
+                    }
+                },
                 "dockerFileContext": {
                     "description": "The Dockerfile context to run docker build on.",
                     "type": "string"
--- a/docs/reference/schema.md
+++ b/docs/reference/schema.md
@@ -162,6 +162,10 @@ Including `image` marks the service as a *container*:
 
 The Dockerfile to build from. Uses the name of the service as the name of the image. Only supported for `tye run` currently.
 
+#### `dockerFileArgs` (`string[]`)
+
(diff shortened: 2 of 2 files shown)


philliphoff:
@danspark Thanks for the contribution.  Would you mind also adding the property to the [JSON schema](https://github.com/dotnet/tye/blob/4ed8c5bd5b84b78fb342011b6f63eb3bfcfd2b74/src/schema/tye-schema.json#L368)?

danspark:
> @danspark Thanks for the contribution. Would you mind also adding the property to the [JSON schema](https://github.com/dotnet/tye/blob/4ed8c5bd5b84b78fb342011b6f63eb3bfcfd2b74/src/schema/tye-schema.json#L368)?

Done!

image

src/schema/tye-schema.json
```diff
@@ -383,6 +383,13 @@
                     "description": "The Dockerfile to use.",
                     "type": "string"
                 },
+                "dockerFileArgs": {
+                    "description": "Build arguments to use when building the image.",
+                    "type": "array",
+                    "items": {
+                      "type": "object"
```
philliphoff, line 390: Should the `items` type be `string`?

philliphoff: Or, rather, whether there's a better way to model a string --> string dictionary in JSON schema.

danspark: Looks like that would give a syntax error, from what I saw it really is an object

image

https://github.com/dotnet/tye/blob/5096dc54c3dc9daf71f67b2ebd919f726ca3aaf6/test/E2ETest/testassets/projects/single-phase-dockerfile-multi-args/tye.yaml#L7-L14

philliphoff: Yes, it looks like we accept whatever the string version of the YAML scalar node, so it's essentially a string --> object.  In that case, I think that change fine.

philliphoff (approved):

Issue #1108, opened 2021-07-02
Service metrics in json format


rafaelldi:
Hello!
Endpoint `/api/v1/metrics/{{service}}/` produces response like this:
```
Microsoft.AspNetCore.Hosting/failed-requests{instance="api_70f4d32f-5"} 0
System.Runtime/gen-1-gc-count{instance="api_70f4d32f-5"} 0
Microsoft.AspNetCore.Hosting/current-requests{instance="api_70f4d32f-5"} 0
Microsoft.AspNetCore.Hosting/total-requests{instance="api_70f4d32f-5"} 0
Microsoft.AspNetCore.Hosting/requests-per-second{instance="api_70f4d32f-5"} 0
```
I found this format difficult to parse. Would you mind making it more json-way, please?

rafaelldi:
@philliphoff @pratiksanglikar 
Could you please review my question? I can open a pull request for this change. I think it would also be handy for vs code extension if you plan to add a metrics dashboard to it.

philliphoff:
@rafaelldi I had to take a little time to dig into the code and determine why it is using this particular format and it turned out to be Kubernetes own metrics logging format.  I imagine it was done this way to enable integration with existing Kubernetes metrics monitoring tools.

That said, I don't see any reason why Tye couldn't support both, if JSON was a simpler format to consume.  In fact, there's a bug in the `/metrics/<service>` endpoint where it claims to return JSON even though it clearly is not.  (The `/metrics` endpoint doesn't return any `content-type`, though it arguably should.)

Here's what I suggest, if you're willing to submit a PR:

 - Either by default or when text is specifically requested, for both `/metrics` and `/metrics/<service>` endpoints, return the current Kubernetes text format with the appropriate text `content-type` header in the response.
 - When JSON is specifically requested, for both endpoints, return a JSON format with the appropriate JSON `content-type` header in the response.

As for the JSON format, perhaps something along these lines?

```json
[
  {
    "name": "Microsoft.AspNetCore.Hosting/failed-requests",
    "metadata": {
      "instance": "api_70f4d32f-5"
    },
    "value": 0
  }
]
```

rafaelldi:
Thank you for your time. I'll try to do it as you've advised soon.

davidfowl:
It's currently the Prometheus format I believe

rafaelldi:
[...]

PR #1130, opened 2021-07-23
Add json format for metrics

rafaelldi:
This PR resolves what we've discussed in issue #1108. What do you think?

### DIFF
--- a/src/Microsoft.Tye.Hosting/Model/V1/V1Metric.cs
+++ b/src/Microsoft.Tye.Hosting/Model/V1/V1Metric.cs
@@ -1,0 +1,9 @@
+﻿namespace Microsoft.Tye.Hosting.Model.V1
+{
+    public class V1Metric
+    {
+        public string? Name { get; set; }
+        public string? Value { get; set; }
+        public V1MetricMetadata? Metadata { get; set; }
+    }
+}
--- a/src/Microsoft.Tye.Hosting/Model/V1/V1Metric.cs
+++ b/src/Microsoft.Tye.Hosting/Model/V1/V1Metric.cs
@@ -1,4 +1,8 @@
-﻿namespace Microsoft.Tye.Hosting.Model.V1
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Tye.Hosting.Model.V1
 {
     public class V1Metric
     {
--- a/src/Microsoft.Tye.Hosting/TyeDashboardApi.cs
+++ b/src/Microsoft.Tye.Hosting/TyeDashboardApi.cs
@@ -380,8 +380,8 @@ namespace Microsoft.Tye.Hosting
                 {
                     var metric = new V1Metric
                     {
-                        Name = key, 
-                        Value = value, 
+                        Name = key,
+                        Value = value,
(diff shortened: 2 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/tye?pullRequest=1130) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/tye?pullRequest=1130) before we can accept your contribution.<br/><br/>:x: rafaelldi [sign now](https://cla.dotnetfoundation.org/dotnet/tye?pullRequest=1130)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/tye?pullRequest=1130) it.</sub>

rafaelldi:
Hello! @philliphoff, could you take a look at my request if you have time?

src/Microsoft.Tye.Hosting/Model/V1/V1Metric.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.Tye.Hosting.Model.V1
```
philliphoff, line 1: Nit: add the standard file license header to each new file.  (It looks like there are a couple of existing files that don't have it, which we should fix at some point, but is outside the scope of this PR.)

src/Microsoft.Tye.Hosting/TyeDashboardApi.cs
```diff
@@ -262,6 +264,52 @@ private Task AllMetrics(HttpContext context)
...
+                return JsonSerializer.SerializeAsync(context.Response.Body, new
+                {
+                    message = $"Unknown service {name}"
+                },
+                _options);
+            }
+
+            var acceptHeader = context.Request.Headers[HeaderNames.Accept];
+
+            if (acceptHeader == MediaTypeNames.Application.Json)
```
philliphoff, line 299: Requests can declare multiple acceptable response formats; should we support those cases (i.e. tease out whether the preferred response is JSON and, if so, use that, else fallback to text (assuming that's even in the acceptable list))?

src/Microsoft.Tye.Hosting/Model/V1/V1Metric.cs
rafaelldi: Fixed

src/Microsoft.Tye.Hosting/TyeDashboardApi.cs
rafaelldi: It seems to me that the Prometheus format for metrics is more expected. The json format may be needed in rare cases, for example, as I now find it more convenient to parse a json string than to work with plain text in my plugin. Most users, in my opinion, get metrics through specialized services that know how to work with the Prometheus format. So I think it's more correct to give json only if explicitly asked for it, and in all other cases, respond in Prometheus format.
What do you think?

philliphoff: I'll buy that. :-)

philliphoff (approved):
Approved pending making the formatter and build happy.

rafaelldi:
@philliphoff I'm not sure what I should do next. I've fixed formatting, but one workflow doesn't start automatically. Could you help me, please?

PR #1132, opened 2021-07-26
Add timeout for `Process.WaitForExit()`

### LINKED ISSUE #1128: Tye cannot shutdown with executable service
When an "executable" service is part of the application, Tye can't shutdown.  This occurs at least within a VS Code Dev Container, with a Node.js based service, but perhaps occurs elsewhere.  This happens both when using Ctrl+C from the terminal as well as when using Tye's shutdown command on its control plane.  Tye shuts down everything else, but seems to be stuck waiting for that service to stop.

pratiksanglikar:
Because of the bug [dotnet/runtime#29232](https://github.com/dotnet/runtime/issues/29232), `process.WaitForExit()`
hangs for processes that spawn another long-running processes.
Since these are expected to be long-running processes and we're typically not concerned with capturing all of its output
i.e. it's probably ok for some output to be lost on shutdown since Tye is shutting down anyway, we added the timeout of 60 seconds to the `WaitForExit()` method so that we can shutdown the long-running processes.

Fixes #1128

### DIFF
--- a/src/Microsoft.Tye.Core/ProcessUtil.cs
+++ b/src/Microsoft.Tye.Core/ProcessUtil.cs
@@ -23,6 +23,8 @@ namespace Microsoft.Tye
 
         private static readonly bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
 
+        private const int ProcessExitTimeoutMs = 60 * 1000; // 1 minute timeout for the process to exit.
+
         public static Task<int> ExecuteAsync(
             string command,
             string args,
@@ -118,7 +120,12 @@ namespace Microsoft.Tye
                 {
                     // Even though the Exited event has been raised, WaitForExit() must still be called to ensure the output buffers
                     // have been flushed before the process is considered completely done.
-                    process.WaitForExit();
+                    // Because of the bug in the dotnet runtime https://github.com/dotnet/runtime/issues/29232, Process.WaitForExit()
+                    // hangs for processes that spawn another long-running processes.
+                    // Since these are expected to be long running processes and we're typically not concerned with capturing all of its output
+                    // i.e. it's probably ok for some output to be lost on shutdown, since Tye is shutting down anyway,
+                    // we call Process.WaitForProcessExit(ProcessExitTimeoutMs).
+                    process.WaitForExit(ProcessExitTimeoutMs);
                 }
 
                 if (throwOnError && process.ExitCode != 0)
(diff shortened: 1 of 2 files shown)


alexdresko:
60 seconds sounds like a very long time. I don't have any experience with the bug, but I don't think I'd love having to wait 60 seconds if the child process itself closes relatively quickly.

src/Microsoft.Tye.Core/ProcessUtil.cs
```diff
@@ -118,7 +120,12 @@ public static class ProcessUtil
...
-                    process.WaitForExit();
+                    // Because of the bug in the dotnet runtime https://github.com/dotnet/runtime/issues/29232, Process.WaitForExit()
+                    // hangs for processes that spawn another long-running processes.
+                    // Since these are expected to be long running processes and we're typically not concerned with capturing all of its output
+                    // i.e. it's probably ok for some output to be lost on shutdown, since Tye is shutting down anyway,
+                    // we call Process.WaitForProcessExit(ProcessExitTimeoutMs).
+                    process.WaitForExit(ProcessExitTimeoutMs);
                 }
 
                 if (throwOnError && process.ExitCode != 0)
```
karolz-ms, line 131: If a timeout occurs, what is the `process.ExitCode`? Should we explicitly check for timeout and do something different when it occurs?

pratiksanglikar: This fix is a workaround for a bug in dotnet-runtime - [dotnet/runtime#29232](https://github.com/dotnet/runtime/issues/29232) where the child process is shut down, the call `WaitForExit()` never returns.
This bug appeared just now because we switched our dapr support to use `dapr.exe` which in turn calls a long-running process `daprd` as opposed to calling `daprd` directly.

Now, checking for the exit code in case of a timeout would make sense to probably show a warning to the user that the process didn't shut down gracefully, but at this point, there is no way for us to know whether the child process *really* didn't shut down gracefully or it's just because of the bug in dotnet runtime.

Therefore I think it wouldn't add much value to check for the exit code and do additional actions in the case of timeout.

karolz-ms (approved):
We should document the assumption that

1. `process.ExitCode` is valid even if `WaitForExit()` timed out. I believe this is the case since we got `process.Exited` event.
2. Since process has exited, no additional data will be written to either `outputBuffer`, or `errorBuffer`, so it is (thread-) safe to call `ToString()` on both of them.

Other than that, looks good.

Issue #1138, opened 2021-08-04
Exit code not properly logged in error message


philliphoff:
### Describe the bug
If a watched replica exits due to a non-zero exit code, Tye logs an error message but does not properly include the exit code.

### To Reproduce
1. Start a Tye application in "watch" mode.
2. Kill one of the replica processes.
3. Note `watch: Exited with error code Microsoft.Tye.ProcessResult` message.

### Further technical details

The bug is [here](https://github.com/dotnet/tye/blob/836b0baa8a0ed5d8ca3bebc27c7d4f6d0a7bdfd9/src/Microsoft.Tye.Hosting/Watch/DotNetWatcher.cs#L76); only the `ExitCode` property should be logged, not the whole `ProcessResult`.

PR #1142, opened 2021-08-06
Improves error code log message in watch mode

FossBrandon:
This addresses #1138. I was unable to find a unit test for this change so let me know if you want one added or if one does exist and I would be happy to make any changes.

Before:
image

After:
image

### DIFF
--- a/src/Microsoft.Tye.Hosting/Watch/DotNetWatcher.cs
+++ b/src/Microsoft.Tye.Hosting/Watch/DotNetWatcher.cs
@@ -73,7 +73,7 @@ namespace Microsoft.DotNet.Watcher
                     {
                         // Only show this error message if the process exited non-zero due to a normal process exit.
                         // Don't show this if dotnet-watch killed the inner process due to file change or CTRL+C by the user
-                        _logger.LogError($"watch: Exited with exit code {processTask.Result.ExitCode}");
+                        _logger.LogError("watch: {Replica} process exited with exit code {ExitCode}", replica, processTask.Result.ExitCode);
                     }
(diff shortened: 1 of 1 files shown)


philliphoff:
@FossBrandon I'm not going to turn down an offer to add a unit test, so have at it!  That said, if it ends up being significantly more difficult to do than the substantive change being tested itself, I'm ok without having one.

pratiksanglikar:
@philliphoff Do you think the process name along with the exit code would be useful here?
i.e. The error message could look like - `[15:33:07 ERR] watch: Process 'service_1' exited with an exit code 1`.

philliphoff:
@pratiksanglikar If we have it (or, maybe even better, the replica name) handy, sounds good to me.

FossBrandon:
I would be very happy to try and write a unit test for this and figure out how hard it may be 😀
And I like the suggested improvements on making the log message even more helpful. 

I'll attempt to make those changes in the next couple of days when I get a bit more time!

FossBrandon:
I just pushed the logging changes to include the replica information in a format that matches logging from the Process Runner.
image

I also attempted to create a test for this code in the `TyeRunTests` file by creating a new `testassets` project that just exits immediately with a non zero exit code, but I ran into issues when it came to retrieving the logs. As far as I can tell, we don't have a method of reading the tye logs that are sent to standard output so I can't directly test this logging. If you guys know of a good way to get these logs or can think of a better way to test this, let me know. Otherwise, I don't think it is directly testable.

pratiksanglikar (approved):

PR #1146, opened 2021-08-11
Log the Pid after Tye application is started

pratiksanglikar:
The tooling needs the `pid` of the application after it is started to match it to the applications the tooling has started and then execute operations on it (e.g. shutdown).

[microsoft/vscode-tye#127](https://github.com/microsoft/vscode-tye/issues/127) is such request from tooling where `pid` is required.

This PR adds a log to emit the `pid` after the application has started.

### DIFF
--- a/src/Microsoft.Tye.Hosting/TyeHost.cs
+++ b/src/Microsoft.Tye.Hosting/TyeHost.cs
@@ -109,6 +109,7 @@ namespace Microsoft.Tye.Hosting
             try
             {
                 await _processor.StartAsync(_application);
+                _logger.LogInformation($"Application {_application.Name} started successfully with Pid: {Process.GetCurrentProcess().Id}");
             }
             catch (TyeBuildException ex)
             {


src/Microsoft.Tye.Hosting/TyeHost.cs
```diff
@@ -109,6 +109,7 @@ public async Task<IHost> StartAsync()
             try
             {
                 await _processor.StartAsync(_application);
+                _logger.LogInformation($"Application {_application.Name} started successfully with Pid: {Process.GetCurrentProcess().Id}");
```
philliphoff, line 112: I don't know if it makes any significant difference, but would it be better to log this as early as possible to reduce the length of time required to scan the process output?

pratiksanglikar: It doesn't really wait for all the processes to start. The heavy wait processors like `DockerRunner` don't wait on the containers to be up, so we get this output relatively quickly.

The current information message gives the impression that all the processes are started, but I think I can modify the information message a little bit to reflect that not all the processes are started, but Tye Process is started.

philliphoff: Ok, that's fine.  Since the intent is for tooling to read this message, maybe something that's a little more "regular" or "extractable"?  I know there's a balance with wanting readibility, but maybe something more terse like:

```
Application: {_application.Name}
PID: {Process.GetCurrentProcess().Id}
```

philliphoff (approved):

pratiksanglikar:
The unit tests are flaky and fail randomly with errors that aren't related to this PR.
Opened a bug to make unit tests more robust. #1148

PR #1193, opened 2021-10-04
Switch to 1ES R&D pools on main

jonfortescue:
Per https://github.com/dotnet/core-eng/issues/14276.


philliphoff (approved):
Looks good, assuming successful builds.

PR #1194, opened 2021-10-05
Support Dapr "standalone" applications

### LINKED ISSUE #1172: Cannot run a dapr service that does not expose HTTP
### Describe the bug
Today, it looks like the Dapr sidecar only loads if the binding is set to HTTP.  That is a catch 22 for a non http service, like a headless client or server.  If the service binding is set to HTTP then tye will wait to see that port get exposed before running.  And if you turn off HTTP, then sidecar will not load.

### To Reproduce
1) Clone the Dapr Hello-world quickstart from this fork branch:
https://github.com/paulyuk/quickstarts/tree/paulyuk/tye
`git clone -b paulyuk/tye https://github.com/paulyuk/quickstarts.git`

2) cd to /quickstarts/hello-world

3) Per the readme, do the npm install, and pip3 commands to restore packages.

4) `tye run`
** the pythonapp sidecar will not load, so this service cannot resolve the name and port of `nodeapp`

5) uncomment the explicit 5000 port binding for pythonapp
https://github.com/paulyuk/quickstarts/blob/paulyuk/tye/hello-world/tye.yaml#L30

sadly, the app still breaks because tye is waiting forever for this headless app to load on port 5000, which it isn't designed to do.

### LINKED ISSUE #1178: Dapr extension fails silently if you do not set explicit HTTP bindings
Dapr extension fails silently if you do not set explicit HTTP bindings.  I did not realize this but if you dont set a binding, it will default to TCP and then Dapr will mysteriously not load side cars or throw errors/warnings (couple of hours debugging this).

Paul did not specify http on the bindings and this made them all start as tcp, which in turn stopped dapr from running the sidecars

Can we do better here, default to http maybe? Detect sidecar not started and give message in log
Maybe a setting to say a sidecar is NOT needed for a service so we can then give a good error message if sidecar does not start

philliphoff:
Tye currently assumes that all Dapr-ized services are intended to be invoked by other services (or by the Dapr runtime itself, in terms of receiving messages from subscriptions).  This implies that all Dapr-ized services have an HTTP binding (in order to be reachable by the Dapr runtime).  If a service has no HTTP bindings, no Dapr runtime will be started for that service.

However, not all services require that; some may *only* make outbound calls to other services, directly or via pub-sub (e.g. subscribe to and receive message from an alternative source, process them, then invoke another service via Dapr).  In this case, while the Dapr runtime is still needed but the application may not expose an HTTP endpoint (i.e. intended for Dapr as it has no need).  The Dapr runtime supports this scenario, by omitting the `--app-port` parameter on startup.

This change allows more nuance in the Dapr extension.  The default behavior remains that, if a service has an HTTP binding, a Dapr runtime will be started for it, using that bound port.  Otherwise, it will not.  However, the developer can now override that behavior using a new per-service configuration setting.  In addition, if a service has only a single binding, and that binding has no explicit protocol, it will be assumed to be HTTP for the purposes of Dapr-ization.

```yaml
extensions:
- name: dapr
  services:
    app2:
      enabled: true
    app3:
      enabled: false
services:
# app1 will be Dapr-ized because it has an HTTP binding
- name: app1
  bindings:
  - port: 3000
  - protocol: http
# app2 will be Dapr-ized because it was explicitly enabled (but without specifying the `--app-port`)
- name: app2
# app3 will *not* be Dapr-ized because it was explicitly disabled above
- name: app3
  bindings:
  - port: 3000
  - protocol: http
```

> This new per-service configuration section will also form the basis for other general Dapr settings to be overridden; this PR only allows the `placement-port` setting to be overridden.

Resolves #1172
Resolves #1178

### DIFF
--- a/src/Microsoft.Tye.Core/Serialization/YamlParser.cs
+++ b/src/Microsoft.Tye.Core/Serialization/YamlParser.cs
@@ -77,6 +77,31 @@ namespace Tye.Serialization
             return app;
         }
 
+        public static Dictionary<string, object> GetDictionary(YamlNode node)
+        {
+            if (node.NodeType != YamlNodeType.Mapping)
+            {
+                throw new TyeYamlException(node.Start,
+                    CoreStrings.FormatUnexpectedType(YamlNodeType.Mapping.ToString(), node.NodeType.ToString()));
+            }
+
+            var dictionary = new Dictionary<string, object>();
+
+            foreach (var mapping in (YamlMappingNode)node)
+            {
+                var key = YamlParser.GetScalarValue(mapping.Key);
+
+                dictionary[key] = mapping.Value.NodeType switch {
+                    YamlNodeType.Scalar => YamlParser.GetScalarValue(key, mapping.Value)!,
+                    YamlNodeType.Mapping => YamlParser.GetDictionary(mapping.Value),
+                    _ => throw new TyeYamlException(mapping.Value.Start,
+                            CoreStrings.FormatUnexpectedType(YamlNodeType.Mapping.ToString(), mapping.Value.NodeType.ToString()))
+                };
+            }
+
+            return dictionary;
+        }
+
         public static string GetScalarValue(YamlNode node)
         {
             if (node.NodeType != YamlNodeType.Scalar)
--- a/samples/dapr/pub-sub/tye.yaml
+++ b/samples/dapr/pub-sub/tye.yaml
@@ -8,10 +8,6 @@ name: dapr
 extensions:
 - name: dapr
 
-  # If you want to use a different tag or container port
-  # placement-image: daprio/dapr
-  # placement-container-port: 50005
-
   # log-level configures the log level of the dapr sidecar
   log-level: debug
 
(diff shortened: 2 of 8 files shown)


philliphoff:
Oops, this was not intended to be published until #1191 was committed, as it builds off of it.

src/Microsoft.Tye.Core/Serialization/YamlParser.cs
```diff
@@ -77,6 +77,32 @@ public ConfigApplication ParseConfigApplication()
...
+                    CoreStrings.FormatUnexpectedType(YamlNodeType.Mapping.ToString(), node.NodeType.ToString()));
+            }
+
+            var dictionary = new Dictionary<string, object>();
+
+            foreach (var mapping in (YamlMappingNode)node)
+            {
+                var key = YamlParser.GetScalarValue(mapping.Key);
+
+                dictionary[key] = mapping.Value.NodeType switch
```
ravipal, line 94: Should we throw error when the key already present?

ravipal (approved):

Issue #618, opened 2020-08-05
Add url paths on the tye.yaml to make it easier to navigate to available endpoints


malisancube:
### What should we add or change to make your life better?

The idea is to have something like this on the `tye.yaml`

```
- name: results
  project: results/results.csproj
  -paths: /swagger;/hangfire;/lively;/page=2&pageSize=10
```
This would then be exposed on the dashboard as URLs. e.g. https://localhost:5001/swagger

### Why is this important to you?

This would make it easier to simply navigate to endpoints that you want to use in that service and save a couple of keystrokes.

spboyer:
@davidfowl / @jkotalik - could we look at Routes and provide these in the dashboard without having to define them explicitly?

davidfowl:
No routes are part of the application. There's no way to get that information statically from outside the app.

digitaldias:
+1 for this. I'm writing a set of microservices that do not have coupling between them (no REST endpoints) but instead use a shared messagebroker for all actions. As such, it'd be nice to have all my services point straight to their /graphql endpoint when clicked from the tye dashboard. This is an important feature to add. It can be read from the launchSettings.json in the property "launchUrl"

ante-maric:
+100 for this! Judging by the time this issue was created...holding our breath could get us killed.

ignacioerrico:
Believe it or not, I was going to ask if anyone was working on this. You beat me to it, @ante-maric. :)

I wanted to discuss some finer details too, like how about `routes` instead of `paths`, and if there was a preferred way to display the links (instead of repeating the host and port for each, I would just display the same as it's being displayed now with the list of routes underneath, all of them clickable).

Example. This configuration (note I didn't include a hyphen in the `routes` item, as in the original post):

```
- name: results
  project: results/results.csproj
  routes: /swagger;/hangfire;/lively;/page=2&pageSize=10
```

would produce something like this:

```
http://localhost:54321/
- /swagger
- /hangfire
- /lively
- /page=2&pageSize=10
```

PR #1200, opened 2021-10-15
Added support for additional routes on the Dashboard

ante-maric:
This PR covers issue #618 for supporting additional routes on the Dashboard in Bindings column. This makes it easier to navigate to endpoints that you access frequently like /swagger.


davidfowl:
Some design feedback: Routes should be an array instead of a semicolon delimited String

ante-maric:
> Some design feedback: Routes should be an array instead of a semicolon delimited String

Thanks for the feedback, I have applied the requested changes.

digitaldias:
Any traction on this? Looking like it's closer now :)

philliphoff:
@ante-maric Looks good to me, and rather handy! Thanks for the submission!  The only change I've made is to use blocks for the binding spans, as it starts to get hard to read with many routes.

@ravipal Since I've made changes myself, can you review as well?

ravipal (approved):

Issue #1213, opened 2021-11-08
Move Tye to .NET 6


philliphoff:
### What should we add or change to make your life better?

* Move Tye to .NET 6 now that it's GA.
* Consider moving to all-in-one executable released in both x86 and ARM flavors for each platform.
  * Consider then eliminating dependency on `dotnet tool` mechanism for installation (i.e. have Tye be a standalone CLI)

### Why is this important to you?

Developers will tend to migrate to .NET 6 being the new LTS version, meaning fewer will have .NET Core 3.x installed (currently required for Tye). Moving to .NET 6 will also avoid x86/arm issues currently seen on Apple Silicon.

If we were to also move to all-in-one executables, we could also eliminate the need for a particular version of .NET runtime just for the purposes of running Tye.  Eliminating our dependency on `dotnet tool` would also be helpful for those wishing to use Tye with non-.NET development platforms.

Serge-N:
This essential, stilling having trouble running tye and native dotnet 6.

philliphoff:
Notes re: single-file executables:

* There are uses of `Assembly.Location` which are apparently [not compatible](https://docs.microsoft.com/en-us/dotnet/core/deploying/single-file#api-incompatibility) and would need refactoring.

paulquinn:
The updated CI build instructions don't seem to work (at least for me " error NU1102: Unable to find package microsoft.tye with version (>= 0.11.0)") - has the 0.11 package been published?

Also, those instructions add a dotnet5 package source - should this be moved to dotnet6?

philliphoff:
@paulquinn There were some corporate governance issues that blocked publishing CI builds to the CI feed.  These *should* be now resolved, but it may take some time and/or additional CI builds before they start showing up again.  As for the package source, while that sounds reasonable, I'll have to dig into the build infrastructure to see how it decides which feed Tye packages go into.

cphillips83:
I would argue against removing the `dotnet tool` integration because its easy for a developer to do a `dotnet tool restore` instead of following a large CLI tool installation process. A dotnet tool manifest also helps enforce tool versioning.

cphillips83:
@philliphoff Its still not available when using 

```
dotnet tool update -g Microsoft.Tye --version "0.11.0-*" --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json
```

philliphoff:
[...]

PR #1215, opened 2021-11-11
Move Tye to .NET 6

### LINKED ISSUE #1216: CI: move to latest Mac OS

The macOS-10.14 environment is deprecated [from Azure DevOps] and will be removed on December 10, 2021.  Given that, we should move to a later (or latest) Mac OS environment for our CI builds.

I want the CI build to succeed. :-)

philliphoff:
Moves the Tye CLI and related components (like the container proxy) to .NET 6.  Updates test asset projects to .NET 6 as well.  These changes mean development machines must have .NET 6 installed in order to use Tye *in addition* to the .NET SDK/runtime used by the projects built/started by Tye.  For example, if your application still targets .NET Core 3.1, you will need to have both the .NET 6 runtime *and* the .NET Core 3.1 SDK/runtime installed.

Existing .NET Standard libraries (e.g. the configuration library) will continue to target .NET Standard.

> NOTE: Samples are *not* planned to be updated with this PR.  This work will be done in a follow-up PR.

Related to #1213
Resolves #1216


src/Microsoft.Tye.Proxy/Program.cs
```diff
@@ -19,7 +19,7 @@ public class Program
         static async Task Main(string[] args)
         {
             var serviceName = Environment.GetEnvironmentVariable("APP_INSTANCE");
```
ravipal, line 21: Do we need similar check for "APP_INSTANCE"?

src/shared/KubectlDetector.cs
```diff
@@ -47,10 +47,16 @@ public static Task<bool> IsKubectlConnectedToClusterAsync(OutputContext output)
...
                         continue;
                     }
-                    var major = int.Parse(element.Value.GetProperty("major").GetString());
-                    var minor = int.Parse(element.Value.GetProperty("minor").GetString().Trim('+'));
-                    var version = new Version(major, minor);
-                    return version;
+                    var majorString = element.Value.GetProperty("major").GetString();
+                    var minorString = element.Value.GetProperty("minor").GetString();
+
+                    if (majorString != null && minorString != null)
```
ravipal, line 53: will `element.Value.GetProperty("major").GetString()` ever return null?

ravipal (commented):
Should this https://github.com/dotnet/tye/blob/main/docs/getting_started.md be updated as well?

src/shared/KubectlDetector.cs
philliphoff: `GetProperty()` will return `null` if the property doesn't exist, which is unlikely, but could conceivably happen.

src/Microsoft.Tye.Proxy/Program.cs
philliphoff: No, because we don't rely on it in the same way we do the other (i.e. it's logged but not actually used).

philliphoff:
>  Should this https://github.com/dotnet/tye/blob/main/docs/getting_started.md be updated as well?

Yes, thanks for thinking of that.  I made a quick pass through the docs and updated outdated references to .NET Core.

src/shared/KubectlDetector.cs
ravipal: I think it throws error instead of returning null. Even if it returns null, the .GetString() will result into null ref exception.

ravipal (approved):

src/shared/KubectlDetector.cs
philliphoff: Ah, yes, sorry, my mistake.  It's `GetString()` which could return null (if the property exists but is set to `null`).

robertsundstrom:
Too late, but this is missing:

In order for Tye to work on a Mac that has both the arm64 and x64 SDKs you need to supply the ``` -a x64```parameter when installing the tool. That is because Tye is dependant on .NET Core 3.1 which is x64.

Example:

```
dotnet tool install -a x64 -g Microsoft.Tye --version "0.10.0-alpha.21420.1"
```

Issue #1051, opened 2021-05-19
AAD Pods Managed Identity Integration


emarchal78:
### What should we add or change to make your life better?
Would be very appreciate if we could set the label "aadpodidbinding" by the tye.yaml schema

### Why is this important to you?
AAD Pods Managed Identity are strongly recommended by MS as best practice security on AKS

PR #1222, opened 2021-11-19
Allow optional generation of AKS Azure Active Directory Pod ID Bindings

Hylaean:
[Azure Active Directry Managed Identities](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview), while a new feature, alleviates many from the burden of dealing with secrets (for the price of a SPOF?).
It is becoming a security standard in many organizations. In the one I belong to, we've been using a custom version of tye to comply with this regulation.

In AKS, Azure's managed Kubernetes offering, enabling [this feature](https://github.com/Azure/aad-pod-identity) involves the steps documented there [aadpodidbing](https://docs.microsoft.com/en-us/azure/aks/use-azure-ad-pod-identity)

This pull requests allows to control the generation of the necessary labels directly using two documented settings.

At the project level:

 #### `aksGenerateAadPodIdBindings` (boolean)

AKS specific. Using Managed Identities requires the generation labels for [aadpodidbing](https://docs.microsoft.com/en-us/azure/aks/use-azure-ad-pod-identity). 
When set to true these labels will be generated by convention, unless overriden at the project or ingress level.
The convention is: 
    $"{application.Name}-{project.Name}-service"; // for services
    $"{application.Name}-{ingress.Name}-ingress"; // for ingresses

At the service or ingress levels

#### `aksPodIdentityName` (string)

An optional name of pod identity to use for the service when generating the aadpodidbinding label.


Hylaean:
This PR would resolve #1051

philliphoff:
@Hylaean Thanks for the submission!  While I can see the benefit in having integration with AKS and AAD Pod Identity, I'm a bit reluctant to fold something so coupled to a vendor specific target platform into the general Tye configuration schema.

My initial thought is that this feature would be better done as a Tye extension (like Dapr), which, when configured, adds the labels to the ingress- and project-service-builders, which are then copied over into the Kubernetes manifests by the existing manifest generator step.  The `ProjectServiceBuilder` already has a `ManifestInfo` property where annotations/labels could be added by an extension (which will then find their way into the YAML).  That could be brought over to `IngressBuilder`, so the extension could do the same for those.

Hylaean:
Hi @philliphoff , thanks for the tip, I hadn't thought of that, although I had the same reticence regarding design impact, it was basically a small cleanup of an internal hack, I'll try to do it in the suggested way when I "find time".

davidfowl:
+1 @philliphoff This seems a little azure specific, not sure this belongs in tye.

PR #1227, opened 2021-11-24
Update macOS agent

sebastienros:
https://devblogs.microsoft.com/devops/hosted-pipelines-image-deprecation/


sebastienros:
Saw a notification on my repository, so I assume I would do it here too.

philliphoff:
Thanks for the submission! Closing in favor of #1270. We both had a similar PR at the same time, though I neglected to updating the naming (which is now resolved).

PR #1268, opened 2022-01-14
Respect configured host in bindings

### LINKED ISSUE #721: Services don't reflect the `host` setting
### Describe the bug
It seems that the [`host`](https://github.com/dotnet/tye/blob/master/docs/reference/schema.md#host-string) is ignored. At least for `ServiceType.Project`.
I need my service to be available for other remote services via a given hostname, not just on localhost.

### To Reproduce
When I set [`services->bindings->host`](https://github.com/dotnet/tye/blob/master/docs/reference/schema.md#host-string) it doesn't get reflected in my app. It just falls back to `localhost`:

https://github.com/dotnet/tye/blob/170876742603e8af260f860b66412946fb05cfe0/src/Microsoft.Tye.Hosting/ProcessRunner.cs#L223

On the other hand, the UI shows the host set correctly:

https://github.com/dotnet/tye/blob/170876742603e8af260f860b66412946fb05cfe0/src/Microsoft.Tye.Hosting/Dashboard/Pages/Index.razor#L96

Temporarily, I overcame the issue by specifying `args: --urls http://*:5000` as you can see in my [yaml](https://github.com/petrsvihlik/WopiHost/pull/105/files#diff-cbc33f9a2d863e8cb06546d9534bf32d37a6eb2d4fedf65852c72160e7d6f4db).

### Further technical details

- `tye --version`: 0.5.0-alpha.20520.1+739de2eeb94f0246f50fa4993e953abb4b0da18a
- Platform: Windows

veloek:
When setting ASPNETCORE_URLS env variable in ProcessRunner. This makes it possible to override which interface to bind to and also allow connections from anywhere by using a wildcard.

This is only tested for services configured as projects. Not sure how this affects other aspects of Tye.

Fixes #721

### DIFF
--- a/src/Microsoft.Tye.Hosting/DockerRunner.cs
+++ b/src/Microsoft.Tye.Hosting/DockerRunner.cs
@@ -215,7 +215,7 @@ namespace Microsoft.Tye.Hosting
 
             var dockerImage = docker.Image ?? service.Description.Name;
 
-            async Task RunDockerContainer(IEnumerable<(int ExternalPort, int Port, int? ContainerPort, string? Protocol)> ports, CancellationToken cancellationToken)
+            async Task RunDockerContainer(IEnumerable<(int ExternalPort, int Port, int? ContainerPort, string? Protocol, string? Host)> ports, CancellationToken cancellationToken)
             {
                 var hasPorts = ports.Any();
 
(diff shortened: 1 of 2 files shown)


philliphoff:
@veloek In general, looks good.  The only thing I've noticed in testing is that, while `tye run` will now respect the specified host, `tye run --docker` will *not*.  That is, the container port will still be mapped to all IP addresses of the host (which is the Docker default for mapped ports).  It appears this happens for similar reasons, where [DockerRunner.cs](https://github.com/dotnet/tye/blob/c2bae89a098d361867a4f4e475d0c476f4002087/src/Microsoft.Tye.Hosting/DockerRunner.cs#L240) ignores the `Host` property of the binding.

veloek:
@philliphoff, I didn't test Docker containers myself, but I see your point. Will try to sort that out as well 🙂

veloek:
@philliphoff, good catch! Didn't try running my services in Docker containers, but it makes sense to fix that as well while at it 👍

philliphoff:
@veloek Take a look at my update to the PR and see if anything looks amiss.

@ravipal Since I've made changes, can you do a quick review as well?

veloek:
@philliphoff LGTM. If I were to nitpick, I see you use capital S in `String` which seems inconsistent with the rest of the code.

philliphoff:
Ah, the hazards of owning codebases with entirely different style guides. :-)

PR #1272, opened 2022-01-17
Fix wrong variable in null check

veloek:
Probably just a copy/paste typo. Spotted it coincidentally while browsing the code.


philliphoff (approved):
Good catch!

Issue #998, opened 2021-04-02
Tye build project c# with exiting Dockerfile


thinhnotes:
### Describe the bug
In current I had a issues when I used the tye with run project c# with DockerFile

### To Reproduce
example in your source code in samples/azure-functions/voting

I edit file tye.yaml in bellow:

```
name: VotingSample
services:
- name: worker
  project: worker-function/worker-function.csproj
```

run tye build -v debug

and it return the error:
![image](https://user-images.githubusercontent.com/4660531/113405544-3535cc00-93d4-11eb-8e35-9a6509dba5b9.png)

In current I guest it can mapping the context of build folder to docker build and not file the profile file build
I had try the property a dockerFileContext  but it not working on this case

Could you please to help me to resolve it

jkotalik:
Yeah so tye will automatically grab the DockerFile from the folder if one is present. I think we need to add an option to ignore the docker file present; I've hit this issue as well. 

As a workaround for now, you may need to rename your dockerfile present in the directory.

thinhnotes:
thanks for support, I will be try the workaround solution and wait the new info when the option to ignore the docker file is release, thanks

philliphoff:
I think the question is whether the user really *wants* Tye to use the existing `Dockerfile` by default, if present.  In other words, is the user only telling Tye to ignore the existing `Dockerfile` because Tye doesn't do the right thing with it and not for any other reason.  Would the user prefer to make the choice "opt in", say by using the existing `dockerFile`/`dockerContext` configuration properties (which seem to be ignored for `tye build`), and otherwise have Tye generate its own `Dockerfile` for build and deployment purposes?  If the user's gone to the trouble of creating a Dockerfile, presumably the user knows better than Tye how images should be built.  Therefore, I'm inclined to side with Tye respecting the existing the `Dockerfile` by default.

[...]

PR #1273, opened 2022-01-17
Added configuration option to ignore project Dockerfiles

samtrion:
The PR provides a configurable solution for Issue #998.

### DIFF
--- a/src/Microsoft.Tye.Core/ApplicationFactory.cs
+++ b/src/Microsoft.Tye.Core/ApplicationFactory.cs
@@ -193,6 +193,9 @@ namespace Microsoft.Tye
 
                         // Do k8s by default.
                         project.ManifestInfo = new KubernetesManifestInfo();
+
+                        // Ignore DockerFile, if exists
+                        project.IgnoreDockerfile = configService.IgnoreDockerFile;
                     }
                     else if (!string.IsNullOrEmpty(configService.Image))
                     {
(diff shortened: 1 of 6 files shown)


philliphoff:
@samtrion Thanks for the submission!  As I mention in the related issue, I'm not sure adding a "ignore Dockerfile" flag would be the best approach, unless your use case is that you *really don't* want Tye to use the existing `Dockerfile`.  Rather, if the real issue is that Tye won't work with your existing `Dockerfile`, I'd rather see that addressed, either by defaulting instead to the current practice of using multi-stage `Dockerfiles` (existing or not), and/or adding a flag to the service configuration or `tye build|deploy` command for explicitly choosing single- or multi-stage.

samtrion:
@philliphoff Unfortunately I have the experience that the Project own `Dockerfile` often gets no ❤️ / maintenance. It is simply a dead piece of code / configuration, or the Dockerfiles created by Tye run much cleaner / more stable. (👏 Thank you very much for that)

Currently I have exactly this case, I need for my complete product among other things 3 microservices, which are developed by other teams. Unfortunately they don't maintain the project's own `Dockerfile` and don't accept external PR's to delete, rename or even update them. I helped myself for a while by including a `delete script` in the build process. However, that is now not a solution either, as we are deploying a new component with a heavily customized `dockerfile`, which was also deleted as a result. It was just not the right approach.

I also usually prefer to fix something that doesn't work. However, there will certainly always be cases where this is not possible. Like codefreeze, stubborn maintainers or just lack of permissions.

PR #1290, opened 2022-02-01
Synchronize output buffers

philliphoff:
Avoids a race condition where the CLI utility method may attempt to consume the accumulated standard and error output while output is still being written.

Related to #1280.

### DIFF
--- a/src/Microsoft.Tye.Core/ProcessUtil.cs
+++ b/src/Microsoft.Tye.Core/ProcessUtil.cs
@@ -76,6 +76,27 @@ namespace Microsoft.Tye
                 }
             }
 
+            var outputLock = new SpinLock();
+
+            void WithOutputLock(Action action)
+            {
+                bool gotLock = false;
+
+                try
+                {
+                    outputLock.Enter(ref gotLock);
+
+                    action();
+                }
+                finally
+                {
+                    if (gotLock)
+                    {
+                        outputLock.Exit();
+                    }
+                }
+            }
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.Tye.Core/ProcessUtil.cs
```diff
@@ -129,15 +150,21 @@ public static class ProcessUtil
                     process.WaitForExit(ProcessExitTimeoutMs);
```
ravipal, line 150: `process.WaitForExit()` supposed to wait for the output and error stream to complete, so, if that is the case, when we get to read from `outputBuilder` and `errorBuilder`, it should be completely written right?

philliphoff: For `Process.WaitForExit()` that is the case, but not `Process.WaitForExit(int)`, which can timeout, in which case output may continue to be written.

ravipal: Ah. Just noticed your note/comment.

ravipal (approved):

src/Microsoft.Tye.Core/ProcessUtil.cs
philliphoff: I believe we switched to `Process.WaitForExit(int)` to avoid a deadlock when the process doesn't want to die (which happens in certain cases of a .NET process starting *another* .NET process and piping the output).

PR #1291, opened 2022-02-01
Don't run Mac/Linux in official build

mmitche:
These don't contribute official artifacts and otherwise would need to switch to 1ES pools.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -97,69 +97,70 @@ stages:
             artifactType: Container
             parallel: true
 
-      - job: Linux
-        pool:
-          vmImage: ubuntu-20.04
-        variables:
-        - name: _SignType
-          value: none
-        - name: MSBUILDDISABLENODEREUSE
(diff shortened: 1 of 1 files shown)


philliphoff:
@mmitche Can you provide any more context about this change?  In general, I'm fine with it, as long as tests continue to run, packages continue to be published, and governance tasks continue to scan.

mmitche:
> @mmitche Can you provide any more context about this change? In general, I'm fine with it, as long as tests continue to run, packages continue to be published, and governance tasks continue to scan.

The Mac and Linux jobs don't actually produce any artifacts for official builds. They're also run in the public rolling CI already (e.g. https://dev.azure.com/dnceng/public/_build/results?buildId=1582579&view=results), so they are not needed in the official build. governance and all the rest still runs.

If we don't make this change, we'll need to switch these jobs over to 1ES pools in the internal project.

philliphoff (approved):

PR #1353, opened 2022-04-09
Update 00_run_locally.md

MBDev2k:
GetServiceUri() from Tye extensions library


memark (commented):
I don't see why you would want to do this change. The GetServiceUri() extension method should already be in scope here given the default imports.

PR #1481, opened 2022-12-10
Fix tye's proxy for websockets

davidfowl:
- We need to add the middleware

### DIFF
--- a/src/Microsoft.Tye.Hosting/HttpProxyService.cs
+++ b/src/Microsoft.Tye.Hosting/HttpProxyService.cs
@@ -113,6 +113,8 @@ namespace Microsoft.Tye.Hosting
 
                             builder.Configure(app =>
                             {
+                                app.UseWebSockets();
+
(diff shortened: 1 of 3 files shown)


src/Microsoft.Tye.Hosting/HttpProxyService.cs
```diff
@@ -113,6 +113,8 @@ public async Task StartAsync(Application application)
 
                             builder.Configure(app =>
                             {
+                                app.UseWebSockets();
```
philliphoff, line 116: This seems too easy; wouldn't it have been more fun to replace this entire proxy service with YARP? :-)

philliphoff (approved):

src/Microsoft.Tye.Hosting/HttpProxyService.cs
davidfowl: This is my plan 😄

PR #948, opened 2021-02-19
Add start stop button on ServiceDetails page

dalibormesaric:
This PR adds Start/Stop button on ServiceDetails page to enable starting and stopping service on demand.

tye

This is motivated by the following scenario:
I have several dotnet applications running locally, and I might want to debug one of them using visual studio (code). To do this now, I would need to stop tye, comment out the service I want to debug, run tye again, then run my application with debugger.

Alternative is to add the button on the Index page in every row.

Related to https://github.com/dotnet/tye/issues/876

Thanks to  @dillenmeister for the initial idea.

Issue #876, opened 2021-01-01
Make it possible to start/stop/restart single services


jooooel:
### What should we add or change to make your life better?
I'm using Tye to run a couple of infrastructure containers (SEQ, Postgres, Cassandra, RabbitMQ, etc), a couple of our services as containers and the service(s) I'm currently working on as csproj(s). About ~20 services all in all.

I jump a lot between branches (code reviews, fixes here and there, etc) and most of the time I want to keep a base set of containers running sine they hardly ever change between revisions. But, I would like to be able to stop, start and restart specific services (both containers and csprojs) on demand.

Restarting Tye entirely can take a long time, when maybe all I really want is to restart a single service or change the version of an image. I have tried to use `watch` where that is applicable, but unfortunately that causes VS to hang depending on what parts of the code that changes (a common lib for example). And it doesn't work when changing image versions.

I think there are a couple of different commands that would help me in my daily work. All of these would require that Tye can run in detached mode:

1. `tye run <service name>`. Would make it possible to add a specific service to the current Tye instance.
2. `tye stop <service name>`. Would stop a specific service.
3. `tye restart <service name>`. Would restart a specific service.

I wanted to see how feasible this is with the current architecture of Tye, so I did a small PoC where I made it possible to restart a single (csproj) service from the Dashboard. [You can have a look at it here](https://github.com/jooooel/tye/pull/1/files). It seems to be working fine, but would require a bit of refactoring, it's not the prettiest solution at the moment.

I would love to spend more time on this and make a proper PR, and I really think it would add a lot of value.

### Why is this important to you?
Tye has really simplified my daily workflow, in that it lets me run a combination of containers and csprojs together. I also really like that I can configure them and easily decide what to run in a single file. 

Unfortunately, jumping around between branches forces me to restart Tye entirely between each run, which takes a lot of time.

jkotalik:
Thanks for the feedback @jooooel . Your scenarios is definitely not unique, and the suggestion you have seems promising.

We have a few ideas around improving the inner loop speed in tye, especially around running multiple times. One idea we are thinking about is a daemon mode where tye is always running and commands to tye run would just add them to the daemon. That way in your example, you could arbitrarily start and stop specific services accordingly. This also would allow you to have things always running in the background without having them restart on each execution of tye run.

I could see it looking like this:
- `tye run` is the same as today, ctrl-C to kill all things that are started with tye run
- `tye start` does tye run but returns after all projects/images are started
[...]

PR #1495, opened 2022-12-23
Csteeg/feature/add startstop

csteeg:
I didn't see PR #948 before, but it's kind of the same feature, so this one is also #876 

This feature is because our dev-team wanted to have the @dalibormesaric in #948
We have quite a large set up with loads of services in Tye and watching them all can be stressful for the developer's environment. So, we just want to quit the one service you're working on in Tye and use visual studio for that service.

The other scenario is that we quickly want to start just a few services to test or reproduce something without starting the complete environment. Therefore, I placed the buttons on the overview page instead of the service page, and added a command-line option to skip startup for specific projects (eg we have a Microsoft Fhir server project in there taking quite some time and resources). Now I could also say skip "*" so no project starts automatically and just spint up like the 3 services I need atm from the dashboard

This PR also works for docker containers that are running + I chose to reuse the created method to start the services by adding a reference to the method on the service object. This way we don't have to inject the processrunner anywhere.

### DIFF
--- a/src/Microsoft.Tye.Core/HostOptions.cs
+++ b/src/Microsoft.Tye.Core/HostOptions.cs
@@ -11,6 +11,8 @@ namespace Microsoft.Tye
         public bool Dashboard { get; set; }
 
         public List<string> Debug { get; } = new List<string>();
+        
+        public List<string> NoStart { get; } = new List<string>();
 
         public string? DistributedTraceProvider { get; set; }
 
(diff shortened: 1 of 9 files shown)


davidfowl:
Seems like alot of features in a single PR. Can you split these out into separate changes?

csteeg:
Yes sorry about that, I created my new wishes (there are even more on my current main branch) in one branch hoping that I could do them 1 by 1, but decided to pick the features and create a PR for them separately later. This one now only contains the start/stop feature

allensiho:
Looks like great feature,
If merged how long before available to public?
Cheers

Issue #1013, opened 2021-04-17
.NET 6 hot reload support.


jspuij:
Yesterday I hacked together rudimental hot reload support that is coming in .NET 6 for asp.net projects. For me this is useful because with Service Fabric and angular I'm having inner development loop times of several minutes, and it's untenable. I'm looking at tye with Dapr with Blazor, just for this reason.
This is just a trial to see whether it would work, but to make it into a decent PR, a bit of discussion might be necessary. Also, I'm not familiar with the inner workings of tye, which is why this may seem rather hacky. I'm very open to suggestions for improvement.

The changes are here:

https://github.com/dotnet/tye/compare/main...jspuij:net6hotreload

The way I did it was to disable the internal file watching of tye and run `dotnet watch run` instead of the executable directly. There are some advantages and disadvantages to this approach:

Advantages:

- Hot reload seems to be working and certainly improves inner loop time.
- RCL static web asset resolution works, which doesn't work when you start the executable directly. (this is another issue with the current tye).

Disadvantages:

- Tye file watching is replaced with dotnet file watching. With hot reload the executable will keep running. With hot reload disabled, or in case of a failure, tye routines that would normally run when a process is restarted, won't be run because `dotnet watch run` will keep running, even though the child process might be restarted. No idea whether this is important, but it's a consequence.
- Hot reload does not work with the `--no-build` switch, so tye will do another build of the asp.net project. This might be something that can be debated with the aspnetcore team. @SteveSandersonMS might have some tips on the hot reload mechanism.
- I think there might be very wonky results if two microservices have hot reload enabled and have a reference to a shared library. I'm not sure that the shared library is even monitored for changes, but if it is, this is a scenario that should be tested and looked at.
- Nothing tried or tested with projects run inside a container instead of directly on the host.

This might not be the route to take at all, but even then this issue can serve as a starting point for implementing hot reload support in Tye.

robertsundstrom:
I am waiting for this functionality to enter. 

As I see it:

* Disable internal file watcher for .NET 6 projects (and beyond) that support Hot Reload.
* Launch these projects with the  "dotnet watch" command.

aL3891:
Any word on this one?
would be awesome to see it in the mainline releases :)

AndrewBabbitt97:
Where is this on the radar with .NET 6 now being GA?

robertsundstrom:
I guess that this would require you to switch to .NET 6. Tye is still on .NET Core 3.1.

[...]

PR #1496, opened 2022-12-23
Hotreload feature

csteeg:
Based on  #1013 but now optional
I added a 'hotReload' property to project-info (see samples/frontend-backend/tye.yaml where front-end is now using hotReload)
It only starts when combined with --watch. All projects that don't have this property set, still use DotnetFileWatcher.

Although I don't see when you'd prefer DotnetFileWatcher is preferred over dotnet watch, it still is backwards compatible since you must explicitly tell that a project should use dotnet watch.

### DIFF
--- a/samples/frontend-backend/tye.yaml
+++ b/samples/frontend-backend/tye.yaml
@@ -6,3 +6,4 @@ services:
   project: backend/backend.csproj
 - name: frontend
   project: frontend/frontend.csproj
+  hotReload: true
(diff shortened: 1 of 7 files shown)


davidfowl:
I'm not a fan of this change. The watcher can't be an external calls because we have no control over building in that case. The right way to do this is to integrate watch with tye (the way it is done today).

csteeg:
I somewhat get that, but hot-reload is so much faster then a rebuild and thus a restart. Also, ,the watch from dotnet feels a little more stable (eg I never get dll's that are locked with dotnet watch where I do get that in the tye builder)

But, now with the start-/stop feature from my other PR, we're able to stop the service in tye and start it in hot-reload mode from visual studio. The ingress from tye keeps working to proxy the requests, so that's cool. We could live without this feature now.

Just out of curiosity since I don't see the need at our development environment: Why do we want control over building?

csteeg:
Point 1: The file locking issues are with tye only for us... perhaps I should take a look at that someday.
Point 2: yes, understood. We don't use replica's in our dev-environment but I see that you'd need to be in control here

Issue #1507, opened 2023-01-06
Incoming http/2 requests forward to http/1.1 are not transformed


csteeg:
### Describe the bug
When the browser sends http/2 cookie format (multiple headers with the same name 'Cookie'), these are also forwarded to the services even though those requests are send as http/1.1 
The requests should be transformed to become valid http 1.1 requests.

PR #1508, opened 2023-01-06
Fix http/2 cookies for 1.1 requests

csteeg:
This fixes (at least part of) #1507 

I took the code from Microsoft yarp project, I guess it would be a nice feature to use all proxy-features from that project someday in the future


davidfowl (approved):
We need to migrate to YARP

PR #1610, opened 2023-09-22
Delete NOTICE.txt

sebastienros:
Q: Does my public GitHub repository need a NOTICE?
A: Yes, if you have copied anything that is under an open source license into your repo, then your repo needs a NOTICE. If you only reference open source sitting outside your repo (using a package manager or git submodules), then that referenced open source does not need to be included on a NOTICE. For instructions on creating a NOTICE for your repo, see NOTICE in a Microsoft open source project.

https://docs.opensource.microsoft.com/using/guidance-for-open-source-usage/notice-faq/


karolz-ms:
@davidfowl what say you?

sebastienros:
Why: We started a new oss project and there was no NOTICE in it. I was previously maintaining one in dotnet/crank probably because I saw it in dotnet/tye and the open source project guidelines were not clear enough. Contributors to the new repo were wondering why it would need one so I contacted legal and they pointed me to this FAQ in the same oss guidelines. It's clear now that NuGet references don't need to be mentioned. So I am deleting it in dotnet/crank and following up here too as I didn't find any direct source code mentions, just NuGet usage.

philliphoff:
@sebastienros As I read the FAQ, a NOTICE is required if redistributing other OSS bits, which Tye does by including OSS dependencies in `Microsoft.Tye.<version>.nupkg`. So, while it may not be required for the repo _itself_, it should be included in the `.nupkg`, and that's typically done (at least for VS Code extensions) by including a version kept in the repo and periodically updated using the governance site NOTICE generator. (It appears that that NOTICE isn't actually being included in the `.nupkg`, which is something to be fixed.)

sebastienros:
@philliphoff I had a conversation with @karolz-ms offline and we agreed that a NOTICE is not necessary for NuGet references only but we still need to keep it here. 

Even shipping a NuGet package is fine without this file, contrary to what you are suggesting I believe. However, we need this file when we ship a build that contains third-party libraries (compiled or in source). Since Tye is shipping one (or was) in the form of a dotnet tool, this file is necessary in the shipped assets. But like you are also pointing, having the file in the repo is not sufficient, it needs to be included with the tool. Not in `Microsoft.Tye.<version>.nupkg` though.

Issue #1608, opened 2023-08-31
Seq memory limits, API key etc.


rolfik-mycronic:
I have tried using Seq logging extension and it launches Seq docker container and our services log to it when properly configured.
But I have not found a way how to specify other parameters like:
- [memory limits](https://docs.datalust.co/docs/docker-memory)
- API key, file buffering etc. used by Serilog

To overcome it, I have tried this approach instead of the extension:

```
services:
- name: seq
  image: datalust/seq
  volumes:
    - source: ./.logs
      target: /data
  env:
  - ACCEPT_EULA=Y
  - SEQ_API_KEY=something
  bindings:
  - protocol: http
    port: 5341
    containerPort: 80
    connectionString: "http://${host}:${port}?apiKey$={env:SEQ_API_KEY}"
```

But while it launches Seq container, it does not propagate seq connectionString to our services where it should be used to configure Serilog with Seq.

Any ideas?

Thank You

Phiph:
hey you should be able to access the SEQ connection string for your config using the `CONNECTIONSTRINGS
SERVICE-SEQ` environment variable that gets injected to each of your services. 

Question: Why do you need an API key for a local development when using SEQ?

rolfik-mycronic:
API key is used to change logging level at runtime and it can be used in production as well

Phiph:
Are you attempting to use tye to deploy SEQ aswell?

rolfik:
No, we use docker compose for deployment, but reuse tye environment variables at runtime.

Phiph:
Cool I'll happily take a look, but I'm struggling to understand why you need a SEQ api key for local dev environment - I don't use SEQ in my lifecycle so any more information you can give me will help. 


We can easily extend the SEQ extension to include the other variables, but I'm uncertain whether including the APIkey value is worthwhile considering all config for SEQ may be removed when the container restarts, or when another dev in your project pulls and runs it.

rolfik:
I have not investigated more why Seq connectionString is not available at other service start, but that is what I have seen when degugging a service.
My idea was to specify Seq API key in environment variable file specific for a developer. Or even better in global environment variable.
Seq config is stored on attached local data volume so it is kept there.

Phiph:
I think the reason why SEQ as an extension works is because it set's itself as a dependency for all other services 

[...]

Issue #652, opened 2020-09-01
Control the startup order ?


mmisztal1980:
Is tye capable of controlling the startup order in any shape or form?

For instance, I'd like to use an elasticsearch container, which can take ~60s to start, I wouldn't like to deploy my services beforehand.

In docker-compose v2, one could use the **depends_on** with healthchecks, to start dependent containers when their dependency was in a healthy state. A readiness probe, acting as the gate, would do as well.

Is this doable?

tebeco:
not a maintainer here but what would be the idea "synch" point

that look like a custom route to call on elastic like a healthcheck

that is already doable with an httpclient call in a loop inside the `Program.cs>Main` of your webapi until you get a more official answer

also that would look like the equivalent of `initContainer` there was an issue about it few days ago is there a way to block the `initContainer` until done ? maybe yes or no
or just a loop and curl ;)

mmisztal1980:
I guess an equivalent of a readiness probe would be great. For instance - having an init container with a migrator for a pgsql database is pointless to start before and actual postgres container is up and running, right?

tebeco:
That could also works, it only depends on your particular needs TBH

mmisztal1980:
one would need to structure their `tye.yaml` in a way that ensures that core infrastructure is up and running 1st - which makes sense

cecilphillip:
Control order would be also useful for non-application services. Things like rabbitMQ, Consul, vault and others that depend on each other but need to be started up in a specific order.

tebeco:
To be fair, this look like this should be done by the application code and not dotnet tye.
Especially because when you need, for example to patch the OS of a Node in AKS, K8s will try to drain pods from that node to other available node in the cluster.
you might need to define pod affinity or "taint" to be able to be sure, for example, that all pods of "RabbitMq" are not on the same Node, but still the problem remains.

Your application should be resilent at startup time or at runtime against "middelware health"
The ability to plug HealthCheck can (should/must) be used if you depend on an external system
You could also plug the `readinessProbe` to avoid incomming traffic.

But adding api to "hope" startup time will be fine from `tye` will probably give a false sense of "resiliency" while this problem can happen at any moment in production

Here is something that happened last month:
* We had a Drain
* We had LOTS of pod on that node
* The min/max pod count (surge....) forced K8s to create before deletion
* The equivalent of "Rabbit/Consul..." was down for ... reason + attempt of Drain (like if you need "at least 3 pod for Rabbit to have a consensus)
[...]

Issue #443, opened 2020-05-06
depends_on & healthcheck


staff0rd:
### What should we add or change to make your life better?
Some kind of `depends_on` functionality where services don't immediately come up, and rather wait to come up until other services are either up successfully, or, responding to a successful `healthcheck`
### Why is this important to you?
1.  I have a scenario where a specific project cannot recover from being started _before_ a dependent container is both up completed it's startup.
2. It is not clear without diving into logs whether the entire system is up or not and in a ready state. It would be nice if a `healthcheck` could be configured and then exposed to both `depends_on` and to the dashboard.

davidfowl:
> Some kind of depends_on functionality where services don't immediately come up, and rather wait to come up until other services are either up successfully, or, responding to a successful healthcheck

I think this is nice to have from a developer POV but bad in practice if you don't provide the same when you deploy. I think your application has to be resilient to this.

> It is not clear without diving into logs whether the entire system is up or not and in a ready state. It would be nice if a healthcheck could be configured and then exposed to both depends_on and to the dashboard.

I think the dashboard could be improved to show what's running and healthy and what's not. Replicas and restarts do that but in a very subtle way.

PR #1613, opened 2023-10-11
add dependency map

Phiph:
This PR aims to solve #1608 & #652

it introduces the dependsOn yaml that looks for services that match the name and ensure the startup order is honoured. 

For Example:

```yaml
name: VotingSample
services:
- name: vote
  project: vote.csproj
  dependsOn:
    - redis
- name: worker
  include: ../worker/tye.yaml
- name: redis
  image: redis
  bindings:
    - port: 6379
      connectionString: ${host}:${port}
```

Should await the Redis server to start before the vote service. 

I will write more E2E tests to validate this and ensure that environment variables are added to services.

### DIFF
--- a/src/Microsoft.Tye.Core/ConfigModel/ConfigService.cs
+++ b/src/Microsoft.Tye.Core/ConfigModel/ConfigService.cs
@@ -46,5 +46,7 @@ namespace Microsoft.Tye.ConfigModel
         public string? Version { get; set; }
         public string? Architecture { get; set; }
         public string? CloneDirectory { get; internal set; }
+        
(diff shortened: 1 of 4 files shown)


davidfowl:
> Should await the Redis server to start before the vote service.

This will leave you with brittle service code. It encourages the wrong defaults.

Phiph:
referring issue #443

Phiph:
> > Should await the Redis server to start before the vote service.
> 
> This will leave you with brittle service code. It encourages the wrong defaults.

Totally get that your services should be fault tolerant, but I think It would help support the way tye injects env variables. Do you have any suggestions :)

